PR #11, opened 2025-08-05
configurations-endpoint: adding SOVD configurations endpoint

MarkSchmitt:
- lists all services in service 0x22 and 0x2E, deduplicates them by removing Write, Read and Dump suffixes and return them as one entry plus their uds payloads
 - renamed functional_classes maps, so they are reflecting common naming patterns (lookup) and also added the functional class ID to the service, necessary for the lookup


cda-interfaces/src/datatypes/database_naming_convention.rs
```diff
@@ -0,0 +1,127 @@
...
+                return long_name[affix.len()..].to_string();
+            }
+            if self.long_name_affix_position == DiagnosticServiceAffixPosition::Suffix
+                && long_name_lowercase.ends_with(affix)
+            {
+                return long_name[..long_name.len() - affix.len()].to_string();
+            }
+        }
+        long_name
+    }
```
sagovic, line 69: This functionality can be unit tested.

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -1799,4 +1912,125 @@ impl EcuManager {
...
+    ///
+    /// Returns a tuple `(bitlength, value)` or `None` if no matching parameter is found.
+    fn get_service_id_parameter_value(&self, service: &&DiagnosticService) -> Option<(u32, u32)> {
+        let request = match self.ecu_data.requests.get(&service.request_id) {
+            Some(request) => request,
+            None => {
+                log::warn!(
+                            target: &self.ecu_data.ecu_name,
+                            "No request found for service {service:#?}");
+                return None;
```
sagovic, line 1930: opinion: I would rather add dynamic values as attributes instead of target. But not sure whether log supports this. In tracing this would be 
```tracing::warn!(ecu = self.ecu_dta.ecu_name,,"...")```

File: cda-interfaces/src/datatypes/database_naming_convention.rs [hunk 3]
```diff
@@ -0,0 +1,127 @@
...
+    pub short_name_affixes: Vec<String>,
+    pub long_name_affixes: Vec<String>,
+}
+
+impl DatabaseNamingConvention {
+    /// Trims a diagnostic service long name using the configured affixes and naming position.
+    /// The first matching affix is removed and the result is returned.
+    /// Affixes must be lowercase for correct matching.
+    /// Returns the trimmed name or the original if no affix matches.
+    pub fn trim_long_name_affixes(&self, long_name: String) -> String {
```
sagovic, line 54: suggestion: How about &str here? This way you can pass &str and String w/o explicit conversion. Will help you during unittests =D

alexmohr (approved):

PR #19, opened 2025-08-11
extracted diag_service and split into data / configurations

MarkSchmitt:
- extracted diag_service so it can offer common functionality for both data/{diag_service} and configurations/{diag_service} endpoints.
 - moved post and put handlers from data (which is only read) to configurations (which is only write)

Mark Schmitt [mark.schmitt@mercedes-benz.com](mailto:mark.schmitt@mercedes-benz.com), Mercedes-Benz Tech Innovation GmbH [Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)

### DIFF
--- a/cda-interfaces/src/lib.rs
+++ b/cda-interfaces/src/lib.rs
@@ -61,11 +61,19 @@ pub struct DiagComm {
 
 #[derive(Debug, Clone)]
 #[cfg_attr(feature = "deepsize", derive(DeepSizeOf))]
+/// Enum representing diagnostic communication types according to ASAM SOVD.
+///
+/// Can be mapped to UDS service prefixes with [`DiagCommType::service_prefix`]
 pub enum DiagCommType {
+    /// Service Prefix `0x2E`
     Configurations,
+    /// Service Prefix `0x22`
     Data,
+    /// Service Prefixes `0x14`, `0x19`
     Faults,
+    /// Service Prefixes `0x10`, `0x11`, `0x28`, `0x85`, `0x27`, `0x29`
     Modes,
+    /// Service Prefixes `0x2F`, `0x31`, `0x34`, `0x36`, `0x37`
     Operations,
 }
 
--- a/cda-sovd/src/sovd/components/ecu/mod.rs
+++ b/cda-sovd/src/sovd/components/ecu/mod.rs
@@ -13,16 +13,22 @@
 
 use axum::{
     Json,
+    body::Bytes,
     extract::{Query, State},
     response::{IntoResponse, Response},
 };
-use cda_interfaces::{UdsEcu, diagservices::DiagServiceResponse, file_manager::FileManager};
-use http::StatusCode;
+use cda_interfaces::{
+    DiagComm, UdsEcu,
+    diagservices::{DiagServiceResponse, DiagServiceResponseType},
+    file_manager::FileManager,
+};
+use http::{HeaderMap, StatusCode, header};
 use serde::Deserialize;
 
 use crate::sovd::{
     IntoSovd, WebserverEcuState,
-    error::{ApiError, ErrorWrapper},
+    error::{ApiError, ErrorWrapper, api_error_from_diag_response},
+    get_payload_data,
 };
 
 pub(crate) mod configurations;
@@ -72,6 +78,7 @@ pub(crate) async fn get<
             variant,
             locks: format!("{base_path}/locks"),
             operations: format!("{base_path}/operations"),
+            configurations: format!("{base_path}/configurations"),
             data: format!("{base_path}/data"),
             sdgs: sdgs.map(|sdgs| sdgs.into_sovd()),
             single_ecu_jobs: format!("{base_path}/x-single-ecu-jobs"),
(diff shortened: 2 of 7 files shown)


cda-sovd/src/sovd/components/ecu/mod.rs
```diff
@@ -142,3 +149,106 @@ impl IntoSovd for cda_interfaces::datatypes::ComParamSimpleValue {
...
+            type_: DiagCommType::Data,
+            lookup_name: None,
+        },
+        Op::Write | Op::Update => DiagComm {
+            name: service_name.clone(),
+            action: DiagCommAction::Write,
+            type_: DiagCommType::Configurations,
+            lookup_name: None,
+        },
+    };
```
theswiftfox, line 177: I'm not sure if I like this approach.  
I think this is originally from it being in /data and reused there - but for example deciding if the type is `Configuration` or `Data` seems very fragile. How would we handle it if we have a PUT in /data for some reason in the future? (even if its only theoretical, the architecture should imo support it)

I think sharing everything from `get_payload_data` on is fine, but the DiagComm should probably come from outside here - then `DiagCommType::Configurations` can be set in the /configuration handlers and `DiagCommType::Data` from /data

cda-sovd/src/sovd/mod.rs
```diff
@@ -189,12 +189,15 @@ pub async fn route<
                     .get(locks::ecu::lock::get),
             )
             .route("/configurations", routing::get(configurations::get))
+            .route(
+                "/configurations/{diag_service}",
+                routing::get(configurations::diag_service::post)
+                    .post(configurations::diag_service::put),
```
theswiftfox, line 195: see comment above - I'd suggest not sharing the complete handlers - but only the common part *inside* of the handler.

theswiftfox (changes requested):

cda-sovd/src/sovd/components/ecu/mod.rs
```diff
@@ -142,3 +149,106 @@ impl IntoSovd for cda_interfaces::datatypes::ComParamSimpleValue {
...
+                        data: mapped_data,
+                    }),
+                )
+                    .into_response()
+            }
+        }
+    }
+}
+
+pub(crate) enum Op {
```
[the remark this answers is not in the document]
MarkSchmitt: I did move it from data to here, the fully qualified path is crate::sovd::components::ecu::Op - a good place imho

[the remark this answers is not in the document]
MarkSchmitt: I moved it again .. the Op is only used in operations and lost one of it's values

cda-sovd/src/sovd/mod.rs
MarkSchmitt: they are not shared, configurations and data both have a sub-mod diag_service, perhaps that was misleading?

imho it makes sense to have the sub-module diag_service in both cases because they are accessing the diagnostic services.

theswiftfox (approved):

cda-interfaces/src/lib.rs
```diff
@@ -59,6 +59,14 @@ pub struct DiagComm {
...
 }
 
+/// Enum representing diagnostic communication types according to ASAM SOVD.
+///
+/// # Variants
+/// - `Configurations`: Write services (`0x2E`)
+/// - `Data`: Read services (`0x22`)
+/// - `Faults`: Fault-related services (`0x14`, `0x19`)
+/// - `Modes`: Mode-related services (`0x10`, `0x11`, `0x28`, `0x85`, `0x27`, `0x29`)
+/// - `Operations`: Operation services (`0x2F`, `0x31`, `0x34`, `0x36`, `0x37`)
```
theswiftfox, line 69: Suggestion: put the docstrings for the variants directly to the variants like so:
```rust
#[derive(Debug, Clone)]
#[cfg_attr(feature = "deepsize", derive(DeepSizeOf))]
/// Enum representing diagnostic communication types according to ASAM SOVD.
///
/// Can be mapped to UDS service prefixes with [`DiagCommType::service_prefix`]
pub enum DiagCommType {
    /// Service Prefix `0x2E`
    Configurations,
    /// Service Prefix `0x22`
    Data,
    /// Service Prefixes `0x14`, `0x19`
    Faults,
    /// Service Prefixes `0x10`, `0x11`, `0x28`, `0x85`, `0x27`, `0x29`
    Modes,
    /// Service Prefixes `0x2F`, `0x31`, `0x34`, `0x36`, `0x37`
    Operations,
}
```

This allows to hover on the concrete variant for example with rust-analyzer and get the description there

Image

MarkSchmitt: oh, that's cool, thanks, will do

floroks: The service prefixes for Configurations and Data are bot - 0x22 and 0x2E, and both support GET/PUT. The included services are determined due to their functional class(es).

theswiftfox: Yeah the original names of the variants are a bit misleading - it is actually an internal mapping used to validate that a Read action doesn't call a 0x2E for example. 
We could change `Configurations` to `Write` and `Data` to `Read` though, if we want to differentiate from the sovd terms.

theswiftfox (approved):
looks good to me.

PR #25, opened 2025-08-13
Use  aide for openapi generation

theswiftfox:
This PR switches from utoipa to aide as it is closer to the code and adds missing api docs to the static routes.

Generating schemas & openapi for services from ecu database will be in a follow up


cda-sovd/src/sovd/components/ecu/genericservice.rs
```diff
@@ -42,21 +44,26 @@ pub(crate) async fn put<
             .into_response();
         }
     };
-
-    let data = match get_payload_data::<DataRequestPayload>(&headers, &body) {
```
theswiftfox, line 46: Note: I changed `get_payload_data` and extracted `get_octet_stream_payload` from it, as it seemed clearer when reading the code, that this endpoint only accepts octet-stream. I stumbled upon it when writing the docs and specifying the input.

cda-sovd-interfaces/src/apps.rs
```diff
@@ -22,8 +22,56 @@ pub mod sovd2uds {
 
     pub mod data {
         pub mod network_structure {
+            use serde::Serialize;
+
+            #[derive(Serialize)]
+            #[serde(rename_all = "PascalCase")]
+            #[cfg_attr(feature = "openapi", derive(schemars::JsonSchema))]
+            pub struct Ecu {
+                pub qualifier: String, // name
```
alexmohr, line 31: question: why not use docstrings instead?

alexmohr (commented):
lgtm, only minor question.

cda-sovd-interfaces/src/apps.rs
theswiftfox: the embarrassing answer: because of copy & paste without checking 🙈
I will fix that - thx! :D

PR #29, opened 2025-08-19
implement and test data extraction/packing according to iso

alexmohr:
* implement data packing according to chapter 7.3.6.4
* implement data extraction according to chapter 7.3.6.3
* add tests for data extraction and packing

### DIFF
--- a/cda-core/src/diag_kernel/ecumanager.rs
+++ b/cda-core/src/diag_kernel/ecumanager.rs
@@ -364,10 +364,10 @@ impl cda_interfaces::EcuManager for EcuManager {
                         .get(&coded_const.diag_coded_type)
                     {
                         Some(coded_const_type) => {
-                            if coded_const_type.base_datatype != DataType::UInt32 {
+                            if coded_const_type.base_datatype() != DataType::UInt32 {
                                 break;
                             }
-                            match &coded_const_type.type_ {
+                            match &coded_const_type.type_() {
                                 DiagCodedTypeVariant::StandardLength(standard_length) => {
                                     standard_length
                                 }
@@ -637,31 +637,17 @@ impl cda_interfaces::EcuManager for EcuManager {
                     })?;
 
                     let uds_val = operations::diag_coded_type_to_uds(
-                        diag_type.base_datatype,
+                        diag_type.base_datatype(),
                         &coded_const_value,
                     )?;
 
-                    match diag_type.type_ {
-                        datatypes::DiagCodedTypeVariant::LeadingLengthInfo(ref val) => {
-                            let byte_len = (val / 8).max(1) as usize;
-                            uds.extend(&uds_val[uds_val.len() - byte_len..]);
-                        }
-                        datatypes::DiagCodedTypeVariant::MinMaxLength(_) => {
-                            todo!("what type is min/max length?? bits, bytes? sausages?")
-                        }
-                        datatypes::DiagCodedTypeVariant::StandardLength(ref val) => {
-                            let byte_len = (val.bit_length / 8).max(1) as usize;
-                            if let Some(mask) = &val.bitmask {
-                                for i in uds_val.len() - byte_len..uds_val.len() {
-                                    uds.push(uds_val[i] & mask[i]);
-                                }
-                            } else {
-                                uds.extend(&uds_val[uds_val.len() - byte_len..]);
-                            }
-                        }
-                    }
+                    diag_type.encode(
+                        uds_val,
+                        &mut uds,
+                        param.byte_pos as usize,
+                        param.bit_pos as usize,
+                    )?;
                 }
-                // skip for now. maybe validate payload according to params and their constraints?
                 _ => break,
             }
         }
@@ -680,14 +666,9 @@ impl cda_interfaces::EcuManager for EcuManager {
                                 param.short_name
                             ))
                         })?;
-                        if let Some(value) = json_values.get(&short_name)
-                            && let Some(uds_val) = self.map_param_to_uds(param, value)?
-                        {
-                            operations::extend_with_bit_pos(
-                                &mut uds,
-                                uds_val,
-                                param.bit_pos as usize,
-                            );
+
+                        if let Some(value) = json_values.get(&short_name) {
+                            self.map_param_to_uds(param, value, &mut uds)?;
                         }
                     }
                 }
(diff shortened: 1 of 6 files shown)


cda-database/src/datatypes/diag_coded_type.rs
```diff
@@ -34,13 +35,822 @@ pub enum DataType {
...
+        Ok(Self {
+            base_datatype,
+            type_,
+            is_high_low_byte_order,
+        })
+    }
+
+    pub fn base_datatype(&self) -> &DataType {
+        &self.base_datatype
+    }
```
theswiftfox, line 150: suggestion: as DataType is a small enum that supports `Copy` we can return by value here. This makes the code a bit easier to read in the places this is used because no deref is required then
```suggestion
    pub fn base_datatype(&self) -> DataType {
        self.base_datatype
    }
```

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -1586,7 +1567,7 @@ impl EcuManager {
                     ))
                 })?;
                 let expected =
-                    operations::diag_coded_type_to_uds(diag_type.base_datatype, &const_value)?
+                    operations::diag_coded_type_to_uds(*diag_type.base_datatype(), &const_value)?
                         .into_iter()
                         // .filter(|v| *v != 0) // remove 0 padding
```
theswiftfox, line 1572: suggestion: Though its not changed by you in this PR, just noticed this commented out line. Should we remove it?

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -2020,6 +2001,7 @@ impl EcuManager {
             DiagDataTypeContainer::RawContainer(DiagDataTypeContainerRaw {
                 data: vec![*payload],
                 data_type: DataType::UInt32,
+                bit_len: 32,
```
theswiftfox, line 2004: suggestion: using of the [constant in u32](https://doc.rust-lang.org/std/primitive.u32.html#associatedconstant.BITS) `u32::BITS`

cda-database/src/datatypes/diag_coded_type.rs
```diff
@@ -34,13 +35,822 @@ pub enum DataType {
...
+        if bit_len == 0 {
+            return Ok((Vec::new(), 0));
+        }
+
+        unpack_data(
+            bit_len,
+            bit_pos,
+            mask.as_ref(),
+            &uds_payload[start_pos..start_pos + (bit_len.div_ceil(8))],
+            self.byte_order(),
```
theswiftfox, line 354: suggestion: If we add `#[derive(Copy, Clone)]` to the `ByteOrder` enum, we can reuse the `byte_order` variable here instead of calling the function again.

cda-database/src/datatypes/diag_coded_type.rs
```diff
@@ -34,13 +35,822 @@ pub enum DataType {
...
+        self.validate_bit_pos(bit_pos)?;
+        let (packed_bytes, bit_len, mask) = match &self.type_ {
+            DiagCodedTypeVariant::LeadingLengthInfo(bit_len) => {
+                // Subtract the next multiple of 8 from the length in usize as bits.
+                // Shift the length up this amount of bits to put the relevant data into
+                // the MSBs of the length usize.
+                let len_byte_count = bit_len.div_ceil(8);
+                let len_bit_len = len_byte_count * 8;
+                let mut data = (input_data.len()
+                    << ((std::mem::size_of::<usize>() * 8) - len_bit_len as usize))
```
theswiftfox, line 387: suggestion: `std::mem::size_of::<usize>() * 8` can be replaced with `usize::BITS`

cda-database/src/datatypes/diag_coded_type.rs
```diff
@@ -34,13 +35,822 @@ pub enum DataType {
...
+    }
+}
+
+/// Normalizes the byte order of a data slice according to the specified order.
+fn normalize_byte_order(data: &mut [u8], byte_order: ByteOrder) {
+    match byte_order {
+        ByteOrder::Keep => {}
+        ByteOrder::ReorderPairs => {
+            // Reverse the byte order in pairs for Unicode2String
+            for i in (0..data.len()).step_by(2) {
```
theswiftfox, line 840: suggestion/question: not sure what your opinion here is, but for something like this I really like `chunks_mut`:
```rust
for chunk in data.chunks_mut(2) {
    if chunk.len() == 2 {
        chunk.swap(0, 1);
    }
}
```
or in iterator style
```rust
data.chunks_mut(2)
    .for_each(|chunk: &mut [u8]| {
        if chunk.len() == 2 {
            chunk.swap(0, 1);
        }
    });
```
But this is purely preference - mostly curious what you think :)

cda-core/src/diag_kernel/ecumanager.rs
alexmohr: cool, done

cda-database/src/datatypes/diag_coded_type.rs
```diff
@@ -34,13 +35,822 @@ pub enum DataType {
...
+        byte_pos: usize,
+        bit_pos: usize,
+    ) -> Result<(), DiagServiceError> {
+        self.validate_bit_pos(bit_pos)?;
+        let (packed_bytes, bit_len, mask) = match &self.type_ {
+            DiagCodedTypeVariant::LeadingLengthInfo(bit_len) => {
+                // Subtract the next multiple of 8 from the length in usize as bits.
+                // Shift the length up this amount of bits to put the relevant data into
+                // the MSBs of the length usize.
+                let len_byte_count = bit_len.div_ceil(8);
```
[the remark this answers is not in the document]
alexmohr: yes but actually no. `usize::BITS` is a u32 we would need the `as usize` here 

```
(usize::BITS - len_bit_len)
 ^ u32.           ^ usize
```


I'd rather keep it in the `[ ]` operator because it's a bit nicer to read when `as usize` is at the end instead of the middle of an expression.
Or am I missing something?

cda-database/src/datatypes/diag_coded_type.rs
alexmohr: Your suggestion is way more readable and more concise, love it!
I opted for the non iterator style, although I like the iterator style more I think saving a tiny bit of time is worth it here.
(I'd be curious why the iterator is slower though, but I don't think it's really worth going down that rabbit hole 😓 )

cda-database/src/datatypes/diag_coded_type.rs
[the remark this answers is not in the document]
theswiftfox: Oh, heh - got me there. I didn't check if both my suggestions applied still compiles :D (std::mem::size_of returned a usize 🙈)
In that case, agreed with your suggestion of keeping it in the `[ ]` operator! Reads better that way!

theswiftfox (approved):
Looks good to me! :)

PR #57, opened 2025-09-18
add support for extended dtcs

alexmohr:
add support for reading extended DTCs including their schema

---
Example request url
 
`curl localhost:20002/vehicle/v15/components/my_ecu/faults/P164456`

See discussion for response and reasoning 

### DIFF
--- a/cda-comm-uds/src/lib.rs
+++ b/cda-comm-uds/src/lib.rs
@@ -18,13 +18,16 @@ use std::{
 };
 
 use cda_interfaces::{
-    DiagComm, DiagCommAction, DiagCommType, DiagServiceError, EcuGateway, EcuManager,
-    SchemaProvider, SecurityAccess, ServicePayload, TesterPresentControlMessage, TesterPresentMode,
-    TesterPresentType, TransmissionParameters, UdsEcu, UdsResponse, datatypes,
+    DiagComm, DiagCommAction, DiagCommType, DiagServiceError,
+    DiagServiceError::InvalidRequest,
+    EcuGateway, EcuManager, SchemaDescription, SchemaProvider, SecurityAccess, ServicePayload,
+    TesterPresentControlMessage, TesterPresentMode, TesterPresentType, TransmissionParameters,
+    UdsEcu, UdsResponse, datatypes,
     datatypes::{
-        ComponentConfigurationsInfo, DataTransferError, DataTransferMetaData, DataTransferStatus,
-        DtcMask, DtcReadInformationFunction, DtcRecordAndStatus, Ecu, Gateway, NetworkStructure,
-        RetryPolicy,
+        ComponentConfigurationsInfo, DTC_CODE_BIT_LEN, DataTransferError, DataTransferMetaData,
+        DataTransferStatus, DtcCode, DtcExtendedInfo, DtcMask, DtcReadInformationFunction,
+        DtcRecordAndStatus, DtcSnapshot, Ecu, ExtendedDataRecords, ExtendedSnapshots, Gateway,
+        NetworkStructure, RetryPolicy,
     },
     diagservices::{DiagServiceResponse, DiagServiceResponseType, UdsPayloadData},
     service_ids,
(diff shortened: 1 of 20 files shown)


cda-comm-uds/src/lib.rs
```diff
@@ -564,6 +567,236 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsMana
...
+
+        let extended_data_json = extended_data_response.into_json()?;
+        let extended_data: Option<HashMap<_, _>> =
+            extended_data_json.data.as_object().and_then(|obj| {
+                obj.iter()
+                    .find_map(|(_, value)| value.as_array())
+                    .map(|arr| {
+                        arr.iter()
+                            .filter_map(|item| item.as_object())
+                            .filter_map(|obj| {
```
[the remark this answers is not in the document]
alexmohr: yields the exact same and is definitely better to read. thank you :)

cda-comm-uds/src/lib.rs
```diff
@@ -564,6 +567,237 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsMana
...
+
+    async fn map_snapshots(
+        &self,
+        ecu_name: &str,
+        dtc_code: DtcCode,
+        include_schema: bool,
+    ) -> Result<(Option<ExtendedSnapshots>, Option<serde_json::Value>), DiagServiceError> {
+        fn extract_schema_properties(
+            schema_desc: &Option<SchemaDescription>,
+        ) -> Option<serde_json::Value> {
```
theswiftfox, line 713: question: do you think it makes sense to push these as a small helper fn into SchemaDescription?
to be able to call it something like `let param_properties = schema_desc.as_ref()?.get_param_properties()?`

I just noticed that almost the identical codeblock is ins map_extended_data, where only the transform over the param_properties is different if i read it correcly.

cda-comm-uds/src/lib.rs
```diff
@@ -1163,7 +1397,7 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsEcu
             .filter(|(_, lookup)| {
                 scope
                     .as_ref()
```
theswiftfox, line 1399: suggestion: `.is_none_or(|scope| scope.to_lowercase() == lookup.scope.to_lowercase())` - The deref is not required to call to_lowercase.

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -2089,6 +2081,9 @@ impl EcuManager {
                 }
                 Ok(())
             }
+            ParameterValue::TableStructParam(_) => {
+                todo!();
```
theswiftfox, line 2085: suggestion: same as for line 2060, tracing::error instead of todo?

cda-core/src/diag_kernel/payload.rs
```diff
@@ -20,6 +20,7 @@ pub(in crate::diag_kernel) struct Payload<'a> {
     current_index: usize,
     slices: VecDeque<(usize, usize)>,
     last_read_byte_pos: usize,
+    additional_bytes_to_skip: usize,
```
theswiftfox, line 23: question: would `bytes_to_skip` also be okay as a name? I think it would make it a bit more straight forward to read 😅

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -2067,6 +2056,9 @@ impl EcuManager {
                                     .to_owned(),
                             ))
                         }
+                        DataOperationVariant::DynamicLengthField(_) => {
+                            todo!()
```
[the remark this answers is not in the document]
alexmohr: Hmm I'd be fine w/ changing it to an error but I don't think it's a good idea to only log it. I would change `:DataOperationVariant::StaticField` too so it's still consistent (that's actually the reason I've opted for todo 😆 ). Does that work for you?

floroks:
> > I can only encourage to stay as close to the standard as possible.
> 
> That's a good point, which I cannot and won't argue against. The endpoint has this field for a reason so it should be used. I made the changes in a separate commit in case @floroks has a different opinion and good arguments to back it up 😄

Using the standard is good - I'd suggest to also use the naming as in the standard - e.g.:

```json
"environment_data": {
   "extended_data_records": ...
   "snapshots": ...
}
```

cda-comm-uds/src/lib.rs
alexmohr: thanks

cda-core/src/diag_kernel/payload.rs
alexmohr: agree 👍

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -2740,6 +2832,28 @@ fn mux_case_from_selector_value<'a>(
     })
 }
 
+fn map_sd_sdg(
```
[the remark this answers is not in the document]
alexmohr: i reverted this change, it's not necessary anymore. if we should need it at some point we can put it in operations.

cda-interfaces/src/datatypes/sdg.rs
```diff
@@ -15,7 +15,7 @@
 use deepsize::DeepSizeOf;
 use serde::Serialize;
 
-#[derive(Debug, Serialize)]
+#[derive(Clone, Debug, Serialize)]
```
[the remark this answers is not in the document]
alexmohr: No it's not needed anymore, similar to extracting the map_sdg function. Reverted this.

alexmohr:
> > > I can only encourage to stay as close to the standard as possible.
> > 
> > 
> > That's a good point, which I cannot and won't argue against. The endpoint has this field for a reason so it should be used. I made the changes in a separate commit in case @floroks has a different opinion and good arguments to back it up 😄
> 
> Using the standard is good - I'd suggest to also use the naming as in the standard - e.g.:
> 
> ```json
> "environment_data": {
>    "extended_data_records": ...
>    "snapshots": ...
> }
> ```

Awesome, thank you. I also changed the other fields to snake_case, to be closer to the SOVD standard 👍

alexmohr:
> Awesome, thank you. I also changed the other fields to snake_case, to be closer to the SOVD standard 👍

Oh one thing about that @floroks . 
The spec is wildly mixing camelCase and snake_case. The example for status is _actually_ camelCase instead. But it looked really odd that everything is snake_case but the `status` field. 
As this is OEM specific anyway, i opted for snake_case here.

cda:
```
    "status": {
      "test_failed": false,
```

spec: 
```
"status":  
        { 
          "testFailed": "1", 
```

This way we are at least somewhat consistent within a response.

floroks:
Unfortunately the spec is very inconsistent, let's use lower snake case there, and if we need other casings, we can think about a config option.

PR #79, opened 2025-10-29
Change database format to flatbuf

alexmohr:
## Summary
Change the ECU data to flatbuf instead using protobuf.
* Decrease memory and CPU footprint significantly 
* Simplify interaction with the DB, as all indirections are resolved by flatbuf and don't have to be resolved manually anymore. 

Depends on https://github.com/eclipse-opensovd/odx-converter/pull/14

### Comparison

Flatbuf is faster in all metrics expect flush_duration and send_duration there the difference is less than 0.2µs which is within the tolerance of the measurement accuracy.

```
Metric                           protobuf         flatbuf      Difference     Winner
--------------------------------------------------------------------------------
flush_duration                     1.97µs          2.08µs          5.45%   protobuf
lock_duration                    328.76µs        320.31µs         -2.57%    flatbuf
mapping_duration                  20.85µs         10.21µs        -51.01%    flatbuf
payload_build_duration            49.82µs         46.28µs         -7.11%    flatbuf
receive_duration                 467.47µs        451.27µs         -3.47%    flatbuf
response_duration                997.30µs        261.27µs        -73.80%    flatbuf
send_ack_duration                159.04µs        149.35µs         -6.09%    flatbuf
send_duration                      2.59µs          2.78µs          7.15%   protobuf
total_duration                   997.07µs        492.56µs        -50.60%    flatbuf
```

## Checklist

## Notes for Reviewers


alexmohr:
> @alexmohr awesome. Thanks for the perf. metrics. Do you still see a use-case for the custom allocator?

@sagovic It's worth to re-evaluate at least. During the tests I did to generate the table above, we had 2.6 GB of allocations (and frees)  with the 'old' protobuf on heap memory, whereas flatbuf only had 26MB
Both tests did not include loading the database and measurements where started around 5s after the service was ready.

cda-comm-uds/src/lib.rs
```diff
@@ -1310,21 +1308,9 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsEcu
...
-                    "Write" => DiagCommAction::Write,
-                    unknown => {
-                        return Err(format!(
-                            "Variant Detection for ECU {ecu_name} failed due to unknown operation \
-                             type {unknown}"
-                        ));
-                    }
-                };
-
+                let (name, _) = req.split_once('_').unwrap();
```
theswiftfox: I know this has been here before - just noticed now that we have an unchecked `unwrap` here. 
If it is expected to never faild we should probably use `.expect` with a descriptive message.

If you agree, I can create an issue to check for any unwraps in the code and if any are found to either change them to use results or expect for cases that are practically infallible.

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -98,13 +96,30 @@ pub struct EcuManager<S: SecurityPlugin> {
...
+    session: Option<String>,
+    security: Option<String>,
     /// resets session and or security access back to the default
     /// after a given time
     access_reset_task: Option<JoinHandle<()>>,
 }
 
+#[derive(Default)]
+struct DbCache {
+    pub(crate) diag_services: Mutex<HashMap<StringId, Option<CacheLocation>>>,
```
theswiftfox: question: out of curiosity is there a specific reason for Mutex instead of RwLock here?

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -454,74 +420,70 @@ impl<S: SecurityPlugin> cda_interfaces::EcuManager for EcuManager<S> {
...
+                    None
+                }
             })
         }) {
             // in case of a positive response update potential session or security access changes
-            if *t == datatypes::ResponseType::Positive {
-                if let Some(new_session) = payload.new_session_id.as_ref() {
-                    self.set_session(*new_session, Duration::from_secs(u64::MAX))?;
-                }
+            let response_type = response.response_type().try_into()?;
```
theswiftfox: nitpick: lines 460 and 461 should be swapped. the comment is related to the if on line 462

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -454,74 +420,70 @@ impl<S: SecurityPlugin> cda_interfaces::EcuManager for EcuManager<S> {
...
-                        _ => false,
-                    }
+        if let Some((response, params)) = responses.iter().find_map(|r| {
+            r.params().and_then(|params| {
+                if params.iter().map(datatypes::Parameter).any(|p| {
+                    p.byte_position() == 0
+                        && p.specific_data_as_coded_const()
+                            .is_some_and(|c| c.coded_value().is_some_and(|v| v == sid))
+                }) {
+                    Some((r, params))
```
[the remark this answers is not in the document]
alexmohr: yes by storing the mapped params in a vector and using this instead of the original params.
Saves two calls to .map() (495 and while getting the base_offset). I don't think it's makes a difference for performance but it's a bit nicer to read 👍

cda-comm-uds/src/lib.rs
alexmohr: https://github.com/eclipse-opensovd/classic-diagnostic-adapter/issues/80

cda-interfaces/src/lib.rs
```diff
@@ -127,10 +174,44 @@ pub mod service_ids {
...
+    LazyLock::new(|| HashSet::from([service_ids::READ_DATA_BY_IDENTIFIER]));
+
+static FAULTS_PREFIXES: LazyLock<HashSet<u8>> = LazyLock::new(|| {
+    HashSet::from([
+        service_ids::CLEAR_DIAGNOSTIC_INFORMATION,
+        service_ids::READ_DTC_INFORMATION,
+    ])
+});
+
+static MODES_PREFIXES: LazyLock<HashSet<u8>> = LazyLock::new(|| {
```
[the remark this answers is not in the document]
theswiftfox: Thanks for the change - juuust another very small nitpick. I think we should declare them as `const` instead of `static` as they should not change during the runtime of the program (see https://doc.rust-lang.org/rust-by-example/custom_types/constants.html#constants)

theswiftfox:
Hm, good point. I'd like to have this in soon, as otherwise it will be a continuous chain of rebasing to do which always has the risk of accidentally loosing stuff. 
Maybe we can put a hint in the readme that this requires the feature branch of the converter at the time being? Not sure what the best approach here is, as it looks like there are quite a few open topics on the converter still (that won't change the format though)

alexmohr:
@theswiftfox squashed the changes into two commits (your initial commit and the rest), lost the approval in the process. Info on slack is posted. as soon as it's approved again I'll merge it :)

PR #89, opened 2025-11-06
add integration tests

alexmohr:
## Summary

* add converted mdd file
* add basic integration tests

Depends on #88 

## Checklist

## Notes for Reviewers


testcontainer/cda/Dockerfile
```diff
@@ -0,0 +1,28 @@
+# Build stage
+FROM rust:1.88-trixie AS builder
+
+WORKDIR /build
+
+COPY Cargo.toml Cargo.lock ./
+COPY clippy.toml rustfmt.toml deny.toml ./
```
lh-sag:
```suggestion
```
As the line below copies the current directly, does it?

testcontainer/cda/entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh -e
+
+DOIP_TESTER_IP=$(ip -4 a show scope global | grep -oP '(?<=inet\s)\d+(\.\d+){3}')
```
lh-sag: I use ```0.0.0.0``` in my setup and it works. Avoids to discover the IP manually and is often used in services.

testcontainer/ecu-sim/docker/build_docker.sh
```diff
@@ -1,9 +1,16 @@
-#!/bin/sh -e
-SCRIPT_DIR=$(dirname "$(realpath "$0")")
-TARGET_DIR=$(realpath "$SCRIPT_DIR/..")
+FROM gradle:8-jdk21 AS builder
```
lh-sag:
```suggestion
ARG JAVA_VERSION=21
FROM gradle:8-jdk${JAVA_VERSION}-alpine AS builder
```

testcontainer/ecu-sim/docker/build_docker.sh
```diff
@@ -1,9 +1,16 @@
...
+WORKDIR /build
+COPY . .
+RUN gradle build --no-daemon
 
-echo "Building ecu-sim"
-"$TARGET_DIR/gradlew" -p "$TARGET_DIR" build shadowJar
+FROM eclipse-temurin:21
```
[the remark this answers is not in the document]
lh-sag: Same as above.

testcontainer/README.md
```diff
@@ -1,9 +1,79 @@
...
+### Prerequisites
+- Docker and Docker Compose installed
+- ODX converter built (optional, for PDX to MDD conversion)
+
+### Setup
+
+1. **Generate ODX files and prepare environment:**
+   ```sh
+   cd testcontainer
+   ./setup.sh
```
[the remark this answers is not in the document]
alexmohr: Yes but actually no. I decided not to provide this file as it was a bit over engineered and didn't bring to much benefit anyway. I forgot to remove it from here.

testcontainer/ecu-sim/docker/build_docker.sh
```diff
@@ -1,9 +1,16 @@
...
 
-echo "Building docker container for ecu-sim"
-docker build -f "$TARGET_DIR/docker/Dockerfile.run" "$TARGET_DIR" -t ecu-sim
+RUN apt-get -y update && apt-get -y install tcpdump vim iproute2 jq curl
```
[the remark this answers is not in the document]
alexmohr: jq and iproute2 are needed for the ipcli.sh, whichs adds additional ip addresses, so we can simulate multiple gateways. The rest of the depencies is not needed and removed

testcontainer/cda/Dockerfile
alexmohr: 🙈 thank you

testcontainer/ecu-sim/docker/build_docker.sh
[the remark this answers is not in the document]
floroks: the other dependencies are extremely helpful when you want to generate tcpdump for analysis in automated pipelines if something fails, vim and curl are just a nice-to-have for when you need to work inside the docker container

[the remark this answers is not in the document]
alexmohr: I messed up here anyway by putting the dockerfile into the shell script 🤦
I've added an argument `DEV_TOOLS` to the docker container. If set it will install the additional packages.
When using the build_script (which is now reverted back to the sensible state before) it will install the tools. docker compose does not the the argument so it's not installed in the CI.

testcontainer/cda/entrypoint.sh
alexmohr: @floroks does that actually matter within docker compose?

floroks: The "no point" was a bit hyperbolic - I just think it's best to catch possible integration issues as early as possible by using a setup that's closest to the real world use-case. I'm not firm on using ubuntu as the base image, there are several other small images which might be a better choice due to smaller size, like debian-slim or distroless. If building the base image is taking too long for some reason, we could also pre-build our own base-images in ghcr, and use them. In the end we just need to make sure to get a good balance between size/duration of execution and closeness to real-world use-case.

lh-sag: @floroks @alexmohr using ```0.0.0.0``` is actually a common practice in docker/compose/etc. You can also assign an IP in compose.yml instead.

What was my initial motivation? The description to setup the testcontainer simply did not work and if I removed these snippets to retrieve ip/etc it worked. And I was not the only one who stumbled across these issues.

Bottom line, we should provide something which works out of the box and avoid people turning away from this projects. So feel free to pick your solution which fits the best but provide a good first user experience with OpenSOVD.

alexmohr: I Changed the docker build quite a bit since yesterday.
* Split the CDA docker images into a bunch of smaller stages to utilize caching better (or at all)
* Using debian trixie slim as base image now. 
* Adjusted the CI, so it actually pushes into the cache. 

This brought build time down significantly. 
Prior to this the runner had to rebuild everything on every push, so the base image doesn't matter that much anymore. 
Imho debian slim is a good trade off between realism and size. 

About the IP address: For the integration tests I'd like to keep it close to real world and thus keep one interface address. 
@lh-sag Could you open an issue with some details what you did and some output like @floroks asked as well? I'd like to deal with this separately, from my point of view this PR is ready

PR #94, opened 2025-11-12
Feature/offline detection

alexmohr:
## Checklist

## Notes for Reviewers


testcontainer/odx/generate.py
```diff
@@ -175,3 +175,17 @@ def generate_for_ecu(
     gateway_address=0x1000,
     functional_address=0xFFFF,
 )
+
+# These ECUs are used to test duplicate logical addresses
+generate_for_ecu(
```
alexmohr: @floroks What's your opinion on this? 
On the one hand it's kinda neat to use this to test how duplicated ECUs behave in an integration test, on the other hand it feels a bit wrong to explicitly add this 'broken' state into the testcontainer. 
Most things can (and are) tested in unit tests too.

floroks: In general I think it's fine, I'd prefer to make a more realistic use-case out of it - have FLXC1000 and FLXCNG1000 with the same logical addresses, but disjunctive variants. This would mirror an use-case, where for different markets different hardware revisions (for whatever reasons) are used, but due to having the same function, the logical address would be the same.

In the sim we can then switch between one and the other by changing the variant dynamically (it also doesn't have to be FLXC, can be a new ECU and logical address.

alexmohr: > In the sim we can then switch between one and the other by changing the variant dynamically (it also doesn't have to be FLXC, can be a new ECU and logical address.

Does this mean we have to able which 'duplicate' is currently active during variant detection?
Right now they are always marked as duplicate if they share the same addresses. During variant detection we're probably able to detect which of both is currently used, mark that as online and all duplicates as offline.

floroks: > Does this mean we have to able which 'duplicate' is currently active during variant detection? 

Yes, you can have multiple ECUs with the same logical address, and the logical address is present, the ECU is marked as Duplicate, Once one of the variants matches, it needs to be marked as Online.

alexmohr: whoops 🙈 thanks for clarifying, well in that case it makes a lot of sense to implement the more realistic use case as you suggested. That will keep me busy for a bit 😆

alexmohr: Done, implemented exactly like you suggested. Also the DB of the variants detected as duplicate is now unloaded. (Although we have to reload it during variant detection)

cda-comm-uds/src/lib.rs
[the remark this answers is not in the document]
floroks: > hmm well if we have two different ECUs running under the exact same name but different logical addresses, this still fails as self.ecus has the ecu name as key

In theory, if only one of the two logical addresses is online, you could determine the one that is visible. This is however not a use-case we need to support. 

When multiple mdds with the same ecu name are found, we could:
1. Ignore all of them, and output an error when logical addresses are different
2. Use the newest one (determined by revision) when logical addresses are the same (also output a warning in the log)

PR #95, opened 2025-11-13
fault-by-id make schema param optional

alexmohr:
without adding a default or wrapping the field
'include-schema' in an option, serde cannot parse the value when the field is not given.
Fixed by adding a default_false impl.

## Checklist

## Notes for Reviewers

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2839,6 +2839,7 @@ dependencies = [
  "schemars",
  "serde",
  "serde_json",
+ "serde_qs",
  "strum",
 ]
 
--- a/cda-sovd-interfaces/Cargo.toml
+++ b/cda-sovd-interfaces/Cargo.toml
@@ -26,6 +26,9 @@ serde_json = { workspace = true }
 schemars = { workspace = true, features = ["hashbrown"] }
 strum = { workspace = true, features = ["derive"] }
 
+[dev-dependencies]
+serde_qs = { workspace = true }
+
 [features]
 default = []
 auth = []
--- a/cda-sovd-interfaces/src/components/ecu/mod.rs
+++ b/cda-sovd-interfaces/src/components/ecu/mod.rs
@@ -466,6 +466,7 @@ pub mod faults {
                 /// If true, snapshot data from 0x19 04 is included in the response
                 #[serde(default = "default_true")]
                 pub include_snapshot_data: bool,
+                #[serde(default)]
                 pub include_schema: bool,
             }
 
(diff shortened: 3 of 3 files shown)


lh-sag (commented):
IMHO this one is worth adding a unit tests.

In general how should query parameters be handled?
* Fail with 502
* Assuume default value and continue
* Consistent behaviour across alls queries or with exceptions

alexmohr:
I don't see too much benefit of a unit test here, added one anyways, that tries to parse the object from a query string. 
I won't add tests that verify this in the `get` handler where this is used in the end, as this will open another can of worms. 

Handling query parameters in general, it depends. For all standard parameters (which include-schema is) the SOVD spec defines how they should be handled if they are optional etc. 
As far as I can tell the spec is not defining what should be done in an error case. 
When a parameter is not optional but is not parsed 400 (bad request) makes most sense. If we failed to parse it  500 or 502 seems appropriate.
Assuming default values only makes sense for optional parameters.  

> Consistent behaviour across alls queries or with exceptions

We should be as consistent as possible but right now I'm not sure whether the SOVD actually is. At least with case styles it's not, so I would assume it deviates sometimes from a given pattern.

alexmohr:
> > I don't see too much benefit of a unit test here, added one anyways, that tries to parse the object from a query string. I won't add tests that verify this in the `get` handler where this is used in the end, as this will open another can of worms.
> > Handling query parameters in general, it depends. For all standard parameters (which include-schema is) the SOVD spec defines how they should be handled if they are optional etc. As far as I can tell the spec is not defining what should be done in an error case. When a parameter is not optional but is not parsed 400 (bad request) makes most sense. If we failed to parse it 500 or 502 seems appropriate. Assuming default values only makes sense for optional parameters.
> > > Consistent behaviour across alls queries or with exceptions
> > 
> > 
> > We should be as consistent as possible but right now I'm not sure whether the SOVD spec actually is. At least with case styles it's not, so I would assume it deviates sometimes from a given pattern.
> 
> I meant unittests on the router level but checking the source base I see none. Take a look at https://github.com/tokio-rs/axum/blob/main/examples/testing/src/main.rs for the gist of it.
> 
> On this level test should be very easy and could handle lots of cases. I agree that otherwise unittests dont make much sense. Maybe you discuss this internally if you want to adopt tests on this level. --> FYI: No need to add unit tests for this PR if routing tests are not possible. As always, feel free to reach out to me if you want to discuss this in one of our meetings.

Building new routes in unit tests is not only tedious but it's also easy to get wrong, miss something or have subtle differences and we would mainly test axum. 
What we should unit test though is the implementation of the handlers, with mocked `UdsEcu` and write more integration tests. 
This is way out of scope for this mini PR. I will create a ticket which describes how we imagine testing the `cda-sovd` level. 
Also neither of these tests would have caught the missing attribute as we would have used the cda-sovd-interfaces types directly.

PR #123, opened 2025-12-04
log cda version at startup.

alexmohr:
## Summary
* add version function for opensovd-cda crate
* log version at startup

## Checklist

## Notes for Reviewers

---

### DIFF
--- a/cda-main/src/lib.rs
+++ b/cda-main/src/lib.rs
@@ -677,3 +677,9 @@ pub fn setup_tracing(config: &Configuration) -> Result<TracingGuards, TracingSet
         _otel: otel_guard,
     })
 }
+
+/// Retrieve the version of the opensovd-cda crate.
+#[must_use]
+pub fn cda_version() -> &'static str {
+    env!("CARGO_PKG_VERSION")
+}
--- a/cda-main/src/main.rs
+++ b/cda-main/src/main.rs
@@ -17,7 +17,7 @@ use cda_plugin_security::{DefaultSecurityPlugin, DefaultSecurityPluginData};
 use clap::Parser;
 use futures::future::FutureExt;
 use opensovd_cda_lib::{
-    AppError,
+    AppError, cda_version,
     config::configfile::{ConfigSanity, Configuration},
     setup_tracing, shutdown_signal,
 };
@@ -79,7 +79,7 @@ async fn main() -> Result<(), AppError> {
     args.update_config(&mut config);
 
     let _tracing_guards = setup_tracing(&config)?;
-    tracing::info!("Starting CDA...");
+    tracing::info!("Starting CDA - version {}", cda_version());
 
     let webserver_config = cda_sovd::WebServerConfig {
         host: config.server.address.clone(),


jleuser:
Not sure whether this MR is the right place, but would it make sense, to also log the URL to this repo/issue tracker to make it easy to file issues with the version? 
Assuming not everyone using the CDA is also building it him/herself and therefore might not be aware of where to report issues.

alexmohr:
We can create a [bug ticket link](https://github.com/eclipse-opensovd/classic-diagnostic-adapter/issues/new?template=bug_report.md) but the template does allow pre-filling versions (Without specifying everything via query parameters again)

There is not option to build [custom forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms) but this means versions must be pre-filled there. As the template is stored on the opensovd organization and not here this doesn't make much sense either. 

So I'd suggest to use the bug ticket link above and log it as separate line on info level. 
Something like 
```
Report issues at https://github.com/eclipse-opensovd/classic-diagnostic-adapter/issues/new?template=bug_report.md)
```

Alternative would be (I prefer that one)
```
Contribute at https://github.com/eclipse-opensovd/classic-diagnostic-adapter
```

lh-sag:
> > > Not sure whether this MR is the right place, but would it make sense, to also log the URL to this repo/issue tracker to make it easy to file issues with the version? Assuming not everyone using the CDA is also building it him/herself and therefore might not be aware of where to report issues.
> > 
> > 
> > I am hesitant to add such link in the binary/log. If someone did not build the binary him/herself, the bug should be given to the distributor of the binary. In the past I had bad experience e.g. received bugs from various vendors which contained closed source and hence where not reproducible and more. Especially with such a project which will not be used much outside of automotive & friends.
> > But this is just my opinion.
> 
> Hmm, thats a valid point. Would you rather avoid having something like this in the open source part at all, or is it rather not having the link to the opensource issue tracker in a potential oem binary? For the second case, we could have it be provided as a optional struct or string that is set when loading the cda library in the main.rs. If it is built as opensovd-cda from this source we would include the contribution hint to this repository - and oems can provide whatever contribution / issuetracker / etc. hint they want to have in their binary, if at all.

My personal experience is that none of the packaging/vendor folks ever changed the defaults e.g. nearly all GNU projects have e.g.  https://gcc.gnu.org/install/configure.html for bug and documentation (--with-bugurl=url). But they never get replaced...

If we provide defaults, expect them to get into the final product =D

alexmohr:
This is a very valuable discussion, but it moved into a way broader scope as originally anticipated with this. 
I would suggest to implement this as is for now, as having a function to get the CDA version will be useful anyway. Removing the log statement from main would be okay from my side as well. 

I will put an item on the agenda of the architecture meeting to discuss the topic raised by @jleuser. 
We can use the pipeline repo to define a shared text. For CI builds it's easy to pick this up, when building locally it would be great to find a solution where we don't have to define the text in every repository.

theswiftfox (approved):

PR #125, opened 2025-12-07
operations: fix linear and scale linear

alexmohr:
## Summary
fix broken calculation of scaling in scale linear
and add missing inverted direction.

## Checklist

## Notes for Reviewers

### DIFF
--- a/cda-core/src/diag_kernel/operations.rs
+++ b/cda-core/src/diag_kernel/operations.rs
@@ -147,10 +147,18 @@ fn compu_lookup(
                 compu_lookup_scale_linear(diag_type, lookup, scale)
             }
             datatypes::CompuCategory::TextTable => compu_lookup_text_table(scale),
-            datatypes::CompuCategory::CompuCode => todo!("CompuCode lookup not implemented"),
-            datatypes::CompuCategory::TabIntp => todo!("TabIntp lookup not implemented"),
-            datatypes::CompuCategory::RatFunc => todo!("RatFunc lookup not implemented"),
-            datatypes::CompuCategory::ScaleRatFunc => todo!("ScaleRatFunc lookup not implemented"),
+            datatypes::CompuCategory::CompuCode => Err(DiagServiceError::RequestNotSupported(
+                "compu_lookup for CompuCode is not implemented".to_owned(),
+            )),
+            datatypes::CompuCategory::TabIntp => Err(DiagServiceError::RequestNotSupported(
+                "compu_lookup for TabIntp is not implemented".to_owned(),
+            )),
+            datatypes::CompuCategory::RatFunc => Err(DiagServiceError::RequestNotSupported(
+                "compu_lookup for RatFunc is not implemented".to_owned(),
+            )),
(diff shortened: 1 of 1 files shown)


cda-core/src/diag_kernel/operations.rs
```diff
@@ -169,137 +350,180 @@ fn compu_convert(
...
+            }
+            DataType::Float32 => {
+                let v: f32 = value.try_into()?;
+                Ok(v.to_be_bytes().to_vec())
+            }
+            DataType::Float64 => {
+                let v: f64 = value.try_into()?;
+                Ok(v.to_be_bytes().to_vec())
+            }
+            _ => unreachable!(),
```
theswiftfox: suggestion: This `unreachable` statement could potentially be dangerous, as it expects the caller of the function to know that it should never be called with a different DataType than the numeric ones checked here, or am I missing something?  
What do you think about adding a sanity check for the passed in parameter to see if it is any of those 4 and otherwise returning an early error - then it would indeed be safe to use unreachable here, as it would never be possible to reach this.

alexmohr: Yes makes sense, I missed that while extracting this into a new function. Kept the unreachable here, but added the sanity as first thing in this function.

cda-core/src/diag_kernel/operations.rs
```diff
@@ -72,66 +141,16 @@ fn compu_lookup(
...
-            }
-            _ => {
-                todo!()
+                compu_lookup_scale_linear(diag_type, lookup, scale)
             }
+            datatypes::CompuCategory::TextTable => compu_lookup_text_table(scale),
+            datatypes::CompuCategory::CompuCode => todo!("CompuCode lookup not implemented"),
+            datatypes::CompuCategory::TabIntp => todo!("TabIntp lookup not implemented"),
+            datatypes::CompuCategory::RatFunc => todo!("RatFunc lookup not implemented"),
+            datatypes::CompuCategory::ScaleRatFunc => todo!("ScaleRatFunc lookup not implemented"),
```
theswiftfox: question (I know this PR didn't change the behavior here, just restructure it): should we return an error for those branches with the not impl message instead? (can be done in seperate issue, just want the opinion on this) Iirc the `todo!` would cause the tokio runtime thread to panic, which could have unintended side effects.

cda-core/src/diag_kernel/operations.rs
```diff
@@ -380,16 +632,17 @@ fn numeric_type_str_to_byte_vec(
     // (basically all integer types)
     value
         .split_whitespace()
+        .flat_map(|v| v.split(','))
```
theswiftfox: suggestion: combining the whitespace and ',' splitting into a single `.split([' ', ','])` expression.
I did a veery small test in a [rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=0dff7e1fc95f75f864a804d673744fe5) and it seems to do the same and perform similar within margin of runtime variance. And I think it improves readability quite a bit. 
Not sure if the input string I used is representative of what we try to achieve here, so if you take it into consideration please check if my assumption about performance is correct.

PR #129, opened 2025-12-12
ecu-manager: improve payload parsing

alexmohr:
## Summary
* ignore garbage bytes at end of pdu
* catch edge case where no data was consumed at end of pdu causing infinite loop.

This was noticed when a simulated ECU returned garbage bytes at end of PDU, which is okay according to the spec, but was not properly handled. 

## Checklist

## Notes for Reviewers

---

### DIFF
--- a/cda-core/src/diag_kernel/ecumanager.rs
+++ b/cda-core/src/diag_kernel/ecumanager.rs
@@ -2719,8 +2719,8 @@ impl<S: SecurityPlugin> EcuManager<S> {
 
         for i in 0..static_field_dop.fixed_number_of_items() {
             let param_byte_pos = param.byte_position();
-            let start = (param_byte_pos
-                .saturating_add(i.saturating_mul(static_field_dop.item_byte_size())))
+            let start = param_byte_pos
+                .saturating_add(i.saturating_mul(static_field_dop.item_byte_size()))
                 as usize;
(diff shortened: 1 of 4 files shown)


cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -2860,10 +2862,15 @@ impl<S: SecurityPlugin> EcuManager<S> {
...
                 }
             }
-            uds_payload.consume();
+            let consumed = uds_payload.consume();
             uds_payload.pop_slice()?;
             if uds_payload.exhausted() {
                 break;
+            } else if consumed == 0 {
+                return Err(DiagServiceError::BadPayload(
+                    "EndOfPdu did not consume any bytes, breaking potential infinite loop"
```
theswiftfox: Question: Can you elaborate under what conditions exhausted is false and nothing can be consumed? Is this the case if we cannot build a complete struct again with the remaining bytes in the payload, or something completely different?

alexmohr: Doesn't happen anymore at the moment in combination with the other changes I made, but this prevents endless loops in case no data is consumed if we would introduce a new issue later on. Parsing will still fail but at least it won't lock up the CDA

theswiftfox (approved):

PR #131, opened 2025-12-17
fix: add default NRC handler for unknown negative response codes

### LINKED ISSUE #60: Unknown NRC handling
Currently, when a NRC isn't defined as possible result for service, the CDA returns:
```json
{
  "message": "UDS lookup error: No matching response found for Service ID 127",
  "error_code": "vendor-specific",
  "vendor_code": "not-found"
}
``` 

If the specific NRC isn't defined, we should use a default nrc handler, that returns for all standardized NRCs a response that is compatible to our regular NRC error response.

ANAVHEOBA:
fix: add default NRC handler for unknown negative response codes

  Currently when an ECU returns a negative response code that isn't explicitly
  defined in the database, we just throw a generic "lookup error" which isn't
  very helpful for debugging. This change makes the CDA smarter about handling
  these cases - it now parses the raw UDS response and looks up the NRC in the
  ISO 14229 standard table to give a proper description.

  Fixes eclipse-opensovd/classic-diagnostic-adapter#60

  

  ## Summary

  Added fallback NRC handling when database mappings are missing. Instead of
  failing with a vague error, we now extract the NRC byte from the raw response
  and return a proper description using the ISO 14229 standard codes.

  For example, an unmapped NRC 0x33 will now return "Security Access Denied"
  instead of "UDS lookup error: No matching response found".

  ## Checklist

  ## Related

  Fixes #60

  ## Notes for Reviewers

  The main changes are in `diagservices.rs`:
  - Modified `as_nrc()` to fall back to raw data parsing when `mapped_data` is None
  - Added `extract_nrc_from_raw_data()` helper function
  - Added 7 unit tests covering various NRC values and edge cases

### DIFF
--- a/cda-core/src/diag_kernel/diagservices.rs
+++ b/cda-core/src/diag_kernel/diagservices.rs
@@ -21,7 +21,7 @@ use cda_interfaces::{
     },
 };
 
-use crate::diag_kernel::{DiagDataValue, operations};
+use crate::diag_kernel::{DiagDataValue, iso_14229_nrc, operations};
 
 #[derive(Debug, Clone)]
 pub struct DiagServiceResponseStruct {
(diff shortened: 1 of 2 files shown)


cda-core/src/diag_kernel/diagservices.rs
```diff
@@ -372,3 +373,131 @@ impl DiagServiceResponseStruct {
...
+        let data = vec![0x7F, 0x19, 0x10];
+        let result = extract_nrc_from_raw_data(&data).unwrap();
+
+        assert_eq!(result.code, Some(0x10));
+        assert_eq!(result.sid, Some(0x19));
+        assert_eq!(result.description, Some("General Reject".to_owned()));
+    }
+
+    #[test]
+    fn test_extract_nrc_from_raw_data_response_pending() {
```
alexmohr: suggestion: The first three tests added here, do test pretty much the same thing with different nrcs. I would prefer only keeping one of them, as the other tests do not add any real value imho.

alexmohr (changes requested):

Issue #132, opened 2025-12-17
exploration: enable webserver spawn without database and ecumanager


alexmohr:
## Description

Spawning the webserver before the ecu manager and database is ready.
This makes custom routes more useful, as they can provide data even before the server is fully ready. 

## Motivation

* Improves custom routes.
* Enabled refactoring of health endpoint into main webserver

## Additional Context

> [!IMPORTANT]
> Make sure to link this issue with the PR for your improvement.

PR #133, opened 2025-12-17
feature: Spawn server early, add routes later (#132) 

### LINKED ISSUE #118: Improvement: webserver custom routes
## Description

The webserver custom routes added in #111 only support basic routes, without access to the internal sovd server state and without a custom state. 
This makes it possible to register simple endpoints, but custom endpoints that need access to ecu functionality or hold OEM states are not supported right now. 

In the context of this issue the RouteProvider trait should be extended (or a new trait added) to support routes with access to the ECU or a custom state.

## Motivation

This makes the custom routes more useful and even user defined ecu endpoints can be supported, for example to support custom flashing logic. 

## Proposed Solution

See above. 
An alternative could be to rework the API as suggested by @lh-sag [here](https://github.com/eclipse-opensovd/classic-diagnostic-adapter/pull/111#pullrequestreview-3528802265) 
Before the API is changed, it should be discussed and aligned here and in the workstream meeting. 

## Additional Context

> [!IMPORTANT]

alexmohr:
## Summary
This reworks the startup of the webserver.
Routes can now be added at runtime instead only during startup. For example, this enables us to provide a health endpoint, which is available early on the same webserver as the normal operation.

It resulted from the discussion in #130. 
From a security point of view it _might_ make sense to have the health endpoint listening on a different port. 
I would like a discussion about how we're going to proceed with #130 and what's the best approach. 
Both variants have their own pros and cons. 
In any case having this particular PR makes sense in my opinion because at least the static routes are available right away and already can provide useful information. 

As a side effect the custom endpoint traits is not needed anymore. When constructing the webserver, we get the dynamic router back. This can now be used to inject custom routes either before or after the vehicle routes have been added, giving maximum flexibility.

Furthermore this change implements a second commit that fixes #118. The construction of webserver data has been split up. The new `VehicleData` struct contains everything that is necessary to provide custom vehicle endpoints, as all types in this struct implement clone. 
It is necessary though to provide a separate webserver state if this is needed, as the sovd webserver state is not exposed. 

## Checklist

## Related

Solves #132, #118

## Notes for Reviewers

### DIFF
--- a/cda-main/src/lib.rs
+++ b/cda-main/src/lib.rs
@@ -25,6 +25,7 @@ use cda_interfaces::{
     file_manager::{Chunk, ChunkType},
 };
 use cda_plugin_security::SecurityPlugin;
+use cda_sovd::Locks;
 use tokio::{
     signal,
     sync::{RwLock, mpsc},
@@ -55,6 +56,7 @@ type LoadedEcuMap<S> = HashMap<String, (EcuManager<S>, EcuMetadata)>;
 pub struct VehicleData<S: SecurityPlugin> {
     pub file_managers: FileManagerMap,
     pub uds_manager: UdsManagerType<S>,
+    pub locks: Arc<cda_sovd::Locks>,
 }
 
 /// Loads vehicle databases and sets up SOVD routes in the webserver.
@@ -110,9 +112,11 @@ pub async fn load_vehicle_data<
         vdetect.start_variant_detection().await;
     });
 
+    let ecu_names = uds.get_ecus().await;
     Ok(VehicleData {
         uds_manager: uds,
         file_managers,
+        locks: Arc::new(Locks::new(ecu_names)),
     })
 }
 
@@ -326,14 +330,15 @@ async fn load_database<S: SecurityPlugin>(
                     continue;
                 };
 
-                let ecu_payload: Vec<u8> = if let Some(payload) =
-                    ecu_data.into_iter().next().and_then(|c| c.payload)
-                {
-                    payload
-                } else {
-                    tracing::error!(ecu_name = %ecu_name, "No payload found in diagnostic description for ECU");
-                    continue;
-                };
+                let ecu_payload: Vec<u8> =
+                    if let Some(payload) = ecu_data.into_iter().next().and_then(|c| c.payload) {
+                        payload
+                    } else {
+                        tracing::error!(
+                        ecu_name = %ecu_name,
+                        "No payload found in diagnostic description for ECU");
+                        continue;
+                    };
 
                 let diag_data_base = match cda_database::datatypes::DiagnosticDatabase::new(
                     mdd_path.clone(),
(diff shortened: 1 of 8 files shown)


theswiftfox:
Nice rework 🤩 - from code review LGTM, but I want to run some tests to see if any unintended side effects occur with this change (don't expect so) before final approval :)

lh-sag:
I am not familiar with the code base so my comments may seem off.

Do we need to change routes dynamically? IMHO once the routes are setup once, they is no need to alter them e.g.

1. Setup SOVD routes

```
{base_uri}/components/
{base_uri}/components/{component_id}
```

2. Setup additional routes (optional)
```
{base_uri}/health
```

3. Serve http requests

e.g. loading the database can populate ECU dynamically but the routes are still static. Optional routes should not change either.

Any objections if we decouple the webserver from the diagnostic/database part?

alexmohr:
We cannot fully decouple it due to the ECU routes. They depend on which ecus have been loaded successfully and therefore we can add these routes only after loading the databases, so while the routes are static the ECUs are not.

lh-sag:
> > theory its straight forward as the webserver already uses the UdsManager to access the diagnostic layer and there we would only have to introduce the additional path parameter for the component name
> 
> I'm not sure that this will work well with the open API documentation and the swagger ui.
> 
> > We are adding a feature which is not needed at all because our architecture is like this =D
> 
> I beg to differ here, this change is removing more lines of codes than it's adding but providing more functionality.

My remark was not specifically to this PR. But all related to health, custom and dynamic endpoints.

> 
> > With decoupling the routes we do not need to handle /health, or /custom routes special and we have the flexibility of handling arbitrary topologies of ECUs, even dynamically.
> 
> That already works with this change, which is why I removed the custom endpoints trait. Health is another story because it might make sense to run that on a different port, but that's another discussion.
> 
> @lh-sag you seem to have a very strong opinion on the Webserver parts, as each change is turning into a fundamental discussion. Maybe we should dedicate a part of the next CDA meeting to discuss these things and align on the architecture?

Not intentionally and my feedback is rather overall architecture. The webserver happend to be the part I am most familiar. With different OEMs, we are differently opinionated. Feel free to add it to the agenda. I try to discuss not to force or gate the PR.

Just one last technical question, for the /data endpoint, do you want to statically register every data_id per ECU?

lh-sag:
> > I'm not sure that this will work well with the open API documentation and the swagger ui.
> 
> Yeah, we would only have a single generic description for the components and subpaths with a path param for the component id. While this would theoretically be fine for online description I think, it is still a bit of a drawback imho. Thats one of the reasons I still prefer this approach.

@theswiftfox Does this mean CDA supports /{any-path}/docs endpoint?

 FYI, I personally differentiate between two specs. OK if your definition differs =D:
* Online capability description as documented in 7.5 /{any-path}/docs
* Generic OpenAPI spec e.g. /openapi.json

The latter openapi spec will be used by e.g. openapi generators to generate clients and server skeletons. Personally I am not really interested in. For the former one I am =D

Issue #97, opened 2025-11-14
improvement: Add error message to DiagService::NotFound errors


alexmohr:
## Description

I propose adding error messages to all DiagService::NotFound errors. 

## Motivation

Especially during development it will be much easier to find out what went wrong and which lookup failed. 
At the moment it's necessary to step through to code with a debugger to find out where the CDA failed to find something

## Proposed Solution

Use DiagServiceError::NotFound with an error message and make this message mandatory instead of optional. 
This depends on #4

LucaFue:
I'd like to take this issue and start working on it. However, I currently dont have the rights to do so. Can you please assign me to the issue? Thanks!

alexmohr:
Make sure to sign the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/eca/), so we're able to merge your PR once you're ready :)

alexmohr:
Implemented in #259

PR #140, opened 2026-01-14
functional ECU communication

theswiftfox:
## Summary

This PR introduces changes to support functional communication with ECUs. To to so support for reading of a functional description database is added. Additionally the UDS communication layer is extended with a functional send method, that takes care of creating the request and waiting for all expected responses.
On the SOVD layer additional routes are added according to the specification in architecture. The responses contain the collection of all ecu responses and errors that are part of the functional group the request is made for.

## Notes for Reviewers

~~This PR is currently a draft, as some further additions / changes are expected and some todos need resolving. But due to the volume of the changes I wanted to already provide it as is.~~
PR will be squashed into single commit in the end.

---

### DIFF
--- a/integration-tests/tests/sovd/locks.rs
+++ b/integration-tests/tests/sovd/locks.rs
@@ -35,7 +35,7 @@ async fn lock_unlock() -> Result<(), TestingError> {
     let (runtime, _lock) = setup_integration_test(true).await?;
     let auth = auth_header(&runtime.config, None).await?;
 
-    for endpoint in endpoints() {
+    for endpoint in ENDPOINTS {
         // Check if the lock is created successfully and deleted after the timeout
         {
             let expiration_timeout = Duration::from_secs(2);
@@ -151,7 +151,7 @@ async fn cannot_lock_ecu_with_existing_functional_log() -> Result<(), TestingErr
 
     create_lock(
         default_timeout(),
-        &ecu_endpoint(),
+        ECU_ENDPOINT,
         StatusCode::CONFLICT,
         &runtime.config,
         &auth,
@@ -188,7 +188,7 @@ async fn ownership() -> Result<(), TestingError> {
     let auth_owner = auth_header(&runtime.config, None).await?;
     let auth_other = auth_header(&runtime.config, Some("ownership-test")).await?;
 
-    for endpoint in endpoints() {
+    for endpoint in ENDPOINTS {
         let lock_id: String = response_to_json_to_field(
             &create_lock(
                 default_timeout(),
@@ -333,7 +333,7 @@ async fn test_vehicle_lock_delete_hierarchy() -> Result<(), TestingError> {
         let ecu_lock_id: String = response_to_json_to_field(
             &create_lock(
                 default_timeout(),
-                &ecu_endpoint(),
+                ECU_ENDPOINT,
                 StatusCode::CREATED,
                 &runtime.config,
                 user,
@@ -369,7 +369,7 @@ async fn test_vehicle_lock_delete_hierarchy() -> Result<(), TestingError> {
         runtime: &TestRuntime,
     ) {
         lock_operation(
-            &ecu_endpoint(),
+            ECU_ENDPOINT,
             Some(ecu_lock_id),
             &runtime.config,
             user,
@@ -504,25 +504,16 @@ async fn test_vehicle_lock_cannot_be_deleted_by_non_owner() -> Result<(), Testin
 pub(crate) const FUNCTIONAL_GROUP_ENDPOINT: &str =
     "functions/functionalgroups/fgl_uds_ethernet_doip_dobt/locks";
 
-pub(crate) fn ecu_endpoint() -> String {
-    format!("{}/locks", sovd::ECU_FLXC1000_ENDPOINT)
-}
+pub(crate) const ECU_ENDPOINT: &str =
+    const_format::formatcp!("{}/locks", sovd::ECU_FLXC1000_ENDPOINT);
+
 pub(crate) const VEHICLE_ENDPOINT: &str = "locks";
 
 #[cfg(feature = "functional-locks-tests")]
-pub(crate) fn endpoints() -> [&'static str; 3] {
-    // leak ecu_endpoint into static str - this is fine for integration test but should
-    // never be done in the regular application
-    let ecu_endpoint: &'static str = Box::leak(ecu_endpoint().into_boxed_str());
-    [FUNCTIONAL_GROUP_ENDPOINT, VEHICLE_ENDPOINT, ecu_endpoint]
-}
+pub(crate) const ENDPOINTS: [&str; 3] = [FUNCTIONAL_GROUP_ENDPOINT, VEHICLE_ENDPOINT, ECU_ENDPOINT];
+
 #[cfg(not(feature = "functional-locks-tests"))]
-pub(crate) fn endpoints() -> [&'static str; 2] {
-    // leak ecu_endpoint into static str - this is fine for integration test but should
-    // never be done in the regular application
-    let ecu_endpoint: &'static str = Box::leak(ecu_endpoint().into_boxed_str());
-    [VEHICLE_ENDPOINT, ecu_endpoint]
-}
+pub(crate) const ENDPOINTS: [&str; 2] = [VEHICLE_ENDPOINT, ECU_ENDPOINT];
 
 pub(crate) async fn lock_operation(
     endpoint: &str,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -585,6 +585,26 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2459377285ad874054d797f3ccebf984978aa39129f6eafde5cdc8315b612f8"
 
 [[package]]
+name = "const_format"
+version = "0.2.35"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7faa7469a93a566e9ccc1c73fe783b4a65c274c5ace346038dca9c39fe0030ad"
+dependencies = [
+ "const_format_proc_macros",
+]
+
+[[package]]
+name = "const_format_proc_macros"
+version = "0.2.34"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1d57c2eccfb16dbac1f4e61e206105db5820c9d26c3c472bc17c774259ef7744"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "unicode-xid",
+]
+
+[[package]]
 name = "core-foundation"
 version = "0.9.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1613,6 +1633,7 @@ dependencies = [
  "cda-sovd",
  "cda-tracing",
  "chrono",
+ "const_format",
  "futures",
  "http",
  "libc",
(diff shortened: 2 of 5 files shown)


cda-comm-doip/src/lib.rs
```diff
@@ -541,6 +541,179 @@ impl<T: EcuAddressProvider + DoipComParamProvider> EcuGateway for DoipDiagGatewa
...
+            transmission_params.repeat_request_count_transmission,
+        )
+        .await?;
+
+        drop(ecu); // release lock before waiting for responses
+        ecu_mtxs.push((
+            transmission_params.ecu_name.to_lowercase(),
+            Arc::clone(gateway_ecu),
+        ));
+        let received_responses: Arc<RwLock<HashMap<String, Result<DiagnosticMessage, EcuError>>>> =
```
alexmohr: nitpick: the RwLock could be a mutex, as .read() is never called. Does not matter from a performance point of view though.

cda-comm-doip/src/lib.rs
```diff
@@ -541,6 +541,179 @@ impl<T: EcuAddressProvider + DoipComParamProvider> EcuGateway for DoipDiagGatewa
...
+                    // Ignore other message types
+                    continue;
+                }
+                Ok(Err(e)) => {
+                    return Some(Err(e));
+                }
+                Err(_) => {
+                    // Receiver closed
+                    return None;
+                }
```
alexmohr: question: does it make sense to return `DiagServiceError::Timeout` here instead? Then you would not need the loop in line 674

cda-sovd-interfaces/src/components/functional_group.rs
theswiftfox: todo: align naming. either all should be `functional_group` or `functional_groups`

cda-comm-uds/src/lib.rs
```diff
@@ -1888,6 +1979,157 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsEcu
...
+                            ecus: HashMap::from_iter([(logical_addr, ecu_name.clone())]),
+                        },
+                    ) {
+                        tracing::warn!(
+                            functional_group = %functional_group,
+                            gateway_addr = %gateway_addr,
+                            "Multiple Gateway ecus detected for functional group request."
+                        );
+                        // todo: how do we resolve this properly?
+                        // check which one is online?
```
alexmohr: we should be able to use the online as ECU, as the variant detection _should_ detect one correct ecu already. When we cannot find a unique database based on the name and revision we unload both databases. maybe we can do that for the logical address too?

Similar to https://github.com/eclipse-opensovd/classic-diagnostic-adapter/blob/main/cda-comm-uds/src/lib.rs#L1578

cda-comm-doip/src/lib.rs
theswiftfox: ooh good point. was a leftover as initially the final iteration was a read lock. but i changed that to `drain` and forgot to change this. Thanks :)

cda-sovd/src/sovd/functional_groups.rs
```diff
@@ -0,0 +1,874 @@
...
+                            include_schema,
+                        }
+                        .into_response();
+                    }
+                }
+            } else {
+                None
+            };
+
+            let map_to_json = match (accept.type_(), accept.subtype()) {
```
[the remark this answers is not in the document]
theswiftfox: good point. put this into a function now.

cda-comm-uds/src/lib.rs
theswiftfox: Thanks for the hint. I updated the code to only consider ecus that have the status "online" - I think that makes sense, as an offline / not detected / ... ECU is not expected to respond to the functional request. 
This also means that in the error case we now log an error as I think its not recoverable / resolvable if we get the state that two gateways with the same logical address are online at the same time, or is my assumption here wrong?
For that case we would just pick the first one we got and continue with the functional request and insert an error for the duplicated one into the responses.

cda-comm-doip/src/connections.rs
```diff
@@ -54,6 +55,29 @@ impl From<ConnectionError> for EcuError {
...
+                ConnectionError::Decoding(err) => DiagServiceError::DataError(DataParseError {
+                    value: err,
+                    details: String::new(),
+                }),
+                ConnectionError::InvalidMessage(msg) => {
+                    DiagServiceError::UnexpectedResponse(Some(msg))
+                }
+                ConnectionError::Timeout(_) => DiagServiceError::Timeout,
+                // map arbitrary connection errors to connection closed
+                ConnectionError::ConnectionFailed(_)
```
alexmohr: nitpick: should we add a string to connection closed, to provide a reason _why_ it's closed? Could be a separate issue though, similar to #97

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -1634,9 +1659,124 @@ impl<S: SecurityPlugin> EcuManager<S> {
...
         };
 
-        Ok(res)
+        let ecu_name = database
+            .ecu_data()?
+            .ecu_name()
+            .map(ToOwned::to_owned)
+            .ok_or_else(|| DiagServiceError::InvalidDatabase("ECU name not found".to_owned()))?;
+
+        Ok(Self {
```
alexmohr: nitpick: init the struct directly instead of using intermediate variables.
You can close this thread if you prefer the way it is :) (`new_ecu_description` would also be easier to read if initialized inline imho)
```rust

 Ok(Self {
        diag_database: database,
        db_cache: DbCache::default(),
        ecu_name,
        description_type: type_,
        database_naming_convention,
        tester_address: com_params.doip.logical_tester_address.default,
        logical_address: logical_ecu_address,
        logical_gateway_address: com_params.doip.logical_gateway_address.default,
        logical_functional_address: com_params.doip.logical_functional_address.default,
        nack_number_of_retries,
        diagnostic_ack_timeout: com_params.doip.diagnostic_ack_timeout.default,
        retry_period: com_params.doip.retry_period.default,
        routing_activation_timeout: com_params.doip.routing_activation_timeout.default,
        repeat_request_count_transmission: com_params.doip.repeat_request_count_transmission.default,
        connection_timeout: com_params.doip.connection_timeout.default,
        connection_retry_delay: com_params.doip.connection_retry_delay.default,
        connection_retry_attempts: com_params.doip.connection_retry_attempts.default,
        variant_detection: VariantDetection {
            diag_service_requests: HashSet::new(),
        },
        variant_index: None,
        variant: EcuVariant {
            name: None,
            is_base_variant: false,
            state: EcuState::NotTested,
            logical_address: logical_ecu_address,
        },
        duplicating_ecu_names: None,
        protocol,
        access_control: Arc::new(Mutex::new(SessionControl::default())),
        tester_present_retry_policy: com_params.uds.tester_present_retry_policy.default.clone().into(),
        tester_present_addr_mode: com_params.uds.tester_present_addr_mode.default.clone(),
        tester_present_response_expected: com_params.uds.tester_present_response_expected.default.clone().into(),
        tester_present_send_type: com_params.uds.tester_present_send_type.default.clone(),
        tester_present_message: com_params.uds.tester_present_message.default.clone(),
        tester_present_exp_pos_resp: com_params.uds.tester_present_exp_pos_resp.default.clone(),
        tester_present_exp_neg_resp: com_params.uds.tester_present_exp_neg_resp.default.clone(),
        tester_present_time: com_params.uds.tester_present_time.default,
        repeat_req_count_app: com_params.uds.repeat_req_count_app.default,
        rc_21_retry_policy: com_params.uds.rc_21_retry_policy.default.clone(),
        rc_21_completion_timeout: com_params.uds.rc_21_completion_timeout.default,
        rc_21_repeat_request_time: com_params.uds.rc_21_repeat_request_time.default,
        rc_78_retry_policy: com_params.uds.rc_78_retry_policy.default.clone(),
        rc_78_completion_timeout: com_params.uds.rc_78_completion_timeout.default,
        rc_78_timeout: com_params.uds.rc_78_timeout.default,
        rc_94_retry_policy: com_params.uds.rc_94_retry_policy.default.clone(),
        rc_94_completion_timeout: com_params.uds.rc_94_completion_timeout.default,
        rc_94_repeat_request_time: com_params.uds.rc_94_repeat_request_time.default,
        timeout_default: com_params.uds.timeout_default.default,
        security_plugin_phantom: std::marker::PhantomData::<S>,
    })

```

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -3545,7 +3685,7 @@ mod tests {
         let neg_response = {
             let nack_param = db_builder.create_coded_const_param(
                 "test_service_nack",
-                &NEGATIVE_RESPONSE.to_string(),
+                &service_ids::NEGATIVE_RESPONSE.to_string(),
```
alexmohr: nitpick: totally optional, but if you prefer prefixing with service_ids:: you could remove the import from the parent module and add the prefix to line 629 too :)

cda-interfaces/src/diagservices.rs
```diff
@@ -67,7 +67,7 @@ pub trait DiagServiceResponse: Sized + Send + Sync + 'static + Clone {
     fn get_dtcs(&self) -> Result<Vec<(DtcField, DtcRecord)>, DiagServiceError>;
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
alexmohr: question: Could we live w/o `Clone`? Seems to build without it and it might prevent us from calling `clone()' when there would be some other solution w/o a copy :)

integration-tests/tests/sovd/locks.rs
```diff
@@ -501,6 +509,21 @@ pub(crate) fn ecu_endpoint() -> String {
 }
 pub(crate) const VEHICLE_ENDPOINT: &str = "locks";
 
+#[cfg(feature = "functional-locks-tests")]
+pub(crate) fn endpoints() -> [&'static str; 3] {
+    // leak ecu_endpoint into static str - this is fine for integration test but should
+    // never be done in the regular application
+    let ecu_endpoint: &'static str = Box::leak(ecu_endpoint().into_boxed_str());
```
alexmohr: question: why did you opt for leaking into static &str instead of returning a `Vec<String>` instead? (Just out of curiosity)

cda-sovd/src/sovd/mod.rs
```diff
@@ -241,6 +231,10 @@ pub async fn route<
     );
 
     for ecu_name in ecu_names.drain(0..) {
+        // todo: use is_physical_ecu to skip functional description databases
+        if ecu_name == functional_group_config.description_database {
+            continue; // do not create a component for the functional description
+        }
```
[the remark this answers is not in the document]
theswiftfox: I think we can even simplify it further by using `uds.get_physical_ecus()` in line 171 instead.
As then we don't need any further checks here and functional group route creation fetches the ecus itself to get the functional ecu.

cda-sovd-interfaces/src/components/functional_groups.rs
```diff
@@ -0,0 +1,112 @@
...
+
+use crate::error::DataError;
+
+/// Response structure for functional group data operations
+/// Returns data for multiple ECUs with ECU names as top-level keys
+pub mod data {
+    use super::{DataError, Deserialize, HashMap, Serialize};
+
+    #[derive(Deserialize, schemars::JsonSchema)]
+    pub struct DataRequestPayload {
```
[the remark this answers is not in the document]
theswiftfox: Counter question 😂
I just saw that these docstrings are missing for the components aswell (cda-sovd-interfaces/src/components/ecu/mod.rs:157)
Considering this, do you prefer me adding them to both or creating an issue and doing that in a follow up PR?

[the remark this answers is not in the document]
theswiftfox: Decided to only add them here for now and create #144 for the rest, as I noticed its quite a few places that still need them added and did not want to blow up this PR.

cda-core/src/diag_kernel/ecumanager.rs
theswiftfox: changed both to inline initialization.

integration-tests/tests/sovd/locks.rs
theswiftfox: I decided to take option 3: added const_format and using that now to create the compile time constant for ECU_ENDPOINT instead of having it as a function. This in turn gets rid of the other functions as I could replace them with const arrays :)

PR #147, opened 2026-01-20
functional-communication: add config for protocol handling

theswiftfox:
## Summary

Follow up PR to #140 [regarding this discussion](https://github.com/eclipse-opensovd/classic-diagnostic-adapter/pull/140#discussion_r2698630969)
As agreed there this PR adds the functionality to configure if protocol for the functional group should be expected ad prefix or suffix. 
This also creates a basis for adding additional cases in the future if the need arises (for example omitting the protocol) as then the setting can simply be expanded.
Additionally it allows configuring if the protocol should be matched case sensitive or not

## Checklist

## Related

----

### DIFF
--- a/cda-core/src/diag_kernel/ecumanager.rs
+++ b/cda-core/src/diag_kernel/ecumanager.rs
@@ -1296,16 +1296,14 @@ impl<S: SecurityPlugin> cda_interfaces::EcuManager for EcuManager<S> {
                                 if self.fg_protocol_case_sensitive {
                                     name.starts_with(protocol_value)
                                 } else {
-                                    name.to_lowercase()
-                                        .starts_with(&protocol_value.to_lowercase())
+                                    util::starts_with_ignore_ascii_case(name, protocol_value)
                                 }
                             }
                             DiagnosticServiceAffixPosition::Suffix => {
                                 if self.fg_protocol_case_sensitive {
                                     name.ends_with(protocol_value)
                                 } else {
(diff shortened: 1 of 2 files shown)


cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -1286,7 +1290,26 @@ impl<S: SecurityPlugin> cda_interfaces::EcuManager for EcuManager<S> {
...
                     .and_then(|dl| dl.short_name())
                     .and_then(|name| {
-                        if name.ends_with(self.protocol.value()) {
+                        let protocol_value = self.protocol.value();
+                        let matches = match self.fg_protocol_position {
+                            DiagnosticServiceAffixPosition::Prefix => {
+                                if self.fg_protocol_case_sensitive {
+                                    name.starts_with(protocol_value)
+                                } else {
+                                    name.to_lowercase()
```
alexmohr: nitpick: to improve performance a tiny bit you can use 
```rust
starts_with_ignore_ascii_case
```
from utils.rs, as this does not allocate a new string.

Note: I'm not 100% sure this is a good idea if the name of the functional group does contain non ascii chars. On the other hand, we would have the problem in other places that use that function too. So we could use it here now and then fix it globally once the issue comes up?

alexmohr: awesome, thank you!

PR #154, opened 2026-01-29
improvement: Add error message to DiagService::NotFound errors

LucaFue:
## Summary

This PR is adding error messages to all DiagService::NotFound errors. This makes it easier to find out what went wrong and which lookup failed.

## Checklist

## Related

Fixes https://github.com/eclipse-opensovd/classic-diagnostic-adapter/issues/97


cda-comm-uds/src/lib.rs
```diff
@@ -136,9 +121,7 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsMana
     fn ecu_manager(&self, ecu_name: &str) -> Result<&RwLock<T>, DiagServiceError> {
         self.ecus
             .get(ecu_name)
-            .ok_or(DiagServiceError::NotFound(Some(format!(
-                "ECU {ecu_name} not found"
-            ))))
+            .ok_or_else(|| DiagServiceError::NotFound(Some(format!("ECU {ecu_name} not found"))))
```
theswiftfox: suggestion: we could change DiagServiceError::NotFound to always take a String instead of the Option<String> - then we don't need the wrapping into an option and enforce that it always has a message attached.

PR #162, opened 2026-02-02
core: Add public APIs to EcuManager for service parameter introspection, MUX case discovery and UDS request parsing

arvindr19:
Adding 3 new public APIs to EcuManager for service parameter introspection, MUX case discovery (discover service capabilities at runtime without parsing MDD files directly.) and parsing UDS request required for uds2sovdproxy.

1. get_service_parameter_metadata()
Purpose: Extract parameter metadata (CODED-CONST, PHYS-CONST, VALUE) from service definitions.

2. get_mux_cases_for_service()
Purpose: Retrieve MUX case information from service positive responses.

3. convert_request_from_uds()
Purpose: Parses incoming **REQUEST** payloads (not responses) using MDD REQUEST structure definitions.
- Takes raw UDS request bytes [SID, DID/Param..., DATA...]
- Parses according to MDD REQUEST definition
- Extracts all parameters with their types and values
- Returns JSON-mapped structure

## Checklist

### DIFF
--- a/cda-core/src/diag_kernel/ecumanager.rs
+++ b/cda-core/src/diag_kernel/ecumanager.rs
@@ -1410,6 +1410,233 @@ impl<S: SecurityPlugin> cda_interfaces::EcuManager for EcuManager<S> {
             response_type,
         })
     }
+    fn get_service_parameter_metadata(
+        &self,
+        service_name: &str,
+    ) -> Result<Vec<cda_interfaces::ServiceParameterMetadata>, DiagServiceError> {
+        use cda_interfaces::ServiceParameterMetadata;
+        fn extract_param_type_metadata(
+            param: &datatypes::Parameter<'_>,
+            service_name: &str,
+            name: &str,
+        ) -> Result<cda_interfaces::ParameterTypeMetadata, DiagServiceError> {
+            use cda_interfaces::ParameterTypeMetadata;
+
+            let param_type = match param.param_type()? {
+                datatypes::ParamType::CodedConst => param
+                    .specific_data_as_coded_const()
+                    .and_then(|cc| cc.coded_value())
+                    .map_or(ParameterTypeMetadata::Value, |v| {
+                        ParameterTypeMetadata::CodedConst {
+                            coded_value: v.to_owned(),
+                        }
+                    }),
+                datatypes::ParamType::PhysConst => param
+                    .specific_data_as_phys_const()
+                    .and_then(|pc| pc.phys_constant_value())
+                    .map_or_else(
+                        || {
+                            tracing::warn!(
+                                "Service '{}' param '{}' PHYS-CONST has no value",
+                                service_name,
+                                name
+                            );
+                            ParameterTypeMetadata::Value
+                        },
+                        |v| ParameterTypeMetadata::PhysConst {
+                            phys_constant_value: v.to_owned(),
+                        },
+                    ),
+                _ => ParameterTypeMetadata::Value,
+            };
+
+            Ok(param_type)
+        }
+
+        let service = self.get_meta_data_service(service_name)?;
+        let Some(request) = service.request() else {
+            tracing::warn!("Service '{}' has no request definition", service_name);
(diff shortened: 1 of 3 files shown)


cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -3590,6 +3806,98 @@ fn create_diag_service_response(
...
 }
 
+impl<S: SecurityPlugin> EcuManager<S> {
+    /// Convert a UDS REQUEST payload into a `DiagServiceResponse` using the
+    /// REQUEST definition in MDD
+    ///
+    /// # Errors
+    /// Returns `Err` if the payload doesn’t match the request definition, or if
+    /// elements cannot be correctly mapped from the raw data.
+    #[allow(clippy::too_many_lines)]
```
alexmohr: suggestion: please provide a reason as comment why.

cda-interfaces/src/ecumanager.rs
```diff
@@ -25,6 +25,41 @@ use crate::{
...
+    CodedConst { coded_value: String },
+    /// PHYS-CONST parameter with constant value from MDD
+    PhysConst { phys_constant_value: String },
+    /// VALUE or other dynamic parameter types
+    Value,
+}
+
+/// MUX case information for service response routing
+#[derive(Debug, Clone, Serialize)]
+pub struct MuxCaseInfo {
```
alexmohr: question: This is likely consistent through the whole database? Does it makes sense to have this configurable via data base naming conventions?

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -1417,6 +1417,222 @@ impl<S: SecurityPlugin> cda_interfaces::EcuManager for EcuManager<S> {
...
+                                    "Service '{}' param '{}' PHYS-CONSTANT-VALUE={}",
+                                    service_name,
+                                    name,
+                                    phys_constant_value
+                                );
+                                ParameterTypeMetadata::PhysConst {
+                                    phys_constant_value: phys_constant_value.to_owned(),
+                                }
+                            } else {
+                                tracing::debug!(
```
alexmohr: question: does this merit a warning?

cda-interfaces/src/ecumanager.rs
arvindr19: The naming patterns for MUX cases follow ISO 22901-1 (ODX) standard conventions and are standardized.
I feel, configurable via data base naming conventions unnecessary

alexmohr:
>  The proxy uses this as a fallback when MUX cases don't exist - it scans parameters looking for DID identifiers (via semantic tags or name patterns), then extracts the actual DID value from either coded_value (for CODED-CONST) or phys_constant_value (for PHYS-CONST). This handles both hardcoded DIDs and physical constant DIDs that aren't in MUX structures.

Is that something we could also move to the CDA? Not sure if we need it but it sounds like this could be common to both applications too.

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -1397,6 +1397,352 @@ impl<S: SecurityPlugin> cda_interfaces::EcuManager for EcuManager<S> {
...
+            return Ok(metadata);
+        };
+
+        tracing::debug!(
+            "Service '{}' has {} request parameters",
+            service_name,
+            params.len()
+        );
+
+        {
```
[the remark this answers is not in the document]
arvindr19: thank you for suggestion, I'll check and add these changes.

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -46,6 +46,12 @@ use crate::{
     },
 };
 
+const SERVICE_IDS_PARAMETER_META_DATA: [u8; 3] = [
```
alexmohr: suggestion: move this into `cda-interfaces/src/lib.rs`, to the other constants related to service ids.

PR #209, opened 2026-02-20
ecumanager: handle None input data correctly

theswiftfox:
## Summary

- Changes behavior if input data is None to treat is as empty Map. This allows us to return an error if mandatory parameters are missing.
- Adds fallback handling for phys-const if value was not provided
- checks for default value if no input was provided for a value param. In case no default is present returns error that the param is missing

## Checklist

## Notes for Reviewers

---


theswiftfox:
Note for self: probably makes sense to have this after #212 as it will otherwise be rebase horror

theswiftfox:
I realized the default values were simply read as JSON string..updated the code to read the default values in the same way as the coded_consts with #212

Issue #226, opened 2026-03-03
[Bug] Diagnostic service name is split after first delimiter


lorinciubeica10:
## Description

 Diagnostic service name is split after first  _ delimiter

## Steps to Reproduce

1. Run CDA
2. Variant identication is running

## Expected Behavior

Just the suffix is removed from the diagnostic service name

## Actual Behavior

Diagnostic service name contains just the first word from the full name of the service  -> after that, variant identication will fail because service name cannot be matched with the database

## Screenshots or Logs (if applicable)

Add logs, screenshots, or crash traces that help explain the problem.

## Environment

- Ubuntu 22.04

## Additional Context

> [!IMPORTANT]
> Link the pull request that addresses this bug once created.

PR #227, opened 2026-03-03
variant-detection: split diagnostic service name before last delimiter

lorinciubeica10:
## Summary

Diagnostic service name is split after first _ delimiter determining a failure in the variant identication because of the name mismatch between processed name and database name

## Checklist

## Related

Fixes https://github.com/eclipse-opensovd/classic-diagnostic-adapter/issues/226

### DIFF
--- a/cda-comm-uds/src/lib.rs
+++ b/cda-comm-uds/src/lib.rs
@@ -18,9 +18,9 @@ use std::{
 
 use async_trait::async_trait;
 use cda_interfaces::{
-    DiagComm, DiagCommType, DiagServiceError, DynamicPlugin, EcuGateway, EcuManager, EcuState,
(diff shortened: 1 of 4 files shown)


cda-comm-uds/src/lib.rs
```diff
@@ -1810,7 +1810,7 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsEcu
             .iter()
             .map(|req| {
                 let name = req
-                    .split_once('_')
+                    .rsplit_once('_')
```
alexmohr: suggestion: Instead of using split, the variant names should be cleaned in ecu manager using `self.database_naming_convention.trim_short_name_affixes()` when building the variant detection data.
This still breaks when the Service name is using a prefix instead of a suffix.

PR #240, opened 2026-03-07
memory-map FlatBuffers data to reduce RSS

alexmohr:
## Summary
Memory-map FlatBuffers data to reduce RSS by \~75% under memory pressure via zero-copy mmap with two-phase `madvise` hints. MDD files are memory-mapped directly and decompressed chunks are rewritten in-place as uncompressed, allowing the OS to page out unused regions automatically. Resident memory becomes proportional to the working set rather than total database size.

### Implementation Details

**Zero-Copy Memory Mapping:**
- MDD files are memory-mapped directly (no heap allocation) 
- First access to each MDD triggers in-place rewrite: compressed chunks (defaults to 'off')

#### RAM

**Setup:**
- 68 loaded databases, \~47 MB compressed on disk
- Linux 6.18.2-arch2-1, 32GB RAM, i5-7200U CPU
- Rust 1.92.0, `--release` profile
- No Simulation running
- Idle measured after \~3m runtime
- Swap disabled

**Results:**

| Configuration | Strategy | RSS (Idle) | RSS (Under Pressure) | Disk Usage | Reduction vs Baseline |
| --- | --- | --- | --- | --- | --- |
| **Baseline** | Heap allocation | 486,900 KB | 469,320 KB | 47 MB | — |
| **Decompressed MDD Data** | **In-place rewrite** | **152,780 KB** | **118,988 KB** | **242 MB** | Idle: −334,120 KB (−68.6%), Pressure: −350,332 KB (−74.7%) |
| Sidecar files | Persistent `.fb` files | 307,552 KB | 171,904 KB | \~282 MB (47 MB + 235 MB) | Idle: −179,348 KB (−36.8%), Pressure: −297,416 KB (−63.4%) |

### CPU

**Setup:**
* Start CDA, loaded 68 databases
* Wait 20s
* Attach perf
* Wait 10s 
* Fill memory with garbage data
* Flash an ECU

**Results:**

| Metric | Main | MMap | Change | Winner |
|--------|------|------|--------|--------|
| **cycles** | 448.5M | 437.1M | **-2.5%** | MMap |
| **instructions** | 194.9M | 194.3M | **-0.3%** | MMap |
| **faults** | 340 | 1,206 | **+255%**  | Main |
| **cache-refs** | 25.8M | 26.0M | +0.7% | Main |
| **cache-misses** | 18.1M (70.21%) | 18.5M (70.98%) | **-7% miss rate** | Main |
| **wall time** | 42.19s | 42.19s | negligible | tie |

**Analysis:**

*Both implementations are essentially equivalent under realistic load
* 2.5% difference is negligible as only a few runs where conducted. 
* 340 faults vs 1,206 (255% more for MMap, to be expected and doesn't have an impact on performance at the scale of the test)

### Architecture Decision Record

Full analysis documented in ADR-002: MDD Memory-Mapping Strategy
- Decision rationale: In-place rewrite vs sidecar files vs two-phase madvise
- Benchmarks showing around 75% RSS reduction under pressure

## Checklist

- [x] I have tested my changes locally (benchmarks run, RSS verified)
- [x] I have added or updated documentation (ADR-002)

## Notes for Reviewers

For anyone who read the PR already:
I redid the tests on Linux and disabled swap, because MacOS is just too good at compressing memory and swapping, so it is not representative at all (except you're running your onboard system on a Mac I guess).
The new measurements yields numbers way closer to reality.
Although they a reduction by \~75% isn't as amazing as 90%, it's still a great improvement.
History is squashed the sidecar impl can be found on https://github.com/eclipse-opensovd/classic-diagnostic-adapter/commit/d37f213d350a0bbffb673bf955688654d8639da9

### Todo
- [x] Cleanup code
- [x] Performance tests
- [x] Cleanup weirdness for some measurements, re-do for current impl and base line

**Safety:**
- SHA-512 verification ensures integrity after rewrite
- Atomic write (temp file + rename) prevents corruption
- Mmap safety: FlatBuffers never mutates mapped data

**Trade-offs:**
- First load per MDD writes file -> Should be moved to MDD update plugin eventually
- Positive: Loading uncompressed MDDs files is slightly faster, so startup is sped up.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -400,15 +400,18 @@ dependencies = [
 name = "cda-database"
 version = "0.1.0"
 dependencies = [
+ "bytes",
  "cargo_toml",
  "cda-build",
  "cda-interfaces",
  "flatbuffers",
+ "memmap2",
  "ouroboros",
  "prost",
  "prost-build",
  "serde",
  "serde_json",
+ "sha2",
  "tokio",
  "toml 0.9.8",
  "tracing",
@@ -438,6 +441,7 @@ name = "cda-interfaces"
 version = "0.1.0"
 dependencies = [
  "async-trait",
+ "bytes",
  "foldhash",
  "hex",
  "mockall",
@@ -1971,6 +1975,15 @@ name = "memchr"
 version = "2.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f52b00d39961fc5b2736ea853c9cc86238e165017a493d1d5c8eac6bdc4cc273"
+
+[[package]]
+name = "memmap2"
+version = "0.9.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "714098028fe011992e1c3962653c96b2d578c4b4bce9036e15ff220319b1e0e3"
+dependencies = [
+ "libc",
+]
 
 [[package]]
 name = "mimalloc"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,7 @@ tokio-util = { version = "0.7.3", default-features = false }
 futures = "0.3"
 regex = "1.12.2"
 uuid = "1.18.1"
+bytes = "1"
 prost = "0.14"
 flatbuffers = "25.9.23"
 xz2 = "0.1.7"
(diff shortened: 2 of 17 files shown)


alexmohr:
I'll look into rewriting the mdd files but adding 'side car' files for now was easier to play around with this more. Path is configurable. This removes the tempfile and the mmapped file limit and all flatbuffer data is loaded from the sidecar file instead, as this allows us to memory map all files without a lot of overhead, we get even better memory usage, between 54 and 68 on my system, it depends a bit on the memory pressure the system has. 
I didn't implement any chunk validation yet.

floroks (approved):

Issue #260, opened 2026-03-20
[Bug] SendKey level parsing


lorinciubeica10:
## Description

When interpreting the SendKey security level, the implementation assumes that the level name consists of exactly two segments separated by a single underscore. As a result, any level name containing more than two underscore‑separated segments will automatically be misinterpreted: will treat it as if both a RequestSeed and a SendKey operation were being requested at the same time.

## Steps to Reproduce

1. Request a SendKey operation with a name containing more than two underscore-separated segments

## Expected Behavior

Positive response for this request. 

## Actual Behavior

Error occured: RequestSeed and SendKey cannot be used at the same time.

## Screenshots or Logs (if applicable)

Add logs, screenshots, or crash traces that help explain the problem.

## Environment

- OS: Ubuntu 22.04
- Version: main

## Additional Context

> [!IMPORTANT]
> Link the pull request that addresses this bug once created.

lorinciubeica10:
@alexmohr @floroks can I take this since the PR is inactive?

PR #261, opened 2026-03-20
fix: handle SendKey level names with multiple underscores

saschabuehrle:
## Bug
https://github.com/eclipse-opensovd/classic-diagnostic-adapter/issues/260 — security mode parsing treated any value with more than two underscore-separated segments as if a service suffix was present, which made valid SendKey requests fail with a mixed RequestSeed/SendKey error.

## Fix
This change now only treats `_RequestSeed` and `_SendKey` as explicit service suffixes. Other underscore segments stay part of the level name. `_SendKey` is normalized to a regular SendKey flow, so key-based requests no longer conflict with suffix parsing.

## Testing
Unable to run Rust tests in this environment because `cargo` is not installed. Verified the parser logic change in `cda-sovd/src/sovd/components/ecu/modes.rs` to ensure only known service suffixes are interpreted.

Happy to address any feedback.

Greetings, saschabuehrle

### DIFF
--- a/cda-sovd/src/sovd/components/ecu/modes.rs
+++ b/cda-sovd/src/sovd/components/ecu/modes.rs
@@ -421,17 +421,15 @@ pub(crate) mod security {
         >,
     ) -> Response {
         fn split_at_last_underscore(input: &str) -> (String, Option<String>) {
-            let parts: Vec<&str> = input.split('_').collect();
-
-            if parts.len() > 2 {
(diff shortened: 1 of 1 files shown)


alexmohr:
Closing this #370 implemented the necessary changes already. If this is not enough, #427 adds infrastructure to override certain functionality with custom logic so this be implemented OEM specific downstream.

PR #273, opened 2026-04-08
operations: add support for async operations

theswiftfox:
## Summary

This PR implement async operations for ECU and functional groups. It adds new unit and integration tests for the functionality.
For this the FLXC description as well as the ECU sim are extended in the addition to the CDA code changes.
Existing functionality for sync operations should continue to work as before this PR.

## Checklist

## Notes for Reviewers

---


cda-sovd-interfaces/src/functions/functional_groups/operations.rs
```diff
@@ -49,3 +56,18 @@ pub mod service {
 pub mod get {
     pub type Query = crate::IncludeSchemaQuery;
 }
+
+/// A single item in the `GET /functions/functionalgroups/{group}/operations` collection.
+/// Spec Table 169 (`OperationDescription`).
+#[derive(Serialize, Deserialize, schemars::JsonSchema)]
+pub struct FgOperationCollectionItem {
```
floroks: Wouldn't it make sense to reuse `OperationCollectionItem`? I believe operations should handle the same, since they are both part of the standardized ecu resource collection.

theswiftfox: Absolutely!
I combined them into the shared module and reduced the amount of duplication now :)

cda-sovd/src/sovd/components/ecu/operations.rs
```diff
@@ -626,84 +753,315 @@ pub(crate) mod service {
...
+                    schema: None,
+                },
             }
+        }
 
-            if map_to_json {
-                let (mapped_data, errors) = match response.into_json() {
+        /// Parses a positive, non-empty UDS response into `(parameters, errors)`.
+        ///
+        /// - `Object` → parameters map + any field-parse errors
```
alexmohr: suggestion: Replace unicode arrows with '->' right away, so we don't have to fix these later when implementing #275 
There are multiple of this, only posting it once though :)

cda-sovd/src/sovd/components/ecu/operations.rs
```diff
@@ -885,11 +1252,304 @@ pub(crate) mod service {
...
+                    .send(
+                        &ecu_name,
+                        diag_service,
+                        &(security_plugin as DynamicPlugin),
+                        None,
+                        true,
+                    )
+                    .await;
+
+                if let Some(exec) = service_executions.write().await.get_mut(&exec_id) {
```
alexmohr: question: Write lock is acquired here, then again in 1470, to call `shift_remove`, but in theory the service executions could be modified if the same `exec_id` is used in another call. probably unlikely but does it make sense to protect against this? If not commenting why would help

cda-sovd/src/sovd/components/ecu/operations.rs
```diff
@@ -419,93 +485,146 @@ pub(crate) mod service {
...
         pub(crate) async fn get<
             R: DiagServiceResponse,
             T: UdsEcu + SchemaProvider + Clone,
             U: FileManager,
         >(
             UseApi(Secured(_security_plugin), _): UseApi<Secured, ()>,
             WithRejection(Query(query), _): WithRejection<Query<sovd_executions::Query>, ApiError>,
-            State(_state): State<WebserverEcuState<R, T, U>>,
+            State(WebserverEcuState {
+                service_executions, ..
```
alexmohr, line 547: question: service_executions can grow indefinitely due to a buggy or malicious client. Does it make sense to add an upper bound here, then return `429 too many requests` once this is reached?

cda-sovd/src/sovd/components/ecu/operations.rs
```diff
@@ -626,84 +753,315 @@ pub(crate) mod service {
...
+                .into_response()
+        }
+
+        /// Handles the async (Stop/RequestResults) POST path: registers the execution,
+        /// returns 202 Accepted with only `id` and `status` per spec Table 184.
+        /// The Start response parameters are stored internally for later GET retrieval.
+        async fn handle_async_post<T: UdsEcu>(
+            response: Option<T::Response>,
+            map_to_json: bool,
+            _include_schema: bool,
```
alexmohr: question: why have this parameter when it's not used? Is this supposed to be added later?

cda-sovd/src/sovd/functions/functional_groups/operations.rs
```diff
@@ -47,11 +168,15 @@ pub(crate) mod diag_service {
         State(WebserverFgState {
             uds,
             functional_group_name,
+            fg_executions,
             ..
         }): State<WebserverFgState<T>>,
         body: Bytes,
     ) -> Response {
         let include_schema = query.include_schema;
+        let suppress_service = query.suppress_service;
```
alexmohr, line 279: question: The ECU level implementation _does_ validate the lock ownership, but this is skipped here, is this on purpose?

cda-sovd/src/sovd/functions/functional_groups/operations.rs
```diff
@@ -10,33 +10,154 @@
...
+                    errors,
+                    schema,
+                },
+            ),
+        )
+            .into_response()
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    #[allow(clippy::too_many_lines)]
```
alexmohr: nitpick: Give a reason for disabling these, as per code style disabling clippy warnings should be always. 

> Allow/Forbid: Use #[allow(...)] only when necessary, and always document the reason. 

Or alternatively split it up into smaller functions of course. (Although you'd still have to disable too many arguments.)

cda-sovd/src/sovd/functions/functional_groups/operations.rs
```diff
@@ -103,68 +361,111 @@ pub(crate) mod diag_service {
...
-        let data = match get_payload_data::<
-            sovd_interfaces::functions::functional_groups::operations::service::Request,
-        >(content_type.as_ref(), &headers, &body)
         {
-            Ok(value) => value,
-            Err(e) => {
-                return ErrorWrapper {
-                    error: e,
-                    include_schema,
+            let mut guard = fg_executions.write().await;
```
alexmohr: question: almost the same as the 'guard_execution' function. can they be combined?

cda-sovd/src/sovd/functions/functional_groups/operations.rs
```diff
@@ -175,26 +476,1061 @@ pub(crate) mod diag_service {
...
+            suppress_service: bool,
+        ) -> sovd_interfaces::components::ecu::operations::OperationDeleteQuery {
+            sovd_interfaces::components::ecu::operations::OperationDeleteQuery {
+                include_schema: false,
+                suppress_service,
+                force: false,
+            }
+        }
+
+        // ── POST: sync operations ─────────────────────────────────────────────
```
alexmohr: nitpick: I said it on another PR and this is pretty much my personal opinion but I want to say it again here: 
I know that LLMs loooove to generate these banner comments for whatever reason. In my personal opinion they are clutter and not helpful at all. 
Especially the
```
        // ── DELETE: Stop ──────────────────────────────────────────────────────
```
One doesn't even make that much sense. 
As there is no agreement against these feel free to keep them, but I'd prefer them gone 🙈

cda-sovd/src/sovd/functions/functional_groups/mod.rs
```diff
@@ -165,13 +170,24 @@ fn create_functional_group_route<T: UdsEcu + Clone>(fg_state: WebserverFgState<T
...
         .api_route(
             "/operations/{operation}",
             routing::post_with(
                 operations::diag_service::post,
                 operations::diag_service::docs_post,
             ),
         )
+        .api_route(
+            "/operations/{operation}/executions/{id}",
+            routing::delete_with(
```
alexmohr: question: Why is there no getter in functional groups, where ecu has one? It's weird because a client cannot fetch the status of an operation.
Note: I did not read the corresponding documentation yet, so it might be documented there, but then providing a back link or a comment would be helpful for the next person stumbling over this.

alexmohr (commented):
I didn't have the time to review all 6k LOC 😆, so I only looked at the parts I was curious about and skipped most of the tests, python and sim implementation so it's no way near a full review. I left some comments, mostly questions though :)

cda-sovd/src/sovd/functions/functional_groups/operations.rs
theswiftfox: Oh thanks. I missed that one -.-
They are god awful, but this one got me as it used a unicode dash instead of `-` 🙈

cda-sovd/src/sovd/functions/functional_groups/operations.rs
theswiftfox: oh, that was an oversight. i think i only fixed in ecu but not in functional_groups. thx! :)

cda-sovd/src/sovd/functions/functional_groups/operations.rs
theswiftfox: yeah, should be checked here too.

cda-sovd/src/sovd/components/ecu/operations.rs
theswiftfox: was an oversight. adding it to the response and using the parameter in this fn in the new commit :)

cda-sovd/src/sovd/components/ecu/operations.rs
theswiftfox: yes and yes. But probably not in the context of this MR. I think we have the same issue for setting comparams in theory - so I would like to have a well thought out common solution for this. We can put up an issue for this though so we keep it in mind.

cda-sovd/src/sovd/components/ecu/operations.rs
theswiftfox: oh, yea thats a bug introduced by me moving stuff around to reduce code. The in flight flag should only be reset in an else branch of `if query.force`. As otherwise the execution is removed anyways and it makes no sense to reset the flag before that. Thanks :)

cda-sovd/src/sovd/components/ecu/operations.rs
theswiftfox: Yeah, I think my editor did some funny thing there converting `->` into a unicode emoji, and I didn't notice due to me being used to ligatures. Would be really nice to have a lint for that one 🙈
I'm replacing them all with `->`, thanks!

cda-sovd/src/sovd/functions/functional_groups/mod.rs
theswiftfox: > question: Why is there no getter in functional groups, where ecu has one? It's weird because a client cannot fetch the status of an operation. Note: I did not read the corresponding documentation yet, so it might be documented there, but then providing a back link or a comment would be helpful for the next person stumbling over this.

I'll get back to you on that one - We should either implement it (+ integration test) or clarify in the docs if we do not plan to have RequestResults on a functional communication level. But I'll clarify which path we want to take.

cda-sovd/src/sovd/components/ecu/operations.rs
theswiftfox: It should be solved in the same way as for FG in the latest commit: only reset the flag in the branch that does not remove the execution. That means while the branch is not decided other request still get the conflict response into a 404 later on retries if the execution was removed.

cda-core/src/diag_kernel/ecumanager.rs
alexmohr, line 3464: optional: As you're changing stuff in ecumanager anyway, it would be great if you could fix the two clippy nightly warnings here, reported by reviewdog, optional as it's technically unrelated and only warned on nightly :)

cda-sovd/src/sovd/components/ecu/operations.rs
alexmohr: You'll create an issue?

cda-sovd/src/sovd/functions/functional_groups/operations.rs
```diff
@@ -175,26 +471,1099 @@ pub(crate) mod diag_service {
...
+            );
+            assert!(
+                result
+                    .get("parameters")
+                    .and_then(serde_json::Value::as_object)
+                    .is_none_or(serde_json::Map::is_empty)
+            );
+        }
+
+        // ── DELETE: Stop ──────────────────────────────────────────────────────
```
alexmohr: nitpick: Please remove this comment too.

theswiftfox: 🤦‍♀️ how did that one even end up there..
thanks, I'll remove it!

cda-sovd/src/sovd/components/ecu/operations.rs
theswiftfox: Yeah sure!

theswiftfox: #281 
Feel free to add any ideas there :)

alexmohr: Thanks :)

cda-core/src/diag_kernel/ecumanager.rs
theswiftfox: #283 
I'd rather do it in a small housekeeping PR

cda-sovd/src/sovd/functions/functional_groups/operations.rs
theswiftfox: removed.

cda-sovd/src/sovd/functions/functional_groups/mod.rs
theswiftfox: Will add the implementation for completeness as the standard in theory supports it.

cda-sovd/src/sovd/functions/functional_groups/operations.rs
theswiftfox: yeah, i moved it to a common helper that is used in both places.

cda-sovd/src/sovd/functions/functional_groups/mod.rs
theswiftfox: It is now implemented :)

theswiftfox:
Rebased on main and added an integration tests for functional group operations as well, now that we have the functional ecu description in the repo :)

cda-sovd/src/sovd/functions/functional_groups/operations.rs
```diff
@@ -103,68 +441,126 @@ pub(crate) mod diag_service {
...
                     include_schema,
                 }
                 .into_response();
             }
         };
 
-        let data = match get_payload_data::<
-            sovd_interfaces::functions::functional_groups::operations::service::Request,
-        >(content_type.as_ref(), &headers, &body)
+        // Guard: mark execution in_flight (replaces shared guard_execution for nested map).
```
alexmohr: question: Why is the shared execution not used here? The comments read like either you had a clear idea in mind or an ai left this during refactoring 🙈 
It's functionally correct and _slightly_ different compared to the shared execution it would help to document why this path was chosen instead of documenting what :)

theswiftfox: mh, i honestly cant tell you if that was an ai or me leaving that in as this part was changed so many times ^^
I originally had an idea to share this, then noticed problems in other locations, and rewrote it again ^^

I probably changed this back (without removing the comment) when separating the FgServiceExecution struct out, but I think if the guard helper is made generic over an additional parameter it can be shared again. I'll take a look tomorrow :)

cda-sovd/src/sovd/components/ecu/operations.rs
```diff
@@ -608,141 +752,349 @@ pub(crate) mod service {
...
+                    Ok(DiagServiceJsonResponse {
+                        data: serde_json::Value::Object(m),
+                        ..
+                    }) => m,
+                    _ => serde_json::Map::new(),
+                },
+                _ => serde_json::Map::new(),
+            };
+            let id = Uuid::new_v4();
+            service_executions
```
alexmohr: question: 
check_running_execution_conflict acquires a read lock, then drops it. The actual insertion in here acquires a write lock later. Between these two points, a concurrent request could also pass the conflict check. 

Could be fixed by just checking the again, although it is not likely to happen in the real world, with precise timing this could be a (minor, but hard to find) issue

Example (could be probably streamlined by extracting this into a helper)

```diff 

diff --git a/cda-sovd/src/sovd/components/ecu/operations.rs b/cda-sovd/src/sovd/components/ecu/operations.rs
index fdd76a1..d5a623f 100644
--- a/cda-sovd/src/sovd/components/ecu/operations.rs
+++ b/cda-sovd/src/sovd/components/ecu/operations.rs
@@ -954,9 +954,23 @@ pub(crate) mod service {
                 _ => serde_json::Map::new(),
             };
             let id = Uuid::new_v4();
-            service_executions
-                .write()
-                .await
+            let mut guard = service_executions.write().await;
+            // Re-check under the write lock to close the time of use / time of check gap between
+            // the read-lock conflict check and this insertion.
+            let has_running = guard.get(&service).is_some_and(|m| {
+                m.values()
+                    .any(|e| e.status == ExecutionStatus::Running)
+            });
+            if has_running {
+                return ErrorWrapper {
+                    error: ApiError::Conflict(format!(
+                        "An execution for operation '{service}' is already in progress"
+                    )),
+                    include_schema,
+                }
+                .into_response();
+            }
+            guard
                 .entry(service)
                 .or_default()
                 .insert(
@@ -967,6 +981,7 @@ pub(crate) mod service {
                         in_flight: false,
                     },
                 );
+            drop(guard);
             let schema = if include_schema {
                 Some(create_schema!(AsyncPostResponse))
             } else {
diff --git a/cda-sovd/src/sovd/functions/functional_groups/operations.rs b/cda-sovd/src/sovd/functions/functional_groups/operations.rs
index e4c6f41..8240722 100644
--- a/cda-sovd/src/sovd/functions/functional_groups/operations.rs
+++ b/cda-sovd/src/sovd/functions/functional_groups/operations.rs
@@ -165,9 +165,23 @@ pub(crate) mod diag_service {
     ) -> Response {
         let id = Uuid::new_v4();
         let operation = operation.to_lowercase();
-        fg_executions
-            .write()
-            .await
+        let mut guard = fg_executions.write().await;
+        // Re-check under the write lock to close the time of use / time of check gap between
+        // the read-lock conflict check and this insertion.
+        let has_running = guard.get(&operation).is_some_and(|m| {
+            m.values()
+                .any(|e| e.status == ExecutionStatus::Running)
+        });
+        if has_running {
+            return ErrorWrapper {
+                error: ApiError::Conflict(format!(
+                    "An execution for operation '{operation}' is already in progress"
+                )),
+                include_schema,
+            }
+            .into_response();
+        }
+        guard
             .entry(operation)
             .or_default()
             .insert(
@@ -178,6 +192,7 @@ pub(crate) mod diag_service {
                     in_flight: false,
                 },
             );
+        drop(guard);
 
         let exec_url = format!("http://{host}{uri}/executions/{id}");
         let schema = if include_schema {


```

theswiftfox: Good point, but as the write is only done **after** the request was sent to the ECU the suggested approach would depend on the ECU correctly rejecting the second start request. Ideally we do not want to even send another request in that case I think. At the same time we cannot hold the lock for too long (as it is implemented now) or otherwise we would block different, unrelated operations.

So my idea is to promote the read lock to a write lock directly after the conflict check and insert an execution there. If the ECU request then fails, we can remove the execution again before returning an error - that way we ensure no two start requests are sent.

Do you see any issues with that approach?

alexmohr: > Ideally we do not want to even send another request in that case I think.

Oh I missed that, thank you. 

> promote the read lock to a write lock directly after the conflict check and insert an execution there

Is it possible to promote the lock without releasing it first? and if not, how can we guarantee that the state isn't changed before we acquire the write lock?

cda-sovd/src/sovd/functions/functional_groups/operations.rs
theswiftfox: Reusing the helper again here now :)

cda-sovd/src/sovd/components/ecu/operations.rs
theswiftfox: > > Ideally we do not want to even send another request in that case I think.
> 
> Oh I missed that, thank you.
> 
> > promote the read lock to a write lock directly after the conflict check and insert an execution there
> 
> Is it possible to promote the lock without releasing it first? and if not, how can we guarantee that the state isn't changed before we acquire the write lock?

I decided to directly use a writelock for the conflict check on POST. we are holding the lock for a very short time only and not doing any I/O or similar, so it should be fine I think.

theswiftfox:
Added 2 new commits addressing the new comments. If they are resolved, I'll squash those into the main feature commit again

alexmohr (approved):
lgtm, thanks for bearing with me 😆

theswiftfox:
> lgtm, thanks for bearing with me 😆

Thank you for the time reviewing and the valuable feedback!! :)

Issue #275, opened 2026-04-10
Improvement: Ban unicode from comments


theswiftfox:
As noted by @floroks in https://github.com/eclipse-opensovd/classic-diagnostic-adapter/pull/267#discussion_r3018010216_ we should check if we can ban the use of unicode in comments via tooling.

alexmohr:
Taking this issue, I don't want to keep writing comments about not using unicode

alexmohr:
done

PR #282, opened 2026-04-15
diag-kernel: thread base offset through structure DOP decoding

### LINKED ISSUE #270: [Bug] Parsing issue for read data
## Description

Reading data after 46c71655203b377616bf5e2b87a7262da5c2692e commit is not working properly. 
Parsing of the payload and its interpretation is faulty. 

L.E.

Several bugs caused incorrect or missing data when decoding UDS responses containing nested structures, DTCs, and snapshot (EnvDataDesc) data:

- Parameters inside Structure DOPs were decoded using a relative byte_position instead of the absolute position in the full response payload, causing reads from the wrong offset.
- map_param_matching_request_from_uds used a relative byte position when calling push_slice, producing a wrong payload slice.
- map_dtc_dop_from_uds stored byte_position() in bit_pos (wrong field) and used a relative offset for byte_pos.
- map_dynamic_length_field_from_uds computed the wrong item count when a param had no explicit byte_position.
- EnvDataDesc decoding used an incorrect base_offset, causing the discriminator lookup to fail and snapshot data to be silently dropped.

## Steps to Reproduce

1. Request any read data 
2. Check the response data

## Expected Behavior

Response contains the correct parsed and mapped data.

## Actual Behavior

Incomplete response or errors occuring complaining of the length of the payload

[cda_log.txt](https://github.com/user-attachments/files/26301150/cda_log.txt)

Add logs, screenshots, or crash traces that help explain the problem.

## Environment

- OS: [e.g. Ubuntu 22.04, Windows 11]
- Version: [e.g. v0.2.1, main]

## Additional Context

> [!IMPORTANT]

lorinciubeica10:
## Summary

Replaced push_slice/pop_slice coordinate shifting in map_structure_dop_from_uds with a base offset parameter propagated through all decode functions. 
This correctly handles both absolute and relative byte position conventions without corrupting the payload cursor for inner params
## Checklist

## Related

Fixes https://github.com/eclipse-opensovd/classic-diagnostic-adapter/issues/270

### DIFF
--- a/cda-comm-uds/src/lib.rs
+++ b/cda-comm-uds/src/lib.rs
@@ -912,6 +912,7 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsMana
         dtc_code: DtcCode,
         include_schema: bool,
         memory_selection: Option<u8>,
+        scope: &str,
     ) -> Result<(Option<ExtendedDataRecords>, Option<serde_json::Value>), DiagServiceError> {
         fn extract_schema_properties(schema_desc: &SchemaDescription) -> Option<serde_json::Value> {
             // todo after solving #54: we are missing the 'Selector' and the case name here
@@ -924,15 +925,20 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsMana
             schema.map(|schema| serde_json::Value::Object(schema.clone()))
         }
 
+        let ext_data_service_type = if DtcReadInformationFunction::UserMemoryDtcByStatusMask
+            .default_scope()
+            .eq_ignore_ascii_case(scope)
+        {
+            DtcReadInformationFunction::UserMemoryDtcExtDataRecordByDtcNumber
+        } else {
+            DtcReadInformationFunction::FaultMemoryExtDataRecordByDtcNumber
+        };
         let (extended_data_response, _scope, schema_desc) = self
             .request_extended_data(
                 ecu_name,
                 security_plugin,
                 dtc_code,
-                vec![
-                    DtcReadInformationFunction::FaultMemoryExtDataRecordByDtcNumber,
-                    DtcReadInformationFunction::UserMemoryDtcExtDataRecordByDtcNumber,
-                ],
+                vec![ext_data_service_type],
                 memory_selection,
                 include_schema,
             )
@@ -1000,6 +1006,7 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsMana
         dtc_code: DtcCode,
         include_schema: bool,
         memory_selection: Option<u8>,
+        scope: &str,
     ) -> Result<(Option<ExtendedSnapshots>, Option<serde_json::Value>), DiagServiceError> {
         fn extract_schema_properties(schema_desc: &SchemaDescription) -> Option<serde_json::Value> {
             let param_properties = schema_desc.get_param_properties()?;
@@ -1018,16 +1025,20 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsMana
                 Some(serde_json::Value::Object(schema))
             }
         }
-
+        let snapshot_service_type = if DtcReadInformationFunction::UserMemoryDtcByStatusMask
+            .default_scope()
+            .eq_ignore_ascii_case(scope)
+        {
+            DtcReadInformationFunction::UserMemoryDtcSnapshotRecordByDtcNumber
+        } else {
+            DtcReadInformationFunction::FaultMemorySnapshotRecordByDtcNumber
+        };
         let (snapshot_data_response, _scope, schema_desc) = self
             .request_extended_data(
                 ecu_name,
                 security_plugin,
                 dtc_code,
-                vec![
-                    DtcReadInformationFunction::FaultMemorySnapshotRecordByDtcNumber,
-                    DtcReadInformationFunction::UserMemoryDtcSnapshotRecordByDtcNumber,
-                ],
+                vec![snapshot_service_type],
                 memory_selection,
                 include_schema,
             )
@@ -2278,33 +2289,7 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsEcu
     ) -> Result<DtcExtendedInfo, DiagServiceError> {
         let dtc_code = decode_dtc_from_str(sae_dtc)?;
 
-        let (snapshots, snapshot_schema) = if include_snapshot {
-            self.map_snapshots(
-                ecu_name,
-                security_plugin,
-                dtc_code,
-                include_schema,
-                memory_selection,
-            )
-            .await?
-        } else {
-            (None, None)
-        };
-
-        let (extended_records, extended_schema) = if include_extended_data {
-            self.map_extended_data(
-                ecu_name,
-                security_plugin,
-                dtc_code,
-                include_schema,
-                memory_selection,
-            )
-            .await?
-        } else {
-            (None, None)
-        };
-
-        let mut dtc_by_mask = self
+        let mut dtc_by_mask: HashMap<DtcCode, DtcRecordAndStatus> = self
             .ecu_dtc_by_mask(
                 ecu_name,
                 security_plugin,
@@ -2321,6 +2306,35 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsEcu
                 .ok_or(DiagServiceError::InvalidRequest(format!(
                     "DTC {sae_dtc} not found in ECU {ecu_name}"
                 )))?;
+
+        let scope = &record_and_status.scope.clone();
+        let (snapshots, snapshot_schema) = if include_snapshot {
+            self.map_snapshots(
+                ecu_name,
+                security_plugin,
+                dtc_code,
+                include_schema,
+                memory_selection,
+                scope,
+            )
+            .await?
+        } else {
+            (None, None)
+        };
+
+        let (extended_records, extended_schema) = if include_extended_data {
+            self.map_extended_data(
+                ecu_name,
+                security_plugin,
+                dtc_code,
+                include_schema,
+                memory_selection,
+                scope,
+            )
+            .await?
+        } else {
+            (None, None)
+        };
 
         Ok(DtcExtendedInfo {
             record_and_status,
--- a/cda-comm-uds/src/lib.rs
+++ b/cda-comm-uds/src/lib.rs
@@ -858,7 +858,7 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsMana
         service_types: Vec<DtcReadInformationFunction>,
         memory_selection: Option<u8>,
         include_schema: bool,
-    ) -> Result<(R, String, Option<SchemaDescription>), DiagServiceError> {
+    ) -> Result<(R, DtcReadInformationFunction, Option<SchemaDescription>), DiagServiceError> {
         let ecu = self.ecu_manager(ecu_name)?;
         let (read_func, extended_data_lookup) = ecu
             .read()
@@ -912,7 +912,7 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsMana
         dtc_code: DtcCode,
         include_schema: bool,
         memory_selection: Option<u8>,
-        scope: &str,
+        scope: DtcReadInformationFunction,
     ) -> Result<(Option<ExtendedDataRecords>, Option<serde_json::Value>), DiagServiceError> {
         fn extract_schema_properties(schema_desc: &SchemaDescription) -> Option<serde_json::Value> {
             // todo after solving #54: we are missing the 'Selector' and the case name here
@@ -925,10 +925,7 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsMana
             schema.map(|schema| serde_json::Value::Object(schema.clone()))
         }
 
-        let ext_data_service_type = if DtcReadInformationFunction::UserMemoryDtcByStatusMask
-            .default_scope()
-            .eq_ignore_ascii_case(scope)
-        {
+        let ext_data_service_type = if scope.is_user_scope() {
             DtcReadInformationFunction::UserMemoryDtcExtDataRecordByDtcNumber
         } else {
             DtcReadInformationFunction::FaultMemoryExtDataRecordByDtcNumber
@@ -1006,7 +1003,7 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsMana
         dtc_code: DtcCode,
         include_schema: bool,
         memory_selection: Option<u8>,
-        scope: &str,
+        scope: DtcReadInformationFunction,
     ) -> Result<(Option<ExtendedSnapshots>, Option<serde_json::Value>), DiagServiceError> {
         fn extract_schema_properties(schema_desc: &SchemaDescription) -> Option<serde_json::Value> {
             let param_properties = schema_desc.get_param_properties()?;
@@ -1025,10 +1022,7 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsMana
                 Some(serde_json::Value::Object(schema))
             }
         }
-        let snapshot_service_type = if DtcReadInformationFunction::UserMemoryDtcByStatusMask
-            .default_scope()
-            .eq_ignore_ascii_case(scope)
-        {
+        let snapshot_service_type = if scope.is_user_scope() {
             DtcReadInformationFunction::UserMemoryDtcSnapshotRecordByDtcNumber
         } else {
             DtcReadInformationFunction::FaultMemorySnapshotRecordByDtcNumber
@@ -2183,7 +2177,7 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsEcu
             .filter(|(_, lookup)| {
                 scope
                     .as_ref()
-                    .is_none_or(|scope| scope.to_lowercase() == lookup.scope.to_lowercase())
+                    .is_none_or(|scope| scope.eq_ignore_ascii_case(lookup.scope.default_scope()))
             })
             .collect();
         if scoped_services.is_empty() {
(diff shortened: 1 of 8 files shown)


lorinciubeica10:
Cargo.lock needed because of build_and_test failure: 
image

theswiftfox:
> Cargo.lock needed because of build_and_test failure: image

Thanks! :)
I created #284 a minute ago, as I stumbled across the same issue in my PR ^^
Feel free to 👍 or 👎 on the issue if you have any thoughts about it :)

andreirusu-aumovio:
Thanks @lorinciubeica10 ! I can confirm that this PR solves one of the issues I encountered while extending ODX for DTC Snapshot/Extended data where the response would be wrongly interpreted: 
59 04 01 is actually the Positive Service response + Subfuction + First Byte of DTC instead of the actual DTC of 01 E2 40
image

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -3891,6 +3939,7 @@ impl<S: SecurityPlugin> EcuManager<S> {
         Ok(())
     }
 
+    #[allow(clippy::too_many_arguments)]
```
theswiftfox: question: Do you think it would make sense to introduce a struct that wraps the `&datatypes::Parameter` together with the `base_offset`?
Something like this:
```rust
struct ParamContext<'a> {
    parameter: &'a datatypes::Parameter,
    base_offset: usize,
}
```
My train of thought being that these two are related to each other, and combining them into a single parameter would avoid the clippy warning at the same time and maybe we have additional context we want to pass with the parameter in the future.

This is more of a discussion point, where I'd like your opinion on it as well. If you prefer it as is that is perfectly fine to me as well.

lorinciubeica10: I think it s a good idea since base_offset is always passed alongside param. I will fix it

lorinciubeica10:
Added several fixes for bugs caused by incorrect or missing data when decoding UDS responses:
- Parameters inside Structure DOPs were decoded using relative byte_position
  instead of absolute position in the full response payload.
- map_dynamic_length_field_from_uds computed the wrong item count when a param
  had no explicit byte_position.
- EnvDataDesc decoding used an incorrect base_offset, causing the discriminator
  lookup to fail and snapshot data to be silently dropped.
- Extended data and snapshot requests always tried both FaultMem and UserMem
  service types; they now select the correct one based on the scope parameter.

cda-core/src/diag_kernel/schema.rs
```diff
@@ -255,6 +255,7 @@ fn params_to_schema(
     })
 }
 
+#[allow(clippy::too_many_lines)]
```
theswiftfox: comment: as per codestyle please add a short comment explaining why we have to disable the lint and cannot reduce the loc of the function by extracting things :)

lorinciubeica10: fixed it

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -4159,43 +4228,92 @@ impl<S: SecurityPlugin> EcuManager<S> {
...
+                    },
+                    |s| s as usize,
+                );
                 uds_payload.pop_slice()?;
-                continue;
+                start = start.saturating_add(item_size);
+            } else if let Some(env_data_desc_dop) = repeated_dop.env_data_desc() {
+                if let Some(env_data_desc) = env_data_desc_dop.specific_data_as_env_data_desc() {
+                    let env_data_desc = datatypes::EnvDataDescDop(env_data_desc);
+                    let outer_data = data.clone();
```
alexmohr: todo: Data does not have to be cloned, `outer_data` is passed as reference into `map_env_data_desc_item_from_uds` anyway, so you can pass `data` directly and don't need the clone.

cda-comm-uds/src/lib.rs
```diff
@@ -2278,13 +2289,33 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsEcu
...
+
+        let scope = &record_and_status.scope.clone();
         let (snapshots, snapshot_schema) = if include_snapshot {
             self.map_snapshots(
                 ecu_name,
                 security_plugin,
                 dtc_code,
                 include_schema,
                 memory_selection,
+                scope,
```
alexmohr, line 2312: todo: A better API design of this function is to expect the enum and leave getting that enum to the caller. Having a string as param allows a user to pass gargabe and still get a result back shomehow due to the else branch in the map_snapshots function.

To do this, update in faults.rs
```diff
diff --git a/cda-interfaces/src/datatypes/faults.rs b/cda-interfaces/src/datatypes/faults.rs
index c972886..d4211fd 100644
--- a/cda-interfaces/src/datatypes/faults.rs
+++ b/cda-interfaces/src/datatypes/faults.rs
@@ -25,7 +25,8 @@ pub const CLEAR_FAULT_MEM_POS_RESPONSE_SID: u8 = 0x54;
 /// Essentially the byte values
 /// are sub functions for service 0x19 (Read DTC information)
 #[repr(u8)]
-#[derive(Clone, Copy, PartialEq, Eq, Hash, strum_macros::EnumIter)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, strum_macros::EnumIter, strum_macros::EnumString)]
+#[strum(ascii_case_insensitive)]
 pub enum DtcReadInformationFunction {
     FaultMemoryByStatusMask = 0x02,
     FaultMemorySnapshotRecordByDtcNumber = 0x04,
@@ -56,6 +57,12 @@ impl DtcReadInformationFunction {
         Self::iter().collect()
     }
 
+    pub fn try_from_str(s: &str) -> Result<Self, crate::DiagServiceError> {
+        std::str::FromStr::from_str(s).map_err(|_| crate::DiagServiceError::InvalidParameter {
+            possible_values: Self::iter().map(|v| format!("{v:?}")).collect(),
+        })
+    }
+
     #[must_use]
     pub fn is_user_scope(&self) -> bool {
         matches!(
```
Then parse the scope here via 

```
let scope = DtcReadInformationFunction::try_from_str(scope)?;
```

This way the caller of this function gets a proper error message with the possible values they can call, the api of map_snapshots is cleaner and there is no silent fallback

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -4301,48 +4419,162 @@ impl<S: SecurityPlugin> EcuManager<S> {
...
+                    }),
+                    _ => Err(DiagServiceError::ParameterConversionError(format!(
+                        "EnvDataDesc discriminator '{param_short_name}' resolved to unexpected \
+                         type: {val:?}"
+                    ))),
+                };
+            }
+            _ => {}
+        }
+        for container in outer_data.values() {
```
alexmohr: question: Is it a good idea to have this fallback instead of returning an error? I would assume if the match block above didn't find a value, this should be an error? Please add a small comment on why this is necessary if you don't agree or I am missing something. Also `.values` of a hashmap gives you ` all values in arbitrary order.` I'm concerned this will lookup the wrong element by chance.

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -2435,7 +2471,7 @@ impl<S: SecurityPlugin> EcuManager<S> {
         let predicate = |service: &datatypes::DiagService<'_>| {
             service.diag_comm().is_some_and(|dc| {
                 dc.short_name()
-                    .is_some_and(|name| starts_with_ignore_ascii_case(name, &lookup_name))
+                    .is_some_and(|name| name.eq_ignore_ascii_case(&lookup_name))
```
alexmohr: suggestion: If possible please change this back to `starts_with_ignore_ascii_case`, iirc this would break lookups in some cases if you use `eq` here.

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -4159,43 +4228,92 @@ impl<S: SecurityPlugin> EcuManager<S> {
...
+            payload_len = uds_payload.len(),
+            count_bytes = ?num_items_data,
+            "DynamicLengthField items loop start"
+        );
+
+        uds_payload.push_slice(items_abs_start, uds_payload.len())?;
+
+        let mut start = 0usize;
 
         for _ in 0..num_items {
```
alexmohr, line 4233: question: I would argue for extracting this into a new function, I'm not sure if I agree with disabling the linter warning about the function beeing to long.

alexmohr: Suggestion:
```diff 
diff --git a/cda-core/src/diag_kernel/ecumanager.rs b/cda-core/src/diag_kernel/ecumanager.rs
index c5cf4d4..4b910d9 100644
--- a/cda-core/src/diag_kernel/ecumanager.rs
+++ b/cda-core/src/diag_kernel/ecumanager.rs
@@ -4159,8 +4159,6 @@ impl<S: SecurityPlugin> EcuManager<S> {
             dlt_context = dlt_ctx!("CORE"),
         )
     )]
-    // allow keeping the function together as it makes sense structurally
-    #[allow(clippy::too_many_lines)]
     fn map_dynamic_length_field_from_uds(
         &self,
         mapped_service: &datatypes::DiagService,
@@ -4215,12 +4213,13 @@ impl<S: SecurityPlugin> EcuManager<S> {
             &num_items_data,
         )?;
 
-        let repeated_dop =
+        let repeated_dop = datatypes::DopField(
             dynamic_length_field_dop
                 .field()
                 .ok_or(DiagServiceError::InvalidDatabase(
                     "DynamicLengthField field is None".to_owned(),
-                ))?;
+                ))?,
+        );
 
         let num_items: u32 = num_items_diag_val.try_into()?;
         let num_items_byte_pos = determine_num_items.byte_position() as usize;
@@ -4244,73 +4243,15 @@ impl<S: SecurityPlugin> EcuManager<S> {
         let mut start = 0usize;
 
         for _ in 0..num_items {
-            let bytes_to_skip_before = uds_payload.bytes_to_skip();
-            uds_payload.push_slice(start, uds_payload.len())?;
-            // Inner item params may omit BYTE-POSITION, falling back to last_read_byte_pos.
-            // Reset it to 0 so stale state from count decode doesn't corrupt item decoding.
-            uds_payload.set_last_read_byte_pos(0);
-            let basic_struct_type = repeated_dop
-                .basic_structure()
-                .map(|d| d.specific_data_type());
-            tracing::debug!(
-                has_basic_structure = repeated_dop.basic_structure().is_some(),
-                has_env_data_desc = repeated_dop.env_data_desc().is_some(),
-                basic_struct_specific_data_type = ?basic_struct_type,
-                "DynamicLengthField item decode metadata"
-            );
-            if let Some(s) = repeated_dop
-                .basic_structure()
-                .and_then(|d| d.specific_data_as_structure().map(datatypes::StructureDop))
-            {
-                let struct_data =
-                    self.map_struct_from_uds(&s, mapped_service, uds_payload, data)?;
-                repeated_data.push(struct_data);
-                let item_size = s.byte_size().map_or_else(
-                    || {
-                        let delta_bytes_to_skip = uds_payload
-                            .bytes_to_skip()
-                            .saturating_sub(bytes_to_skip_before);
-                        uds_payload
-                            .last_read_byte_pos()
-                            .saturating_add(delta_bytes_to_skip)
-                    },
-                    |s| s as usize,
-                );
-                uds_payload.pop_slice()?;
-                start = start.saturating_add(item_size);
-            } else if let Some(env_data_desc_dop) = repeated_dop.env_data_desc() {
-                if let Some(env_data_desc) = env_data_desc_dop.specific_data_as_env_data_desc() {
-                    let env_data_desc = datatypes::EnvDataDescDop(env_data_desc);
-                    let outer_data = data.clone();
-                    let item_data = self.map_env_data_desc_item_from_uds(
-                        mapped_service,
-                        uds_payload,
-                        &outer_data,
-                        &env_data_desc,
-                        0,
-                    )?;
-                    let delta_bytes_to_skip = uds_payload
-                        .bytes_to_skip()
-                        .saturating_sub(bytes_to_skip_before);
-                    let item_size = uds_payload
-                        .last_read_byte_pos()
-                        .saturating_add(delta_bytes_to_skip);
-                    uds_payload.pop_slice()?;
-                    repeated_data.push(item_data);
-                    start = start.saturating_add(item_size);
-                } else {
-                    uds_payload.pop_slice()?;
-                    return Err(DiagServiceError::InvalidDatabase(
-                        "DynamicLengthField env_data_desc DOP is not EnvDataDesc type".to_owned(),
-                    ));
-                }
-            } else {
-                uds_payload.pop_slice()?;
-                return Err(DiagServiceError::InvalidDatabase(
-                    "DynamicLengthField repeated_dop is neither Structure nor EnvDataDesc"
-                        .to_owned(),
-                ));
-            }
+            let (item_data, item_size) = self.decode_dynamic_length_field_item(
+                mapped_service,
+                uds_payload,
+                data,
+                &repeated_dop,
+                start,
+            )?;
+            repeated_data.push(item_data);
+            start = start.saturating_add(item_size);
         }
         uds_payload.pop_slice()?;
         uds_payload.set_last_read_byte_pos(items_abs_start.saturating_add(start));
@@ -4321,6 +4262,80 @@ impl<S: SecurityPlugin> EcuManager<S> {
         Ok(())
     }
 
+    fn decode_dynamic_length_field_item(
+        &self,
+        mapped_service: &datatypes::DiagService,
+        uds_payload: &mut Payload,
+        data: &mut MappedDiagServiceResponsePayload,
+        repeated_dop: &datatypes::DopField,
+        start: usize,
+    ) -> Result<(HashMap<String, DiagDataTypeContainer>, usize), DiagServiceError> {
+        let bytes_to_skip_before = uds_payload.bytes_to_skip();
+        uds_payload.push_slice(start, uds_payload.len())?;
+        // Inner item params may omit BYTE-POSITION, falling back to last_read_byte_pos.
+        // Reset it to 0 so stale state from count decode doesn't corrupt item decoding.
+        uds_payload.set_last_read_byte_pos(0);
+
+        tracing::debug!(
+            has_basic_structure = repeated_dop.basic_structure().is_some(),
+            has_env_data_desc = repeated_dop.env_data_desc().is_some(),
+            basic_struct_specific_data_type = ?repeated_dop
+                .basic_structure()
+                .map(|d| d.specific_data_type()),
+            "DynamicLengthField item decode metadata"
+        );
+
+        if let Some(s) = repeated_dop
+            .basic_structure()
+            .and_then(|d| d.specific_data_as_structure().map(datatypes::StructureDop))
+        {
+            let struct_data =
+                self.map_struct_from_uds(&s, mapped_service, uds_payload, data)?;
+            let item_size = s.byte_size().map_or_else(
+                || {
+                    let delta_bytes_to_skip = uds_payload
+                        .bytes_to_skip()
+                        .saturating_sub(bytes_to_skip_before);
+                    uds_payload
+                        .last_read_byte_pos()
+                        .saturating_add(delta_bytes_to_skip)
+                },
+                |s| s as usize,
+            );
+            uds_payload.pop_slice()?;
+            Ok((struct_data, item_size))
+        } else if let Some(env_data_desc_dop) = repeated_dop.env_data_desc() {
+            let env_data_desc = env_data_desc_dop
+                .specific_data_as_env_data_desc()
+                .ok_or(DiagServiceError::InvalidDatabase(
+                    "DynamicLengthField env_data_desc DOP is not EnvDataDesc type".to_owned(),
+                ))?;
+            let env_data_desc = datatypes::EnvDataDescDop(env_data_desc);
+            let outer_data = data.clone();
+            let item_data = self.map_env_data_desc_item_from_uds(
+                mapped_service,
+                uds_payload,
+                &outer_data,
+                &env_data_desc,
+                0,
+            )?;
+            let delta_bytes_to_skip = uds_payload
+                .bytes_to_skip()
+                .saturating_sub(bytes_to_skip_before);
+            let item_size = uds_payload
+                .last_read_byte_pos()
+                .saturating_add(delta_bytes_to_skip);
+            uds_payload.pop_slice()?;
+            Ok((item_data, item_size))
+        } else {
+            uds_payload.pop_slice()?;
+            Err(DiagServiceError::InvalidDatabase(
+                "DynamicLengthField repeated_dop is neither Structure nor EnvDataDesc"
+                    .to_owned(),
+            ))
+        }
+    }
+
     fn map_mux_dop_from_uds(
         &self,
         mapped_service: &datatypes::DiagService,

```

alexmohr (changes requested):

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -5154,6 +5391,22 @@ fn extract_struct_dop_from_field(
...
 }
 
+fn param_position_order(a: &datatypes::Parameter, b: &datatypes::Parameter) -> std::cmp::Ordering {
+    match (a.has_byte_position(), b.has_byte_position()) {
+        // Both have a position → normal comparison
+        (true, true) => a
+            .byte_position()
+            .cmp(&b.byte_position())
+            .then(a.bit_position().cmp(&b.bit_position())),
+        // Only a has no position → a goes after b
```
alexmohr: suggestion: If possible don't use unicode chars, we will forbid this in #275

cda-core/src/diag_kernel/ecumanager.rs
lorinciubeica10: fixed

cda-comm-uds/src/lib.rs
lorinciubeica10: done

cda-core/src/diag_kernel/ecumanager.rs
lorinciubeica10: this one is tricky. I made this change because of the same reason: if you have 2 names that have the same starting string, it will choose the first one

cda-core/src/diag_kernel/ecumanager.rs
lorinciubeica10: done

cda-core/src/diag_kernel/ecumanager.rs
lorinciubeica10: done

alexmohr: Please don't resolve my threads. I will resolve them once I deem them to be solved. It helps me to keep track over the changes, so I can review deltas better. I unresolved them all for now, doesn't mean I think it still needs work, just helps me to keep an overview. Thanks for understanding.

alexmohr (approved):
lgtm, sorry to keep you waiting for so long.
You might want to squash the clippy commit :)

lorinciubeica10:
> lgtm, sorry to keep you waiting for so long. You might want to squash the clippy commit :)

squash done

PR #297, opened 2026-04-27
fix: correct leading length extraction slice bounds when byte_pos > 0

alexmohr:
## Summary

The leading length extraction was using an incorrect slice range that ignored byte_pos when calculating the end position. This caused the slice to start at byte_pos but end at an absolute position, leading to incorrect behavior when byte_pos > 0.

If byte_pos = 5 and bits = 16, it would try uds_payload.get(5..2)

Added regression tests to verify leading length decoding works correctly with non-zero byte positions for both 8-bit and 16-bit leading lengths.

## Checklist

## Notes for Reviewers


theswiftfox (approved):
lgtm :)

Issue #122, opened 2025-12-03
[Bug] Hard coded protocoll name does not work for all ecus.


alexmohr:
## Description

Some ECUs may not use the hard-coded protocol names `UDS_Ethernet_DoIP_DOBT` and `UDS_Ethernet_DoIP` for the DOIP protocols.
Instead we could use the `PROT-STACK` short name. 

* `SHORT-NAME	ISO_14229_5_on_ISO_13400_2`

But this name will be the same for both protocols, so we have to find some factor to differentiate both.

## Steps to Reproduce

1. Load database where the protocol short name is not equal to the ones mentioned above. 

## Expected Behavior

CDA should find the correct protocols and continue loading the DB.

## Actual Behavior

CDA will log an error that the database cannot be loaded as no valid protocol can be found. 

## Screenshots or Logs (if applicable)

> [!IMPORTANT]
> Link the pull request that addresses this bug once created.

alexmohr:
Only option to solve this is to make it configurable. It's planned to have a configuration with prioritization but we need to draft a proper concept for this first. It's most likely not something that will be solved this year.

alexmohr:
Protocol names are configurable now, this is solved in #304

PR #306, opened 2026-04-29
feature: make protocol lookup optional and protocol names configurable [hackfest fix]

alexmohr:
## Summary
- Make protocol lookup optional for MDD files that do not link ComParams back to a protocol. When only one protocol exists in the database, CDA assumes it is the target protocol without requiring explicit ComParam linkage.
- Make protocol names configurable via `doip_protocol_name` and `doip_dobt_protocol_name` in the DoIP config section, replacing the previously hardcoded `Protocol` enum values.
- Remove `protocol_case_sensitive` option in favor of always matching case-insensitively.
- Move `DatabaseConfig` to `cda-database` crate and add `ignore_protocol` option.

## Checklist

## Related

#122

## Notes for Reviewers

This issue was found during SDV Hackfest 2026.

The `Protocol` enum now carries the protocol name as a `String` instead of being a unit variant, making it configurable per deployment without code changes.

### DIFF
--- a/cda-core/src/diag_kernel/ecumanager.rs
+++ b/cda-core/src/diag_kernel/ecumanager.rs
@@ -134,8 +134,6 @@ pub struct EcuManager<S: SecurityPlugin> {
     protocol: Protocol,
     // functional group: protocol prefixed or postfixed
     fg_protocol_position: DiagnosticServiceAffixPosition,
-    // functional group: is protocol case sensitive
-    fg_protocol_case_sensitive: bool,
     ecu_service_states: Arc<RwLock<HashMap<u8, String>>>,
 
     tester_present_retry_policy: bool,
@@ -219,8 +217,8 @@ impl<S: SecurityPlugin> cda_interfaces::EcuManager for EcuManager<S> {
         self.variant.state
     }
 
-    fn protocol(&self) -> Protocol {
-        self.protocol
+    fn protocol(&self) -> &Protocol {
+        &self.protocol
     }
 
     fn is_loaded(&self) -> bool {
@@ -342,7 +340,7 @@ impl<S: SecurityPlugin> cda_interfaces::EcuManager for EcuManager<S> {
                 cp_ref.protocol().is_some_and(|p| {
                     p.diag_layer().is_some_and(|dl| {
                         dl.short_name()
-                            .is_some_and(|name| name == self.protocol.value())
+                            .is_some_and(|name| name.eq_ignore_ascii_case(self.protocol.str()))
                     })
                 })
             })
@@ -1522,21 +1520,13 @@ impl<S: SecurityPlugin> cda_interfaces::EcuManager for EcuManager<S> {
                     .diag_layer()
                     .and_then(|dl| dl.short_name())
                     .and_then(|name| {
-                        let protocol_value = self.protocol.value();
+                        let protocol_value = self.protocol.str();
                         let matches = match self.fg_protocol_position {
                             DiagnosticServiceAffixPosition::Prefix => {
-                                if self.fg_protocol_case_sensitive {
-                                    name.starts_with(protocol_value)
-                                } else {
-                                    util::starts_with_ignore_ascii_case(name, protocol_value)
-                                }
+                                util::starts_with_ignore_ascii_case(name, protocol_value)
                             }
                             DiagnosticServiceAffixPosition::Suffix => {
-                                if self.fg_protocol_case_sensitive {
-                                    name.ends_with(protocol_value)
-                                } else {
-                                    util::ends_with_ignore_ascii_case(name, protocol_value)
-                                }
+                                util::ends_with_ignore_ascii_case(name, protocol_value)
                             }
                         };
                         if matches {
@@ -2066,7 +2056,7 @@ impl<S: SecurityPlugin> EcuManager<S> {
         let variant_detection =
             variant_detection::prepare_variant_detection(&database, &database_naming_convention)?;
 
-        let data_protocol = into_db_protocol(&database, protocol)?;
+        let data_protocol = into_db_protocol(&database, &protocol)?;
 
         let logical_gateway_address = match database.find_logical_address(
             datatypes::LogicalAddressType::Gateway(
@@ -2162,7 +2152,6 @@ impl<S: SecurityPlugin> EcuManager<S> {
             duplicating_ecu_names: None,
             protocol,
             fg_protocol_position: func_description_config.protocol_position.clone(),
-            fg_protocol_case_sensitive: func_description_config.protocol_case_sensitive,
             ecu_service_states: Arc::new(RwLock::default()),
             tester_present_retry_policy: database
                 .find_com_param(&data_protocol, &com_params.uds.tester_present_retry_policy)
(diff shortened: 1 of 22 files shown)


alexmohr:
broken taplo seems unrelated.... need investigation 👀

alexmohr:
Solves #122

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -5436,7 +5436,7 @@ mod tests {
     ) -> super::EcuManager<DefaultSecurityPluginData> {
         let mut manager = super::EcuManager::new(
             db,
-            Protocol::default_doip(),
+            Protocol::new("UDS_Ethernet_DoIP"),
```
alexmohr: Todo: use a const, so it can be default constructed withoit setting a protocol name every time  or even better use the existing dobt default

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -5892,7 +5883,8 @@ mod tests {
             db_builder.create_diag_coded_type_standard_length(32, DataType::Float32);
         let ascii_diag_type =
             db_builder.create_diag_coded_type_standard_length(32, DataType::AsciiString);
-        let protocol = db_builder.create_protocol(Protocol::DoIp.value(), None, None, None);
+        let protocol_name = Protocol::new("UDS_Ethernet_DoIP").to_string();
```
alexmohr: Todo: use the default instead

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -309,8 +307,9 @@ impl<S: SecurityPlugin> cda_interfaces::EcuManager for EcuManager<S> {
             .filter(|cp_ref| {
                 cp_ref.protocol().is_some_and(|p| {
                     p.diag_layer().is_some_and(|dl| {
-                        dl.short_name()
-                            .is_some_and(|name| name == self.protocol.value())
+                        dl.short_name().is_some_and(|name| {
+                            name.eq_ignore_ascii_case(&self.protocol.to_string())
```
theswiftfox: suggestion: add a str() method to Protocol and use that here to avoid allocating a new string.
```rust
impl Protocol {
...
    #[must_use]
    pub fn str(&self) -> &str {
        &self.0
    }
}
```
and then call as `.is_some_and(|name: &str| name.eq_ignore_ascii_case(&self.protocol.str()))`

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -188,7 +186,7 @@ impl<S: SecurityPlugin> cda_interfaces::EcuManager for EcuManager<S> {
     }
 
     fn protocol(&self) -> Protocol {
-        self.protocol
+        self.protocol.clone()
```
theswiftfox: suggestion: change the trait to return &Protocol and return reference here as the Protocol is not `Copy` anymore.  
I think currently we only use it in one place to then turn the protocol into a string, and we could spare the copy then as the to_string() could also be called on the reference?

cda-core/src/diag_kernel/mod.rs
```diff
@@ -230,29 +230,38 @@ impl TryInto<u32> for DiagDataValue {
...
-            ))
-        })?;
+pub fn into_db_protocol<'db>(
+    database: &'db datatypes::DiagnosticDatabase,
+    protocol: &cda_interfaces::Protocol,
+) -> Result<datatypes::Protocol<'db>, DiagServiceError> {
+    let protocol_name = protocol.to_string();
+    let all_protocols = database.protocols()?;
+
+    // Try exact match first.
```
theswiftfox, line 240: nitpick: technically not an 'exact' match if we match on ignore case ;)

cda-database/src/lib.rs
```diff
@@ -15,7 +15,32 @@ pub(crate) mod flatbuf;
...
+    ///
+    /// This is useful for databases that contain only a single protocol but
+    /// where the protocol short-name recorded in the database does not match
+    /// the one used at runtime. Enabling this flag allows the lookup to succeed
+    /// by falling back to a protocol-agnostic match.
+    ///
+    /// Only valid when the database contains exactly one distinct protocol;
+    /// attempting a lookup with this flag set against a multi-protocol database
+    /// returns `DiagServiceError::InvalidDatabase`.
+    pub ignore_protocol: bool,
```
theswiftfox, line 45: question: (this is a very general question, but felt most fitting to put it here)
It seems like this is almost a different 'mode' to run the CDA in. Something like a `single-protocol-mode` and i'm wondering if it makes sense to wrap these config adaptions into a mode switch for easier use? And allow for example on startup to check if the cda is started in single-protocol-mode but the database contains multiple protocols and in that case return an error?

This is mostly food for thought and not a critique of the current implementation ^^

cda-main/src/main.rs
```diff
@@ -41,10 +41,10 @@ struct AppArgs {
...
     gateway_port: Option<u16>,
 
-    // cannot use Action::SetTrue as it will treat
-    // absent arg same as `= false`
-    #[arg(short, long)]
-    onboard_tester: Option<bool>,
+    /// Protocol name used for com-param lookups in the diagnostic database (matched case-insensitively).
+    /// Examples: `UDS_Ethernet_DoIP`, `UDS_Ethernet_DoIP_DOBT`
+    #[arg(long)]
+    protocol_name: Option<String>,
```
theswiftfox, line 48: maybe the other comment would have been more fitting here I just realized, as the change of this flag is what got me thinking in that direction. I'll just dump some more thoughts I had here ^^
Imo, it would be nice to still be able to use this flag in cases where two protocols for doip are defined, but have the other 'operating' mode sorts of parallel.

theswiftfox (commented):
some minor findings, and a question I should have already asked at heckfest 🙈

cda-main/src/main.rs
alexmohr: Initially I left the `onboard-tester` logic in place and just made the protocol names configurable. on #315 @floroks commented https://github.com/eclipse-opensovd/classic-diagnostic-adapter/pull/315#discussion_r3173173162 

> we should probably remove the onboard vs. offboard protocol names and just use a single protocol name configuration via config file (or parameter), with the dobt one being the default

But after @MarkSchmitt opened #323 we might want to have a way to set this globally like here, but overwrite it per ECU. #315 still assumes the protocol is the same for all ECUs

cda-database/src/lib.rs
alexmohr: Right now we support a single protocol anyway, as we cannot set or overwrite the protocol per ECU, but the protocol is passed during startup?

theswiftfox: Ah yes. that was a misleading qustion - I should have read the comment then I would have noticed that it states exactly what i was asking 🤦‍♀️😅

cda-main/src/main.rs
theswiftfox: hm, so in the form of a `base_protocol` and then `ECU.protocol` config? I'd guess it would make sense to do that as a follow up to this PR then to introduce the per ECU config in some form?

cda-core/src/diag_kernel/ecumanager.rs
alexmohr: done

cda-core/src/diag_kernel/ecumanager.rs
alexmohr: done

cda-core/src/diag_kernel/mod.rs
alexmohr: done 😆

alexmohr:
@theswiftfox pipeline also fixed, sorry for the hickups :)

theswiftfox (approved):

PR #309, opened 2026-04-30
feat(config): add --config CLI flag for config file path [HACKFEST FIX]

alexmohr:
## Summary

Add a `--config` / `-c` CLI flag to specify the configuration file path at runtime, giving operators a direct way to point CDA at a specific config without relying solely on environment variables.

Config file resolution priority: CLI `--config` arg → `CDA_CONFIG_FILE` env var → `<CDA_NAME>.toml` default.

## Checklist

## Notes for Reviewers

- The `clap` dependency gains the `env` feature to support `env = "CDA_CONFIG_FILE"` on the arg definition, keeping env-var fallback behavior co-located with the CLI definition.
- `load_config` now accepts `Option<&str>` — fully backwards-compatible since `None` preserves the previous behavior.

---

### DIFF
--- a/cda-main/src/config/mod.rs
+++ b/cda-main/src/config/mod.rs
@@ -17,21 +17,21 @@ use figment::{
 
 pub mod configfile;
 
-/// Loads the configuration from a file specified by the `CDA_CONFIG_FILE` environment variable.
-/// If the variable is not set, it defaults to `opensovd-cda.toml`.
-/// The configuration is merged with default values and environment variables prefixed with `CDA`.
-/// # Returns
-/// A `Result` containing the loaded configuration or an error message if the loading fails
+/// Loads the configuration, merged with defaults and `CDA`-prefixed env vars.
+///
+/// Config file resolved in priority order:
+/// * `config_file` arg (includes `CDA_CONFIG_FILE` env via clap)
+/// * `<CDA_NAME>.toml`
 /// # Errors
 /// Returns an error message if the configuration file cannot be read or parsed.
-pub fn load_config() -> Result<configfile::Configuration, String> {
+pub fn load_config(config_file_path: Option<&str>) -> Result<configfile::Configuration, String> {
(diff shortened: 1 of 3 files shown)


cda-main/src/config/mod.rs
```diff
@@ -17,17 +17,20 @@ use figment::{
...
+/// Loads the configuration, merged with defaults and `CDA`-prefixed env vars.
+///
+/// Config file resolved in priority order:
+/// * `config_file` arg
+/// * `CDA_CONFIG_FILE` env
+/// * `<CDA_NAME>.toml`
 /// # Errors
 /// Returns an error message if the configuration file cannot be read or parsed.
-pub fn load_config() -> Result<configfile::Configuration, String> {
+pub fn load_config(config_file: Option<&str>) -> Result<configfile::Configuration, String> {
```
theswiftfox: nitpick: maybe `config_file_path`? I know the name in the function was already `config_file` but when reading I noticed that this could be confused as being the config file content being passed in ^^

cda-main/src/main.rs
```diff
@@ -29,6 +29,9 @@ const MAIN_HEALTH_COMPONENT_KEY: &str = "main";
 #[derive(Parser, Debug)]
 #[command(version, about, long_about = None)]
 struct AppArgs {
+    #[arg(short, long, env = "CDA_CONFIG_FILE")]
```
theswiftfox, line 32: question: I did not look this up, is the env here only descriptive, or does this also try to parse it from env as part of Clap? If the second is the case, we should be able to get rid of the manual env parsing ourselves 🤔

alexmohr: It also parses the env variable as alternative to the command line arguments, so we can remove the manual parsing. Good catch. Thank you.

alexmohr: done

cda-main/src/config/mod.rs
alexmohr: done

theswiftfox (approved):

PR #312, opened 2026-04-30
operations: add /operations/{operation}/docs

theswiftfox:
## Summary

This PR adds the basic helpers and structures to provide an online capability description (ISO 17978-3 Section 7.5). As a first step the /docs endpoint is added for ../operations/{operation}/docs in both component and functional groups. 

## Checklist

## Notes for Reviewers

---

### DIFF
--- a/cda-core/src/diag_kernel/schema.rs
+++ b/cda-core/src/diag_kernel/schema.rs
@@ -80,6 +80,27 @@ impl<S: SecurityPlugin> EcuSchemaProvider for EcuManager<S> {
             main_schema,
         ))
     }
+
+    async fn schema_for_fg_request(
+        &self,
+        service: &cda_interfaces::DiagComm,
+        functional_group_name: &str,
+    ) -> Result<SchemaDescription, DiagServiceError> {
+        let mapped_service = self
+            .lookup_diag_service(service, Some(functional_group_name), None)
+            .await?;
+        let ctx = service_context(service, &mapped_service);
+
+        let request = mapped_service.request().map(datatypes::Request).ok_or(
+            DiagServiceError::InvalidDatabase(format!(
+                "Missing request for service {} in functional group {}.",
+                service.name, functional_group_name
+            )),
+        )?;
+        let schema = request.json_schema(&ctx, &self.diag_database);
+
+        Ok(schema)
+    }
 }
 
 fn service_context(service: &cda_interfaces::DiagComm, mapped_service: &DiagService) -> String {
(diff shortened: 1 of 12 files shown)


cda-core/src/diag_kernel/schema.rs
```diff
@@ -255,6 +276,135 @@ fn params_to_schema(
...
+        }
+    }
+
+    if let Some(upper) = constr.upper_limit() {
+        let interval: datatypes::IntervalType = upper.interval_type().into();
+        if interval != datatypes::IntervalType::Infinite
+            && let Some(value) = upper.value().and_then(|v| v.parse::<f64>().ok())
+        {
+            let key = if interval == datatypes::IntervalType::Open {
+                "exclusiveMaximum"
```
alexmohr, line 332: question: are these values defined in the standard or are these MB special and should be configurable?

cda-sovd/src/sovd/docs/operations.rs
```diff
@@ -0,0 +1,348 @@
...
+fn sovd_extensions(meta: &OperationDocsMeta) -> IndexMap<String, serde_json::Value> {
+    let mut ext = IndexMap::new();
+    if meta.is_async {
+        ext.insert(
+            "x-sovd-asynchronous-execution".to_owned(),
+            serde_json::Value::Bool(true),
+        );
+    }
+    // proximity_proof_required is always false for classic UDS routines.
+    ext.insert(
```
alexmohr: question: This only defines two of the attributes from the table, 
`x-sovd-lock-scopes` and `x-sovd-lock-required` are probably interseting as well. What's the reason for picking these two?
Furthermore, to make this re-usable i.e. in opensovd-core/server, it might make sense to move the 'x-sovd...' into constants in the sovd-interfaces crate?  In general most of this module looks like it would be very similar for the server?

theswiftfox: oh i think i got things mixed up here. this should be in relation to table 169 in the spec, that defines id, proximity_proof_required and asynchronous_execution as mandatory. name and translation_id as optional.

I agree that the x-sovd prefix should be dropped then, as it is not an extension but in the standard. also the method needs to be renamed 😅

i'd prefer leave the extensions out for now until we implement the functionality for them.

theswiftfox: and, i think moving the strings to constants makes sense

cda-core/src/diag_kernel/schema.rs
theswiftfox: i think thats just the mathematical name for them at least from my memory ^^
```
Inclusive (Closed Interval [a, b] ): both 'a' and 'b' are part of the interval.
Exclusive (Open Interval (a, b) ): 'a' and 'b' are not part of the interval.
```
and ofc the combinations - thats what i was trying to model here.

Correct me if i'm wrong here - happy to update it then :D

alexmohr: oh.. ya no sorry I thought for some reason that name must be found in the database, but obviously that is not the case. Totally agree with the naming you chose :)

theswiftfox: still a good hint, that maybe a comment would be beneficial here ;) I'll take that as a ToDo to add basically my explanation now as a small doc comment :D

cda-sovd/src/sovd/docs/operations.rs
alexmohr: @theswiftfox I'm not quite sure if you still want to change these things in this PR, are you still working on this?

theswiftfox: Yes, just picked it up again today :) Working on fixing this as we speak!

theswiftfox: Pushed an update for this.
Now after rechecking the spec, the x-sovd prefix is required as per spec, as stated in table 126: 
> If true, the execution of the operation is asynchronous and the offline capability description includes the attribute `x-sovd-asynchronous-execution` with the value `true` in the `Opration` object

But in the same line, I noticed, that the proximity_proof_required should be in the pathitem for the capability description according to the same table. So I fixed that ;)

And I extracted the keys into the sovd-interfaces crate as constants.

As for the rest of the module - I'd suggest we evaluate when adding more /docs endpoints, what parts make sense to move to the sovd-interfaces for reuse, and what parts make more sense to keep in the 'local' area of the functionality. Does that sound good to you?

alexmohr: Sounds good to me. We have to talk about `sovd-interfaces` anyway again, best together with someone from `core` as they have implemented something similar but not quite the same. We should align this, but that's another story.

alexmohr (approved):
lgtm, thank you

PR #313, opened 2026-05-01
feat(config): add generate-config subcommand for reference TOML output

alexmohr:
## Summary

Add a `generate-config` CLI subcommand to the opensovd-cda binary that produces a fully-commented TOML reference configuration file (`opensovd-cda.reference.toml`) with all fields documented using Rust doc comments extracted via `schemars`.

Adding this because having a configuration file with all possible values makes it way easier to see what is available and how to configure it. 

### What it does

- `opensovd-cda generate-config [--output PATH]` generates a reference TOML file with **all** configuration fields present but commented out
- Fields with `///` doc comments get their description as a `# comment` line above the field
- The output only includes configuration sections that are compiled in (feature-gated sections like `health`, `tokio-tracing`, `dlt-tracing` appear only when their respective features are enabled)
- Default output path: `opensovd-cda.reference.toml` in the current directory

### Changes

- **`cda-main/src/config/generate.rs`** (new): Core generator module — walks JSON Schema from `schemars::schema_for!(Configuration)`, serializes defaults to TOML, injects doc-comment descriptions
- **`cda-main/src/main.rs`**: `GenerateConfig` CLI subcommand with early exit before async runtime
- **Config structs across 6 crates**: Added `#[derive(schemars::JsonSchema)]` to all configuration types
- **Complex type handling**: Manual `JsonSchema` impl for `ComParamBool`, `#[schemars(schema_with)]` for `Duration` fields, `#[schemars(bound)]` for `ComParamConfig<T>`

## Checklist

## Notes for Reviewers

---


cda-main/src/config/generate.rs
```diff
@@ -0,0 +1,280 @@
...
+        config.logging.tokio_tracing.recording_path =
+            Some("/tmp/tokio-recording".to_owned());
+    }
+
+    config
+}
+
+const SPDX_HEADER: &str = "\
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: 2025 The Contributors to Eclipse OpenSOVD (see CONTRIBUTORS)
```
alexmohr: todo: fix year to 2026

.github/workflows/build.yml
```diff
@@ -52,6 +52,19 @@ jobs:
         uses: eclipse-opensovd/dlt-tracing-lib/.github/actions/setup-dlt@main
       - name: Build all features
         run: cargo build --locked --verbose --all-features
+      - name: Verify reference config is up to date
+        run: |
+          cargo run --locked --all-features -- generate-config --output /tmp/opensovd-cda.toml
+          if ! diff -u opensovd-cda.toml /tmp/opensovd-cda.toml; then
+            echo ""
+            echo "::error::opensovd-cda.toml is out of date."
+            echo "Expected contents (freshly generated):"
```
alexmohr: Todo: remove freshly generated text, keep expected content

cda-main/src/config/generate.rs
```diff
@@ -0,0 +1,282 @@
...
+        false
+    }
+}
+
+/// Post-process the raw TOML string: comment out value lines, inject descriptions.
+fn process_toml(raw_toml: &str, desc_map: &BTreeMap<String, String>) -> String {
+    let mut output = String::new();
+    let mut section_stack: Vec<String> = Vec::new();
+
+    for line in raw_toml.lines() {
```
alexmohr: Todo: rework this, the logic is too nested, prefer functional style

cda-main/src/config/generate.rs
```diff
@@ -0,0 +1,282 @@
...
+///
+/// Panics if the configuration schema cannot be serialized to JSON or TOML.
+#[must_use]
+pub fn generate_reference_config() -> String {
+    let schema = schemars::schema_for!(Configuration);
+    let schema_json =
+        serde_json::to_value(&schema).expect("schema serialization to JSON should never fail");
+
+    let toml_value = toml::Value::try_from(reference_config_instance())
+        .expect("reference config instance serialization to TOML Value should never fail");
```
alexmohr: todo: do not panic, return an error instead.

cda-main/src/config/generate.rs
```diff
@@ -0,0 +1,282 @@
...
+
+fn walk_schema(
+    node: &serde_json::Value,
+    defs: &serde_json::Value,
+    prefix: &str,
+    map: &mut BTreeMap<String, String>,
+) {
+    let resolved = resolve_ref(node, defs);
+
+    if let Some(properties) = resolved.get("properties").and_then(|p| p.as_object()) {
```
alexmohr: todo: rewrite this in a functional style

cda-main/src/main.rs
```diff
@@ -86,6 +99,20 @@ struct AppArgs {
...
 async fn main() -> Result<(), AppError> {
     let args = AppArgs::parse();
+
+    if let Some(Command::GenerateConfig { ref output }) = args.command {
+        let output_path = output
+            .clone()
+            .unwrap_or_else(|| PathBuf::from("opensovd-cda.toml"));
+        let content = opensovd_cda_lib::config::generate::generate_reference_config();
+        std::fs::write(&output_path, content).expect("Failed to write reference config");
+        println!(
```
alexmohr: remove this println

.github/workflows/build.yml
```diff
@@ -52,6 +52,16 @@ jobs:
         uses: eclipse-opensovd/dlt-tracing-lib/.github/actions/setup-dlt@main
       - name: Build all features
         run: cargo build --locked --verbose --all-features
+      - name: Verify reference config is up to date
+        run: |
+          cargo run --locked --all-features -- generate-config --output /tmp/opensovd-cda.toml
```
floroks, line 55: does this generate/compare a minimal configuration?

alexmohr: No it generates the full configuration, like it is currently committed, with optional fields that would not be serialized set to somewhat sensible defaults. 

The idea is to have the full reference configuration versioned. During Hackfest it was a bit annoying that there was no example config with all available options.

cda-main/src/config/generate.rs
```diff
@@ -0,0 +1,321 @@
...
+            })
+            .chain(std::iter::once(line.to_owned()))
+            .fold(String::new(), |mut acc, l| {
+                let _ = writeln!(acc, "{l}");
+                acc
+            });
+    }
+
+    let description = trimmed.find('=').and_then(|eq_pos| {
+        let key = trimmed[..eq_pos].trim();
```
theswiftfox: question: wouldn't that trip clippy due to indexing sclicing?
alternative could be `let key = trimmed.get(..eq_pos).map(str::trim)?;`

cda-main/src/main.rs
```diff
@@ -86,6 +99,30 @@ struct AppArgs {
...
+                std::io::stdout()
+                    .write_all(content.as_bytes())
+                    .map_err(|e| AppError::RuntimeError(format!("Failed to write stdout: {e}")))?;
+            }
+            Some(path) => {
+                std::fs::write(path, &content)
+                    .map_err(|e| AppError::RuntimeError(format!("Failed to write config: {e}")))?;
+            }
+            None => {
+                std::fs::write("opensovd-cda.toml", &content)
```
theswiftfox: suggestion: use `opensovd-cda.reference.toml` as a default - otherwise it might silently overwrite a userconfig named opensov-cda.toml if no --output is passed.
Or was that intentional, that it should then overwrite the config?

cda-interfaces/src/datatypes/com_params.rs
```diff
@@ -116,9 +151,19 @@ impl From<ComParamBool> for bool {
     }
 }
 
+impl schemars::JsonSchema for ComParamBool {
+    fn schema_name() -> std::borrow::Cow<'static, str> {
+        "ComParamBool".into()
+    }
+
+    fn json_schema(_generator: &mut schemars::SchemaGenerator) -> schemars::Schema {
+        serde_json::Map::from_iter([("type".to_owned(), "string".into())]).into()
```
theswiftfox: suggestion: adding an enum to the schema with the values ["true", "false"] (or any other accepted) to convey that this is a boolean and not a free text field?

cda-interfaces/src/datatypes/com_params.rs
```diff
@@ -16,17 +16,52 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer, de::DeserializeOwn
...
     pub doip: DoipComParams,
 }
 
 pub type ComParamName = String;
 
-#[derive(Deserialize, Serialize, Clone, Debug)]
+/// Schema helper for `ComParamConfig<Duration>` fields.
+/// Duration does not implement `JsonSchema`, so we provide a manual schema
+/// that matches the serialized shape of `ComParamConfig<Duration>`.
+fn duration_param_schema(_gen: &mut schemars::SchemaGenerator) -> schemars::Schema {
```
theswiftfox, line 33: suggestion: add a minimal test that serializes a `ComParamConfig<Duration>` and validates it against `duration_param_schema` to catch any occurance where the serialization of the comparam changes silently (dependency update, new toolchain, ...)

cda-main/src/config/generate.rs
```diff
@@ -0,0 +1,321 @@
...
+
+use std::{collections::BTreeMap, fmt::Write};
+
+use cda_interfaces::{FunctionalDescriptionConfig, datatypes::FaultConfig};
+
+use crate::config::configfile::Configuration;
+
+/// Create a Configuration instance with example values for fields that default to `None`.
+/// This ensures they appear in the generated reference config output.
+fn reference_config_instance() -> Configuration {
```
theswiftfox, line 21: question: is there an easy way to check if there are any new options added to the config but not here? Maybe via comparing to the schema?
Ideally we could then have a test for this, so we don't forget to update here 🤔

theswiftfox (commented):
Some small findings, and a few broader questions that didn't fit a specific place in the code. I was in a very nitpicky mood this morning 😂

(Q..Question, S..Suggestion)

- Q: do we want to keep some form of minimal config as reference as well, or only the full config? (eg. have a opensovd-cda.toml with a minimal set and then opensovd-cda.reference.toml with the full blown config as it is generated here?)

- Q: Do we want to enforce that all config options have a description? (We could add a test for that that asserts every property has a description)

- S: Adding a local test that compares the `generate_reference_config()` against `include_str!("opensovd-cda.reference.toml")` to locally catch that there is a drift already. 

- S: Small test that validates `generate_reference_config()` can also be parsed back as a valid config in the cda (or did i miss that when looking over the code here?)

alexmohr:
>     * Q: do we want to keep some form of minimal config as reference as well, or only the full config? (eg. have a opensovd-cda.toml with a minimal set and then opensovd-cda.reference.toml with the full blown config as it is generated here?)

I don't see the point of the 'minimal' config, because it's not a 'minimal' config that is required to start the CDA. It is starting without _any_ config just fine. 
Also the config we had was mainly setting things you could set via command line instead.


>     * Q: Do we want to enforce that all config options have a description? (We could add a test for that that asserts every property has a description)

That's a great idea. I would enforce this, because it makes the config much easier to understand.
If you don't want the doc strings on prod, you can roll your own config, but the config here, at least to me, should serve as documentation.

>     * S: Adding a local test that compares the `generate_reference_config()` against `include_str!("opensovd-cda.reference.toml")` to locally catch that there is a drift already.

Good idea. I would remove the pipeline step though when we have a test that validates this, we don't have to check it twice.

>     * S: Small test that validates `generate_reference_config()` can also be parsed back as a valid config in the cda (or did i miss that when looking over the code here?)

Whops, I'll add that as well. Haven't thought about that at all 🙈

cda-main/src/main.rs
alexmohr: I wanted to overwrite the existing config, but I see your point. I changed this to return an error instead of silently picking any name, it's the cleaner solution imho.

cda-main/src/config/generate.rs
alexmohr: well yes, but actually no 🙈 
So I will change this as you suggested, but clippy doesn't check strings in index slicing.

> This lint does not check for the usage of indexing or slicing on strings. These are covered
> by the more specific `string_slice` lint.
https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/indexing_slicing.rs#L27

--> I think we have another lint to enable :) (not in this PR though, if you agree w/ this I would create a new issue)

theswiftfox: Oooh, that explains it :D
yeah, new issue for it sounds good. thank you! :)

cda-interfaces/src/datatypes/com_params.rs
alexmohr: done

cda-interfaces/src/datatypes/com_params.rs
alexmohr: done :)

cda-main/src/config/generate.rs
alexmohr: added reference_config_covers_all_schema_fields

cda-main/src/config/generate.rs
alexmohr: https://github.com/eclipse-opensovd/classic-diagnostic-adapter/issues/333

theswiftfox (dismissed):

theswiftfox (approved):

PR #322, opened 2026-05-04
cda-extra: add systemd_notify

theswiftfox:
## Summary

This PR adds a cda-extra crate for additional features (inspired by axum-extra)

As a first feature, systemd-notify is implemented in cda-extra. If cda is running in an systemd environment systemd-notify is used to update the status according to the health state.

## Checklist

## Notes for Reviewers

---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ members = [
   "cda-tracing",
   "cda-plugin-security",
   "cda-health",
+  "cda-extra",
   "integration-tests",
   "cda-build",
   "comm-mbedtls/mbedtls-sys",
@@ -44,6 +45,7 @@ cda-plugin-security = { path = "cda-plugin-security" }
 cda-sovd = { path = "cda-sovd" }
 cda-tracing = { path = "cda-tracing" }
 cda-health = { path = "cda-health" }
+cda-extra = { path = "cda-extra" }
 sovd-interfaces = { path = "cda-sovd-interfaces" }
 cda-build = { path = "cda-build" }
 mbedtls-sys = { path = "comm-mbedtls/mbedtls-sys" }
@@ -59,6 +61,7 @@ serde_json = "1.0"
 tokio = { version = "1.48.0", default-features = false }
 tokio-util = { version = "0.7.3", default-features = false }
 futures = "0.3"
+sd-notify = "0.5.0"
 regex = "1.12.2"
 uuid = "1.18.1"
 bytes = "1"
--- a/cda-extra/Cargo.toml
+++ b/cda-extra/Cargo.toml
@@ -1,5 +1,5 @@
 # SPDX-License-Identifier: Apache-2.0
-# SPDX-FileCopyrightText: 2025 The Contributors to Eclipse OpenSOVD (see CONTRIBUTORS)
+# SPDX-FileCopyrightText: 2026 The Contributors to Eclipse OpenSOVD (see CONTRIBUTORS)
 #
 # See the NOTICE file(s) distributed with this work for additional
 # information regarding copyright ownership.
--- a/cda-extra/Cargo.toml
+++ b/cda-extra/Cargo.toml
@@ -1,0 +1,44 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: 2025 The Contributors to Eclipse OpenSOVD (see CONTRIBUTORS)
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+
+[package]
+name = "cda-extra"
+version = "0.1.0"
+edition.workspace = true
+publish = false
+description = "Optional extras for the Eclipse OpenSOVD CDA"
+homepage.workspace = true
+license.workspace = true
+
+[lints]
+workspace = true
+
+[lib]
+path = "src/lib.rs"
+
+[dependencies]
+# internal dependencies
+cda-health = { workspace = true, optional = true }
+cda-interfaces = { workspace = true, optional = true }
+
+# external
+sd-notify = { workspace = true, optional = true }
+tokio = { workspace = true, features = ["time"], optional = true }
+tracing = { workspace = true, optional = true }
+
+[features]
+default = []
+systemd-notify = [
+  "dep:sd-notify",
+  "dep:cda-health",
+  "dep:cda-interfaces",
+  "dep:tokio",
+  "dep:tracing",
+]
--- a/cda-extra/src/lib.rs
+++ b/cda-extra/src/lib.rs
@@ -1,0 +1,17 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: 2025 The Contributors to Eclipse OpenSOVD (see CONTRIBUTORS)
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ */
+
+#[cfg(feature = "systemd-notify")]
+mod systemd_notify;
+
+#[cfg(feature = "systemd-notify")]
+pub use systemd_notify::create_sd_notify_task;
--- a/cda-extra/src/lib.rs
+++ b/cda-extra/src/lib.rs
@@ -1,6 +1,6 @@
 /*
  * SPDX-License-Identifier: Apache-2.0
- * SPDX-FileCopyrightText: 2025 The Contributors to Eclipse OpenSOVD (see CONTRIBUTORS)
+ * SPDX-FileCopyrightText: 2026 The Contributors to Eclipse OpenSOVD (see CONTRIBUTORS)
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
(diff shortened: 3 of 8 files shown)


cda-extra/src/lib.rs
```diff
@@ -0,0 +1,17 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: 2025 The Contributors to Eclipse OpenSOVD (see CONTRIBUTORS)
```
alexmohr: nitpick: 2026 :)

cda-extra/src/systemd_notify.rs
```diff
@@ -0,0 +1,286 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: 2025 The Contributors to Eclipse OpenSOVD (see CONTRIBUTORS)
```
alexmohr: nitpick: 2026

cda-extra/src/systemd_notify.rs
```diff
@@ -0,0 +1,286 @@
...
+/// If systemd is not detected or the watchdog is not enabled, returns `None`
+/// and the CDA runs without `sd_notify` behavior.
+pub fn create_sd_notify_task<F>(
+    health_state: Option<HealthState>,
+    shutdown_signal: F,
+) -> Option<JoinHandle<()>>
+where
+    F: Future<Output = ()> + Clone + Send + 'static,
+{
+    if let Ok(true) = sd_notify::booted()
```
alexmohr: opinion: I think this would look a bit nicer, by returning early instead.

```rust
if sd_notify::booted() != Ok(true) {
    tracing::info!("Systemd not detected, skipping sd_notify initialization");
    return None;
}

let Some(interval) = determine_interval() else {
    tracing::info!("Watchdog not enabled, skipping sd_notify initialization");
    return None;
};

// init sd notify task
let watchdog_future = async move {
    let mut interval_timer = tokio::time::interval(interval);
    let mut state = cda_health::Status::Starting;
    loop {
        interval_timer.tick().await;
        state = trigger_watchdog(health_state.as_ref(), state).await;
    }
};

let sd_notify_future = spawn_named!("sd_notify", async move {
    tokio::select! {
        _ = watchdog_future => {},
        () = shutdown_signal => {},
    }
});

tracing::info!(
    "Systemd detected and watchdog enabled, initialized sd_notify task with interval of {:?} seconds",
    interval.as_secs()
);

Some(sd_notify_future)
```

cda-extra/src/systemd_notify.rs
```diff
@@ -0,0 +1,286 @@
...
+        Some(sd_notify_future)
+    } else {
+        tracing::info!(
+            "Systemd not detected or watchdog not enabled, skipping sd_notify initialization"
+        );
+        None
+    }
+}
+
+fn determine_interval() -> Option<Duration> {
```
alexmohr, line 73: suggestion: simplify
```rust
fn determine_interval() -> Option<Duration> {
    sd_notify::watchdog_enabled().map(|duration| {
        duration.checked_sub(Duration::from_secs(5)).unwrap_or(duration)
    })
}
```

alexmohr: Also there is a subtle bug here. If we set the timeout to >=5s, this will yield an interval of 0, turning this into a busy loop. Imho it would make sense to have a floor of 1s interval.

cda-extra/src/systemd_notify.rs
```diff
@@ -0,0 +1,286 @@
...
+    };
+    tracing::debug!("Triggering sd_notify watchdog with status {notify:?}");
+    if let Err(e) = sd_notify::notify(&[notify]) {
+        tracing::warn!(error = %e, "Failed to send sd_notify watchdog notification");
+    }
+    new_status
+}
+
+fn fold_status(acc: cda_health::Status, status: cda_health::Status) -> cda_health::Status {
+    match (acc, status) {
```
alexmohr, line 113: question: What do you think about `use Status::*;` here, would be a bit easier to read when `cda::health...` isn't repeated that much.

cda-extra/src/systemd_notify.rs
```diff
@@ -0,0 +1,286 @@
...
+        ) => cda_health::Status::Starting,
+        (cda_health::Status::Starting | cda_health::Status::Pending, cda_health::Status::Up) => {
+            cda_health::Status::Up
+        }
+        (_, status) => status,
+    }
+}
+
+async fn fold_health_state(health_state: Option<&HealthState>) -> cda_health::Status {
+    if let Some(health_state) = health_state {
```
alexmohr: opinion: (and thus totally optional)
```rust
    health_state
        .map_or(Status::Up, |state| async {
            state
                .query_all_providers()
                .await
                .into_values()
                .fold(Status::Starting, fold_status)
        })
        .await
```

cda-extra/Cargo.toml
```diff
@@ -0,0 +1,44 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: 2025 The Contributors to Eclipse OpenSOVD (see CONTRIBUTORS)
```
alexmohr: nitpick: 2026

.... we way want to tune the re-use script to complain about old year in new file :D

theswiftfox: oops - thats what I get for copy pasting without thinking 🙈
thank you :D

cda-extra/src/systemd_notify.rs
theswiftfox: oh, yeah absolutely! i can update the imports for this

cda-extra/src/systemd_notify.rs
theswiftfox: hm, the map_or doesn't want to take an async block:
```
     = note:       expected enum `cda_health::Status`
            found `async` block `{async block@cda-extra/src/systemd_notify.rs:137:43: 137:48}`
```
         
the next best solution I found:
```rust
    let Some(health_state) = health_state else {
        return cda_health::Status::Up;
    };
    health_state
        .query_all_providers()
        .await
        .into_values()
        .fold(cda_health::Status::Starting, fold_status)
```

alexmohr: thx, should have tested it, sorry :( I like your suggestion still better than the nested if :)

cda-extra/src/lib.rs
theswiftfox: fixed :D

cda-extra/src/systemd_notify.rs
theswiftfox: fixed

cda-extra/src/systemd_notify.rs
theswiftfox: yeah, better this way. went with the suggestion :)

cda-extra/Cargo.toml
theswiftfox: fixed now

alexmohr (approved):
lgtm

cda-extra/src/systemd_notify.rs
```diff
@@ -0,0 +1,279 @@
...
+        interval.as_secs()
+    );
+    Some(sd_notify_future)
+}
+
+fn determine_interval() -> Option<Duration> {
+    sd_notify::watchdog_enabled().map(|duration| {
+        // ensure we are sending a bit more often than the watchdog expects,
+        // to avoid any timing issues
+        if let Some(interval) = duration.checked_sub(Duration::from_secs(5)) {
```
lh-sag, line 77: I would follow the recommended "half the watchdog timeout" as documented here https://www.freedesktop.org/software/systemd/man/latest/sd_event_set_watchdog.html

IMHO i would favor something like  ```determine_wakeup_interval/watchdog_notify_interval``` instead of ```determine_interval``` to make the purpose of the interval clear.

cda-extra/src/systemd_notify.rs
```diff
@@ -0,0 +1,279 @@
...
+    })
+}
+
+async fn trigger_watchdog(
+    health_state: Option<&HealthState>,
+    prev_state: cda_health::Status,
+) -> cda_health::Status {
+    let new_status = fold_health_state(health_state).await;
+    let notify = match (prev_state, new_status) {
+        (cda_health::Status::Starting, cda_health::Status::Up) => sd_notify::NotifyState::Ready,
```
lh-sag, line 91: I am not familiar when the watchdog is armed on systemd side e.g. after ```Ready``` or right after it is spawned. But the readiness is a feature which can be used independent of watchdog configuration so in worst case this process will send a ready signal and miss to trigger the watchdog.

Cargo.toml
```diff
@@ -59,6 +61,7 @@ serde_json = "1.0"
 tokio = { version = "1.48.0", default-features = false }
 tokio-util = { version = "0.7.3", default-features = false }
 futures = "0.3"
+sd-notify = "0.5.0"
```
lh-sag, line 64: Optional: I usually tend to set ```default-features = false``` and list the required features where they are needed. This makes the binaries/crates considerable smaller.

cda-extra/src/systemd_notify.rs
```diff
@@ -0,0 +1,279 @@
...
+        let mut state = cda_health::Status::Starting;
+        loop {
+            interval_timer.tick().await;
+            state = trigger_watchdog(health_state.as_ref(), state).await;
+        }
+    };
+    let sd_notify_future = spawn_named!("sd_notify", async move {
+        tokio::select! {
+            _ = watchdog_future => {},
+            () = shutdown_signal => {},
```
lh-sag, line 62: Optional: You might want to signal ```Stopping``` to systemd

lh-sag (commented):
You might want to

theswiftfox:
@lh-sag oh, thanks for the feedback - I wasn't expecting another review, so it is already merged.
But I will take a look at the comments and potentially create a follow up PR for the findings you had :)

PR #324, opened 2026-05-04
fix(sovd): return 401 for unauthenticated requests to unregistered ECU paths

### LINKED ISSUE #302: [Bug] accessing data w/o token returns 404
## Description

Access 'operations' of an ECU without a valid token, lead to 404 instead of 401

## Steps to Reproduce

See above

## Expected Behavior

401

## Actual Behavior

404

Add logs, screenshots, or crash traces that help explain the problem.

## Environment

- OS: [e.g. Ubuntu 22.04, Windows 11]
- Version: [e.g. v0.2.1, main]

## Additional Context

> [!IMPORTANT]

shegazyy:
Summary
The fallback 404 handler was returning 404 unconditionally for any unmatched route, including unauthenticated requests. This allowed clients to probe ECU existence — 401 meant the ECU was registered, 404 meant it wasn't.

Fix: sovd_not_found_handler now checks for the Authorization header and returns 401 (no body) when absent, per SOVD 6.15.6. Authenticated requests to genuinely missing resources still receive 404.

Also corrected the href casing in the GET /vehicle/v15/components response (/Vehicle/ → /vehicle/) to be consistent with actual registered routes.

Checklist
 I have tested my changes locally
 I have linked related issues or discussions
Related
Closes #302

Notes for Reviewers
The original issue reported 404 for a valid ECU path without a token, this was not reproducible (that path already returns 401 correctly). The fix targets the related case: unregistered paths returning 404 before any auth check runs. Steps to verify:

# Before fix: 404. After fix: 401
curl http://localhost:20002/vehicle/v15/components/doesnotexist/operations

# Unchanged — still 401
curl http://localhost:20002/vehicle/v15/components/flxc1000/operations

# Unchanged — still 200
curl -H "Authorization: Bearer $TOKEN" http://localhost:20002/vehicle/v15/components/flxc1000/operations

### DIFF
--- a/cda-sovd/src/sovd/error.rs
+++ b/cda-sovd/src/sovd/error.rs
@@ -18,7 +18,7 @@ use axum::{
         Request,
         rejection::{JsonRejection, QueryRejection},
     },
-    http::{StatusCode, Uri},
+    http::{HeaderMap, StatusCode, Uri, header},
     middleware::Next,
     response::{IntoResponse, Response},
 };
@@ -409,7 +409,10 @@ pub(crate) async fn sovd_method_not_allowed_handler(
     }
 }
 
-pub(crate) async fn sovd_not_found_handler(uri: Uri) -> impl IntoResponse {
+pub(crate) async fn sovd_not_found_handler(headers: HeaderMap, uri: Uri) -> impl IntoResponse {
+    if !headers.contains_key(header::AUTHORIZATION) {
+        return StatusCode::UNAUTHORIZED.into_response();
+    }
     (
         StatusCode::NOT_FOUND,
         Json(
@@ -423,4 +426,5 @@ pub(crate) async fn sovd_not_found_handler(uri: Uri) -> impl IntoResponse {
             },
         ),
     )
+        .into_response()
 }
--- a/cda-sovd/src/sovd/mod.rs
+++ b/cda-sovd/src/sovd/mod.rs
@@ -459,9 +459,10 @@ async fn get_components<T: UdsEcu + SchemaProvider + Clone>(
     WithRejection(Query(query), _): WithRejection<Query<IncludeSchemaQuery>, ApiError>,
 ) -> Response {
     fn ecu_to_resource(ecu: String) -> Resource {
+        let ecu_lower = ecu.to_lowercase();
         Resource {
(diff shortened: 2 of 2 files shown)


cda-sovd/src/sovd/error.rs
```diff
@@ -409,7 +409,10 @@ pub(crate) async fn sovd_method_not_allowed_handler(
     }
 }
 
-pub(crate) async fn sovd_not_found_handler(uri: Uri) -> impl IntoResponse {
+pub(crate) async fn sovd_not_found_handler(headers: HeaderMap, uri: Uri) -> impl IntoResponse {
+    if !headers.contains_key(header::AUTHORIZATION) {
```
alexmohr, line 413: Question: Some routes can be accessed without a token. I think we should only return a only 401 for routes that actually require a token and not generally when a token is missing.
@floroks what's your take on this?

floroks: The existence of ECUs isn't a secret (get components isn't protected), what needs to be protected, is the list of data/operations without a token (e.g. only services viewable to anybody are listed). 

When the client already knows the exact path, there should be a distinction between 404 and 403/401, so clients can determine if their permissions, or the permissions in the diagnostic description are wrong, or if the ecu doesn't support that feature. Similarly, if you'd sent the request to an ECU, it'd respond either with NRC 34 (not authorized) vs. NRC 31 (Not Found).

In the end it's probably a behavior that should be configurable (e.g. all endpoints need token vs. it depends on the permissions of the specific endpoint), either through a configuration flag, or through a supplier specific plugin that does an additional check/transformation of the result.

alexmohr: @shegazyy Sorry for the delay. I saw you're message on slack but hadn't had the time to reply yet. I totally missed the aspect @floroks commented, that the ECU names are not a secret, so the current behavior, to return a 404 on an invalid ECU name, is correct. Therefore we won't change this to 401. 
If you're interested in implementing the configurable behavior @floroks mentioned in his comment, feel free to open an issue and we will discuss details there before jumping into implementation.

alexmohr:
As mentioned in https://github.com/eclipse-opensovd/classic-diagnostic-adapter/pull/324#discussion_r3257333683 closing this as working as designed.

PR #340, opened 2026-05-13
Storage-API: define api traits & reference impl

theswiftfox:
## Summary

Define Rust traits for the storage API according to the requirement & architecture documentation.
Add a reference implementation with tests for local storage access.

## Checklist

## Notes for Reviewers

---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ members = [
   "cda-plugin-security",
   "cda-health",
   "cda-extra",
+  "cda-storage",
   "integration-tests",
   "cda-build",
   "comm-mbedtls/mbedtls-sys",
@@ -47,6 +48,7 @@ cda-plugin-security = { path = "cda-plugin-security" }
 cda-sovd = { path = "cda-sovd" }
 cda-tracing = { path = "cda-tracing" }
 cda-health = { path = "cda-health" }
+cda-storage = { path = "cda-storage" }
 cda-extra = { path = "cda-extra" }
 sovd-interfaces = { path = "cda-sovd-interfaces" }
 cda-build = { path = "cda-build" }
@@ -91,10 +93,15 @@ ouroboros = "0.18.5"
 toml = "0.9.8"
 cargo_toml = "0.22.3"
 libc = "0.2.177"
+crc32fast = "1.4"
 foldhash = "0.2.0"
 memmap2 = "0.9"
 mockall = "0.14.0"
 sha2 = "0.10"
+rkyv = { version = "0.8.16", default-features = false, features = [
+  "std",
+  "bytecheck",
+] }
 
 # ---- tracing & logging crates ----
 tracing = "0.1.41"
--- a/cda-interfaces/src/storage_api/collection.rs
+++ b/cda-interfaces/src/storage_api/collection.rs
@@ -1,0 +1,179 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ * SPDX-FileCopyrightText: 2026 The Contributors to Eclipse OpenSOVD (see CONTRIBUTORS)
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ */
+
+use std::{
+    path::{Path, PathBuf},
+    sync::Arc,
+};
+
+use super::{
+    error::StorageError,
+    io::{RandomAccessData, ReadableStream},
+    transaction::Transaction,
+};
+
+/// A named grouping of key-value entries in the storage system.
+///
+/// Well-known collection names are represented as enum variants. Custom names are supported
+/// via [`CollectionName::Custom`].
+#[derive(Debug, Clone, PartialEq, Eq, Hash, rkyv::Archive, rkyv::Serialize, rkyv::Deserialize)]
+pub enum CollectionName {
+    /// The primary diagnostic database collection.
+    DiagnosticDatabase,
+    /// Staging area for a pending database update.
+    DiagnosticDatabaseNextUpdate,
+    /// Backup of the diagnostic database before an update.
+    DiagnosticDatabaseBackup,
+    /// A user-defined collection with an arbitrary name.
+    Custom(String),
+}
+
+impl CollectionName {
+    /// Returns a string representation suitable for use as a directory or bucket name.
+    #[must_use]
+    pub fn as_str(&self) -> &str {
+        match self {
+            Self::DiagnosticDatabase => "diagnostic_database",
+            Self::DiagnosticDatabaseNextUpdate => "diagnostic_database_next_update",
+            Self::DiagnosticDatabaseBackup => "diagnostic_database_backup",
+            Self::Custom(name) => name.as_str(),
+        }
+    }
+}
+
+impl std::fmt::Display for CollectionName {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(self.as_str())
+    }
+}
+
+/// Metadata about a single entry in a collection.
+#[derive(Debug, Clone)]
+pub struct Metadata {
+    /// The key (name) of the entry.
+    pub name: String,
+    /// The size of the stored data in bytes.
+    pub data_size: u64,
+    /// Optional custom properties associated with this entry.
+    pub custom_props: Vec<MetadataProperty>,
+}
+
+/// A single custom key-value property attached to entry metadata.
+#[derive(Debug, Clone)]
+pub struct MetadataProperty {
+    /// The property key.
+    pub key: String,
+    /// The property value.
+    pub value: String,
+}
+
+/// A named collection of key-value entries in the storage system.
+///
+/// Read operations do not require a transaction and always return committed state.
+/// Mutation operations require an exclusive `&mut Transaction` reference.
+///
+/// All keys are to be treated **case-insensitively**
+/// Implementations must normalize keys (e.g., `to_lowercase`) before any storage operation.
+pub trait Collection: Send + Sync {
+    /// Read the data for the given key, returning a random-access handle.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`StorageError::KeyNotFound`] if the key does not exist.
+    fn read(
+        &self,
+        key: &str,
+    ) -> impl Future<Output = Result<Arc<impl RandomAccessData + 'static>, StorageError>> + Send;
+
+    /// Retrieve metadata for the given key.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`StorageError::KeyNotFound`] if the key does not exist.
+    fn metadata(&self, key: &str) -> impl Future<Output = Result<Metadata, StorageError>> + Send;
+
+    /// List all keys in this collection.
+    ///
+    /// The returned keys are in their normalized (lowercase) form.
+    fn list(&self) -> impl Future<Output = Result<Vec<String>, StorageError>> + Send;
+
+    /// Return the number of entries in this collection.
+    fn len(&self) -> impl Future<Output = Result<usize, StorageError>> + Send;
+
+    /// Return whether this collection is empty.
+    fn is_empty(&self) -> impl Future<Output = Result<bool, StorageError>> + Send;
+
+    /// Write data from a [`ReadableStream`] to the given key within a transaction.
+    ///
+    /// The data is streamed to a staging file and only applied on [`Transaction::commit`]. If a
+    /// key with the same name already exists, it will be overwritten on commit.
+    ///
+    /// # Errors
+    ///
+    /// Returns a [`StorageError`] if the staging write fails.
+    fn write(
+        &self,
+        tx: &mut Transaction,
+        key: &str,
+        data: &mut impl ReadableStream,
+    ) -> impl Future<Output = Result<(), StorageError>> + Send;
+
+    /// Mark a key for deletion within a transaction.
+    ///
+    /// The deletion is only applied on [`Transaction::commit`].
+    ///
+    /// # Errors
+    ///
+    /// Returns [`StorageError::KeyNotFound`] if the key does not exist.
+    fn delete(
+        &self,
+        tx: &mut Transaction,
+        key: &str,
+    ) -> impl Future<Output = Result<(), StorageError>> + Send;
+
+    /// Mark all keys in this collection for deletion within a transaction.
+    ///
+    /// The deletions are only applied on [`Transaction::commit`].
+    fn delete_all(
+        &self,
+        tx: &mut Transaction,
+    ) -> impl Future<Output = Result<(), StorageError>> + Send;
+}
+
+/// Provides direct filesystem access to collection entries.
+///
+/// This trait is intended for consumers that need long-lived, unguarded access to the underlying
+/// files, typically for memory-mapped I/O. Unlike [`Collection::read`], these methods do **not**
+/// acquire a read lock, so the caller is responsible for coordinating with transactions (e.g.,
+/// evicting cached memory maps before committing an update that touches the collection).
+///
+/// Only implemented by storage backends that are backed by a real filesystem.
+pub trait DirectFileAccess: Send + Sync {
+    /// Returns the base directory path for this collection.
+    ///
+    /// Useful for callers that need to discover all files in the collection (e.g., enumerating
+    /// MDD databases at startup for memory-mapped loading).
+    fn dir_path(&self) -> &Path;
+
+    /// Returns the filesystem path for the given key.
+    ///
+    /// The key is normalized (lowercased) before lookup.
+    ///
+    /// This does **not** acquire a read lock. The caller must ensure the file is not concurrently
+    /// modified (e.g., by evicting cached handles before committing a transaction that touches
+    /// this key).
+    ///
+    /// # Errors
+    ///
+    /// Returns [`StorageError::KeyNotFound`] if the key does not exist in the collection.
+    fn file_path(&self, key: &str) -> Result<PathBuf, StorageError>;
+}
--- a/cda-interfaces/src/storage_api/collection.rs
+++ b/cda-interfaces/src/storage_api/collection.rs
@@ -33,6 +33,8 @@ pub enum CollectionName {
     DiagnosticDatabaseNextUpdate,
     /// Backup of the diagnostic database before an update.
     DiagnosticDatabaseBackup,
+    /// Configuration collection for CDA settings.
+    Configuration,
     /// A user-defined collection with an arbitrary name.
     Custom(String),
 }
@@ -45,6 +47,7 @@ impl CollectionName {
             Self::DiagnosticDatabase => "diagnostic_database",
             Self::DiagnosticDatabaseNextUpdate => "diagnostic_database_next_update",
             Self::DiagnosticDatabaseBackup => "diagnostic_database_backup",
+            Self::Configuration => "configuration",
             Self::Custom(name) => name.as_str(),
         }
     }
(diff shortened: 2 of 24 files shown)


lh-sag:
@copilot+claude-opus-4.6 could you please cross check this implementation against the spec here https://github.com/eclipse-opensovd/classic-diagnostic-adapter/commit/576287b475de266132c9f0942ada9754e67749bf latest doc is available here https://github.com/eclipse-opensovd/classic-diagnostic-adapter/tree/main/docs

Besides the cross check please do an addition code review.

theswiftfox:
I'm fixing the windows build & test first. No need for automated review yet.

theswiftfox:
Pipeline is happy again :) So ready for review now.

lh-sag:
FYI: @theswiftfox just giving copilot a try. Wont raise any review requests from now on.

cda-interfaces/src/storage_api/storage.rs
```diff
@@ -0,0 +1,97 @@
...
+ * https://www.apache.org/licenses/LICENSE-2.0
+ */
+
+use std::sync::Arc;
+
+use super::{
+    collection::{Collection, CollectionName},
+    error::StorageError,
+    transaction::Transaction,
+};
```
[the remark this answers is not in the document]
theswiftfox: pipeline disagrees with you there copilot ;)

cda-interfaces/src/storage_api/collection.rs
```diff
@@ -0,0 +1,179 @@
...
+use std::{
+    path::{Path, PathBuf},
+    sync::Arc,
+};
+
+use super::{
+    error::StorageError,
+    io::{RandomAccessData, ReadableStream},
+    transaction::Transaction,
+};
```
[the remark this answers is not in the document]
theswiftfox: again. this is rust 2024. no need for import here.

theswiftfox:
I let claude check the latest commit agains the findings reported by copiloy @lh-sag ;)
Should be all resolved now.
And I don't mind the AI review in general - but for my taste copilot (and AI in general) can be a bit verbose on the PR, so I prefer to do a local check of a PR with a coding agent myself and then either push the fixes (if its my own PR) or post a filtered review of the findings I confirmed.  
If you prefer a different approach here, I think we can (and maybe should) align it projectwide via the architecture board? :)
```
Summary: All Copilot findings - Resolved or Incorrect
#	Finding	Status	Details
1	Future not imported in storage.rs	                Incorrect finding	Rust 2024 edition includes Future in the prelude. The pipeline confirmed this compiles fine.
2	Future not imported in collection.rs	            Incorrect finding	Same as above -- Rust 2024 prelude.
3	tx_active never reset on error in apply()	        Resolved	        tx_active.store(false, ...) is now called unconditionally after apply_inner returns (line 336), regardless of success or failure.
4	CopyCollection doesn't replace dest / no rollback	Resolved	        CopyCollection now validates paths, and recovery logic (remove_new_artifacts) handles cleanup of newly-created destination collections on rollback.
5	Directory traversal via collection_dir()	        Resolved	        New paths::sanitize_path_segment() (in cda-storage/src/paths.rs) rejects ., .., path separators, and null bytes. Called from collection_dir() at line 88-91.
6	Directory traversal via keys in LocalCollection	    Resolved	        key_path() (line 61) now calls paths::sanitize_path_segment(key)? before joining.
7	wrapping_add overflow in read_wal	                Resolved	        Changed to saturating_add, which naturally causes the subsequent bounds check (payload_end > data.len()) to trigger, resulting in a truncation break -- matching the documented behavior.
8	get_collection not coordinating with data_lock	    Resolved	        get_collection now acquires self.data_lock.read().await (line 99) before checking filesystem existence.
9	get_or_create_collection mutating without lock	    Resolved	        Now checks tx_active, acquires a write lock before create_dir_all, and re-checks existence after lock acquisition (lines 127-140).
10	Recovery relies only on .bak files	                Resolved	        WAL now has a COMMITTING status header. mark_committing() is called at the start of apply, and recovery detects WalStatus::Committing to roll back partially-applied operations even when no .bak files exist.
```

cda-interfaces/src/storage_api/collection.rs
```diff
@@ -0,0 +1,179 @@
...
+
+/// A named grouping of key-value entries in the storage system.
+///
+/// Well-known collection names are represented as enum variants. Custom names are supported
+/// via [`CollectionName::Custom`].
+#[derive(Debug, Clone, PartialEq, Eq, Hash, rkyv::Archive, rkyv::Serialize, rkyv::Deserialize)]
+pub enum CollectionName {
+    /// The primary diagnostic database collection.
+    DiagnosticDatabase,
```
alexmohr, line 31: question: It _could_ be necessary to update the CDA configuration alongside a new MDD file. Should we plan for this too?
I.e. to set ECU specific settings or overwrites.

cda-storage/src/local_collection.rs
```diff
@@ -0,0 +1,243 @@
...
+}
+
+impl Collection for LocalCollection {
+    async fn read(&self, key: &str) -> Result<Arc<impl RandomAccessData + 'static>, StorageError> {
+        let key = normalize_key(key);
+        let path = self.key_path(&key)?;
+
+        let guard = io::acquire_read_lock(&self.data_lock).await;
+
+        let file = std::fs::File::open(&path).map_err(|e| {
```
alexmohr: nitpick: error handling is duplicated in `metadata`

cda-storage/src/wal.rs
```diff
@@ -0,0 +1,381 @@
...
+///
+/// # Errors
+///
+/// Returns a [`StorageError`] if encoding or writing to the WAL file fails.
+pub fn append_operation(wal_path: &Path, op: &Operation) -> Result<(), StorageError> {
+    let payload =
+        rkyv::to_bytes(op).map_err(|e: rkyv::rancor::Error| StorageError::Other(e.to_string()))?;
+
+    let checksum = compute_checksum(&payload);
+    let payload_len: u32 = payload
```
alexmohr, line 96: nitpick: In other places we allow u64 for sizes, but here the payload len is limited to u32. Although unlikely, that we will use more, the sizes should be aligned. 
I.e. ` fn data_size(&self) -> Result<u64, StorageError>;`

cda-storage/src/wal.rs
```diff
@@ -0,0 +1,381 @@
...
+
+        let stored_checksum = u32::from_le_bytes(checksum_bytes);
+        let payload_len = u32::from_le_bytes(len_bytes) as usize;
+
+        let payload_start = offset.saturating_add(ENTRY_HEADER_SIZE);
+        let payload_end = payload_start.saturating_add(payload_len);
+
+        if payload_end > data.len() || payload_end <= payload_start {
+            // Truncated payload -> incomplete entry, stop here.
+            tracing::warn!(
```
alexmohr, line 292: question: Shouldn't we return an error when the payload is truncated? Instead we return this as 'ok' and the caller might not notice that the payload was truncated

cda-storage/src/local_collection.rs
```diff
@@ -0,0 +1,243 @@
...
+        tx.record(op);
+
+        Ok(())
+    }
+
+    async fn delete(&self, tx: &mut Transaction, key: &str) -> Result<(), StorageError> {
+        let key = normalize_key(key);
+
+        // Verify the key exists in committed state.
+        let path = self.key_path(&key)?;
```
alexmohr, line 164: question: Do we have to protect against file modifications made on the filesystem directly?

1. Write a key and commit it so it exists on disk.
2. Begin a new transaction and record a delete.
3. Externally remove the file between delete() and commit().
4. The commit operation will succeed even though the file was not removed through the transaction and was already gone. 

I don't think we have to protect against these in the real world, but we probably should document this?

cda-interfaces/src/storage_api/error.rs
```diff
@@ -0,0 +1,43 @@
...
+    #[error("Transaction error: {0}")]
+    TransactionError(String),
+
+    /// The storage backend has no space left for the requested operation.
+    #[error("No space left: {0}")]
+    NoSpaceLeft(String),
+
+    /// An I/O error occurred during a storage operation.
+    #[error("I/O error: {0}")]
+    Io(#[from] std::io::Error),
```
alexmohr, line 38: suggestion: Add this error to docs/03_architecture/01_general.rst

cda-interfaces/src/storage_api/collection.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+}
+
+/// Metadata about a single entry in a collection.
+#[derive(Debug, Clone)]
+pub struct Metadata {
+    /// The key (name) of the entry.
+    pub name: String,
+    /// The size of the stored data in bytes.
+    pub data_size: u64,
```
alexmohr, line 68: suggestion: Align this size with Collection::len. The docs, state usize for this field as well.

cda-storage/tests/local_storage_tests.rs
```diff
@@ -0,0 +1,778 @@
...
+    ));
+    let read_op_2 = read_ops.get(2).expect("Expected 3 operations");
+    assert!(matches!(
+        read_op_2,
+        cda_interfaces::storage_api::Operation::Delete { key, .. }
+        if key == "old_key"
+    ));
+}
+
+#[tokio::test]
```
alexmohr, line 744: bug: remove_new_artifacts processes operations in order. When a transaction both
creates a collection and writes into it, the CreateCollection check runs first —
at that point the directory is NOT empty (the Write operation file is still there).
So CreateCollection skips removal. Then the Write branch removes the file (no .bak).
After all operations are processed the directory is empty but was never re-checked,
leaving an orphaned empty directory. rollback_partial_commit also won't touch it
because no .bak directory exists.

```rust

#[tokio::test]
async fn recovery_create_collection_with_write_leaves_orphaned_dir() {
    let dir = tempfile::tempdir().unwrap();
    let root = dir.path();
    let collections_dir = root.join("collections");
    let journal_dir = root.join("journal");
    let staging_dir = journal_dir.join("staging");
    std::fs::create_dir_all(&collections_dir).unwrap();
    std::fs::create_dir_all(&staging_dir).unwrap();

    // Simulate partial commit state: the collection directory was created AND a file was
    // written into it before the crash.
    let new_col_dir = collections_dir.join("fresh_collection");
    std::fs::create_dir_all(&new_col_dir).unwrap();
    std::fs::write(new_col_dir.join("data_file"), b"written during commit").unwrap();

    // WAL with COMMITTING status containing both operations in natural order:
    // CreateCollection first, then Write to that collection.
    let wal_path = journal_dir.join("transaction.wal");
    cda_storage::wal::create_wal(&wal_path).unwrap();
    cda_storage::wal::append_operation(
        &wal_path,
        &cda_interfaces::storage_api::Operation::CreateCollection {
            name: CollectionName::Custom("fresh_collection".to_string()),
        },
    )
    .unwrap();
    cda_storage::wal::append_operation(
        &wal_path,
        &cda_interfaces::storage_api::Operation::Write {
            collection: CollectionName::Custom("fresh_collection".to_string()),
            key: "data_file".to_string(),
            staged_path: "/tmp/irrelevant.tmp".to_string(),
        },
    )
    .unwrap();
    cda_storage::wal::mark_committing(&wal_path).unwrap();

    // Recovery should fully roll back — the collection did not exist before this transaction.
    let _storage = LocalStorage::new(root).unwrap();

    // The collection directory must be completely gone after recovery.
    // FAILS: current implementation leaves the empty directory orphaned because
    // CreateCollection's is_dir_empty check runs before Write's file removal.
    assert!(
        !new_col_dir.exists(),
        "Orphaned empty collection directory was left behind after recovery. \
         remove_new_artifacts checks CreateCollection (dir not empty → skip) before \
         the Write branch removes the file, so the now-empty dir is never cleaned up."
    );
}

```

Cargo.toml
```diff
@@ -91,10 +93,15 @@ ouroboros = "0.18.5"
 toml = "0.9.8"
 cargo_toml = "0.22.3"
 libc = "0.2.177"
+crc32fast = "1.4"
 foldhash = "0.2.0"
 memmap2 = "0.9"
 mockall = "0.14.0"
 sha2 = "0.10"
+rkyv = { version = "0.8.16", default-features = false, features = [
```
alexmohr, line 101: suggestion: It would be cool to have a short ADR that describes the WAL format we use on disk and a brief explainer why having a binary WAL is prefered over a verbose text based format. Mainly asking because I had to lookup what's cool about rkyv

cda-storage/tests/local_storage_tests.rs
theswiftfox: Oh god - thanks for catching that! I forgot to do the rollback in reverse order. By adding `iter().rev()` when iterating over the operations in remove_new_artifacts the directory gets cleaned up now. (Rest of the tests still pass as well)
That was the behavior I initially intended, but forgot somewhere in the middle :D
edit: I added the test you posted (in a slightly differing form for the last assert) as well :)

Cargo.toml
theswiftfox: Oh sure, good idea! :)
PS: main reason for rkyv is that bincode is (no longer maintained)[https://crates.io/crates/bincode]. I initially intended to use bincode for having the wal as fast & small as possible, but swapped to rkyv as it was mentioned in the bincode readme as good alternative

But yea, having something explaining the format and reasoning for it is probably a good idea.

cda-storage/src/local_collection.rs
theswiftfox: Hmm, good question. It could make sense to error the transaction on applying the delete if the file isn't there anymore when the delete happens. 
But at the same time, I think in the default implementation of the storage api we should keep it rather simple. And ensuring all external modifications are caught and handled can be quite the rabbit hole. So I'm also inclined on erring on the side of 'luck accident' here and make the transaction go through in that case.

I'd see the full protection against file modifications more in a specialised implementation of the file storage api, if there is a need for it, and would hope that most people try to avoid this from happening by having a system set up with permissions in a way, that no other process should be able to modify the storage the cda uses ^^

theswiftfox: As for documenting it, we can. I'd just add a comment for the absent 'else' branch explaining that the delete will succeed in that case?

cda-storage/src/wal.rs
theswiftfox: Yup. I accidentally changed that when rewriting the original wal. Thinking more about it, I think it needs to be changed a bit further to be as robust as possible. I'm thinking
- mark the wal as errored (with maybe the entry # for logging)
- try to continue reading the wal, by seeking the next entry header (maybe need to extend the entry header with a magic byte to identify it)
- return all operations that could be successfully parsed from wal

Recovery can then undo all operations that were returned, and do an additional error handling (like best effort cleanup for orphaned collections, etc.).

Your thoughts on that? :) Am I still missing something, that could be done better with this approach?

cda-interfaces/src/storage_api/collection.rs
theswiftfox: hm, we will have to brainstorm if its easier to have an option for providing additional 'runtime' configuration via storage api (the cda will need an initial fixed configuration, so it can initialise the update plugin with the correct paths, I imagine) OR if we rather want to handle these things via persistence api and inject the config key - values from there after startup.
Definitely something we should plan out and write down a bit of architecture thoughts :)

Edit:
If you think the strings in the operation enum are a concern, I'd rather check the collection names, keys etc for a specific max length and disalowing unreasonable long string on that layer, as they will probably cause many more problems down the line, when using them as part of the paths ;)

cda-storage/src/wal.rs
theswiftfox: the reason for chosing an u32 here is due to this being the wal file format. (i really should have put in a short document describing it xD)
a payload in the wal can be at most 12-13KB, assuming all strings are PATH_MAX in the collection names and keys, but in the regular case it should be way less. meaning it could be encoded in an u16 as well. However, rkyv wants 4 byte alignment, therefore simply using an u32 is the simpler choice than using u16 + padding.
u64 would work, but would waste another 4 bytes, that are never practically used.

cda-interfaces/src/storage_api/collection.rs
theswiftfox: I did a double check, the rust std lib returns an u64 for the size of a file on **all** targets: https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.len
-> I'd suggest using u64 here as well and updating the docs to use u64 as well. Otherwise we'd risk truncation of file lengths on 32 bit targets (however unlikely they may be)

As for aligning with Collection::len - I'd leave Collection::len as usize, because that is an in memory length and therefore makes sense to use the platform size type.

cda-storage/src/wal.rs
theswiftfox: I thought a bit more about this, and I think I overcomplicated it with my suggestion.
The main thing is to make the caller aware of the failure.

The corruption can either happen due to a crash when writing the wal / fsyncing it (this is **before** any operations are applied) or due to filesystem corruption.
For the first one, we the recovery can simply return an error due to corrupted wal, meaning the whole transaction has to be retriggered.
For the second one, I don't think we have a reasonable recovery strategy here, besides simply trying again. The recovery could compare the filesystem state to the operations it could read from the wal to determine if it was partially written (eg identify the corruption), so the caller can know that the whole collection has to be rewritten potentially.

cda-storage/src/wal.rs
alexmohr: oh, ya that makes sense. Looking forward to the documentation 😆

cda-storage/src/wal.rs
alexmohr: I'm not sure if I would even offer the possibility to 'resume' the write operation in the second case. Implementation would be simpler. Worst case is that we waste write cycles on flash?

cda-storage/src/local_collection.rs
alexmohr: >  set up with permissions in a way, that no other process should be able to modify the storage the cda 

Ya exactly, that's why I'm saying we don't need this in the real world. 

> As for documenting it, we can. I'd just add a comment for the absent 'else' branch explaining that the delete will succeed in that case?

Works for me, thank you.

cda-storage/src/wal.rs
theswiftfox: Yeah, I'm thinking the same.
I landed on this approach in recovery.rs. should be quite simple, but cover the cases:
```
wal has flag COMMITTING + truncated + evidence of partial application (.bak files present for example) -> returns StorageError::Corruption
wal has flag COMMITTING + truncated + no evidence -> safe discard (transaction was lost before apply)
```
Then the caller can decide what to do with the corruption error - if they want to retry, or fallback to a well known good state (eg. factory state etc) and the storage api is transparent on what is happening.

cda-interfaces/src/storage_api/collection.rs
alexmohr: Sounds good, thx :)

cda-interfaces/src/storage_api/collection.rs
theswiftfox: As discussed in chat, will add the enum for `Configuration`. The details on how (and if) it is used, lies then in the update plugin - but the option for it is there :)
I'll also add it to the docs.

cda-storage/src/wal.rs
alexmohr: awesome, thank you :)

docs/04_adr/04_wal_format.rst
```diff
@@ -0,0 +1,160 @@
...
+   memory-mapped / read buffer without parsing or allocating.  Text formats
+   (JSON, TOML) require a full parse pass and allocate intermediate
+   structures.
+
+2. **Simple deserialization**.  The serialized bytes are the payload directly.
+   No text encoding layer (escaping, quoting, base64) sits between the raw
+   operation data and its on-disk representation.
+
+3. **Compact**.  Typical ``Operation`` payloads are < 1 KiB.  A JSON
+   equivalent with escaped strings, keys, and formatting would be 2--5x
```
alexmohr: nitpick:

`2-5x`

alexmohr (commented):
tiny nitpick :)

docs/04_adr/04_wal_format.rst
theswiftfox: 🤦‍♀️ oops, fixed xD thx!

alexmohr (approved):
lgtm :)

PR #342, opened 2026-05-13
feat(database): add strict_parameter_validation config flag, Rejects requests with undefined parameters (400) when enabled

### LINKED ISSUE #215: Improvement: Return an error if unexpected parameters are provided in request
## Description

When handling the JSON payload from the request, we should check that only the parameters that the UDS request accepts are provided in the request, and return an error otherwise.
This could also be something that can be configured, where we have the default 'lenient' mode that will only log a warning for this, and the 'strict' mode where it is strictly enforced that only the accepted parameters are provided.

## Motivation

I'd assume that a client adding parameters that are not relevant to the request is an error on the user side and means that they either intended to do something different (eg. different request) or do not know that the parameter they sent won't have an effect. In both cases it would be beneficial to let the user know that their request wasn't handled as expected.

## Proposed Solution

As mentioned above, check that only parameters allowed by the request are provided and then
- in lenient mode log warning and proceed as it is now
- in strict mode log warning and return an error to the caller (BAD request with the message containing the parameters that were unexpected)

## Additional Context

Related to the [discussion in #209](https://github.com/eclipse-opensovd/classic-diagnostic-adapter/pull/209#issuecomment-3932766265)

> [!IMPORTANT]

shegazyy:
Summary
Adds a new opt-in config flag database.strict_parameter_validation (default: false).

When false: requests containing parameters not defined in the service proceed normally. S warning is logged but the request is forwarded to the ECU.

When true: the request is rejected immediately with 400 Bad Request listing the unexpected parameter names, before any UDS payload is built or sent to the ECU.

Checklist
 I have tested my changes locally
 I have updated documentation
 I have linked related issues
 I have updated tests
Related
Closes #215

Notes for Reviewers
The flag is opt-in and defaults to false to preserve existing behavior. The check runs at the end of process_parameter_map in EcuManager after the existing parameter mapping loop, unexpected keys are detected by diffing the request's parameter map keys against the set of short_names defined in the ODX service.

The opensovd-cda.toml reference config documents the flag with its default. The integration test setup in runtime.rs explicitly sets false to keep existing tests unaffected.

To enable strict mode, set strict_parameter_validation = true under [database] in opensovd-cda.toml and restart the application. With strict mode on, a request containing an undefined parameter will be rejected before reaching the ECU with 400 Bad Request and a message of the form: "Bad payload: Unexpected parameters in request: [\"bogus_extra_param\"]".

Example
Request:

curl -X PUT http://localhost:20002/vehicle/v15/components/flxc1000/data/VINDataIdentifier \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"data": {"VIN": "SCEDT26T8BD005261", "bogus_extra_param": "unexpected"}}'
Expected response (strict_parameter_validation = true):

{
  "message": "Bad payload: Unexpected parameters in request: [\"bogus_extra_param\"]",
  "error_code": "vendor-specific",
  "vendor_code": "bad-request"
}
HTTP 400 Bad Request

Expected response (strict_parameter_validation = false):

Request goes through to the ECU normally, bogus_extra_param is silently ignored. Warning appears in logs.

### DIFF
--- a/cda-core/src/diag_kernel/ecumanager.rs
+++ b/cda-core/src/diag_kernel/ecumanager.rs
@@ -103,6 +103,23 @@ impl VariantData {
     }
 }
 
+/// Configuration options passed to [`EcuManager::new`].
+///
+/// Grouping these together keeps the constructor signature manageable and
+/// makes call sites self-documenting.
+#[derive(Copy, Clone)]
+pub struct EcuManagerConfig {
+    /// Whether this ECU represents a physical ECU (`Ecu`) or a functional
+    /// group description (`FunctionalDescription`).
+    pub type_: EcuManagerType,
+    /// When `true`, variant detection failures fall back to the base variant
+    /// instead of returning an error.
+    pub fallback_to_base_variant: bool,
+    /// When `true`, requests containing parameters not defined in the service
+    /// are rejected with a `BadPayload` error (400 Bad Request).
+    pub strict_parameter_validation: bool,
+}
+
 // Allowed because this holds a bunch of config values.
 #[allow(clippy::struct_excessive_bools)]
 pub struct EcuManager<S: SecurityPlugin> {
@@ -129,6 +146,7 @@ pub struct EcuManager<S: SecurityPlugin> {
     variant_index: Option<usize>,
     variant: EcuVariant,
     fallback_to_base_variant: bool,
+    strict_parameter_validation: bool,
     duplicating_ecu_names: Option<HashSet<String>>,
 
     protocol: Protocol,
@@ -2039,28 +2057,25 @@ impl<S: SecurityPlugin> EcuManager<S> {
         protocol: Protocol,
         com_params: &ComParams,
         database_naming_convention: DatabaseNamingConvention,
-        type_: EcuManagerType,
+        config: EcuManagerConfig,
         func_description_config: &cda_interfaces::FunctionalDescriptionConfig,
-        fallback_to_base_variant: bool,
     ) -> Result<Self, DiagServiceError> {
-        match type_ {
+        match config.type_ {
             EcuManagerType::Ecu => Self::new_ecu_description(
                 database,
                 protocol,
                 com_params,
                 database_naming_convention,
-                type_,
+                config,
                 func_description_config,
-                fallback_to_base_variant,
             ),
             EcuManagerType::FunctionalDescription => Self::new_functional_description(
                 database,
                 protocol,
                 com_params,
                 database_naming_convention,
-                type_,
+                config,
                 func_description_config,
-                fallback_to_base_variant,
             ),
         }
     }
@@ -2072,9 +2087,8 @@ impl<S: SecurityPlugin> EcuManager<S> {
         protocol: Protocol,
         com_params: &ComParams,
         database_naming_convention: DatabaseNamingConvention,
-        type_: EcuManagerType,
+        config: EcuManagerConfig,
         func_description_config: &cda_interfaces::FunctionalDescriptionConfig,
-        fallback_to_base_variant: bool,
     ) -> Result<Self, DiagServiceError> {
         let variant_detection =
             variant_detection::prepare_variant_detection(&database, &database_naming_convention)?;
@@ -2141,7 +2155,7 @@ impl<S: SecurityPlugin> EcuManager<S> {
         Ok(Self {
             db_cache: DbCache::default(),
             ecu_name,
-            description_type: type_,
+            description_type: config.type_,
             database_naming_convention,
             tester_address: database
                 .find_com_param(data_protocol_ref, &com_params.doip.logical_tester_address)?,
@@ -2178,7 +2192,8 @@ impl<S: SecurityPlugin> EcuManager<S> {
                 state: EcuState::NotTested,
                 logical_address: logical_ecu_address,
             },
-            fallback_to_base_variant,
+            fallback_to_base_variant: config.fallback_to_base_variant,
+            strict_parameter_validation: config.strict_parameter_validation,
             duplicating_ecu_names: None,
             protocol,
             fg_protocol_position: func_description_config.protocol_position.clone(),
@@ -2243,9 +2258,8 @@ impl<S: SecurityPlugin> EcuManager<S> {
         protocol: Protocol,
         com_params: &ComParams,
         database_naming_convention: DatabaseNamingConvention,
-        type_: EcuManagerType,
+        config: EcuManagerConfig,
         func_description_config: &cda_interfaces::FunctionalDescriptionConfig,
-        fallback_to_base_variant: bool,
     ) -> Result<Self, DiagServiceError> {
         // Functional group description: use defaults for all com params
         let logical_ecu_address = com_params.doip.logical_ecu_address.value;
@@ -2267,7 +2281,7 @@ impl<S: SecurityPlugin> EcuManager<S> {
             diag_database: database,
             db_cache: DbCache::default(),
             ecu_name,
-            description_type: type_,
+            description_type: config.type_,
             database_naming_convention,
             tester_address: com_params.doip.logical_tester_address.value,
             logical_address: logical_ecu_address,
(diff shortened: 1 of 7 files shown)


cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -3610,6 +3620,28 @@ impl<S: SecurityPlugin> EcuManager<S> {
             };
             self.map_param_to_uds(param, json_values.get(short_name), uds, parent_byte_pos)?;
         }
+
+        let expected_names: HashSet<&str> = mapped_params
```
alexmohr: suggestion: Move this before the `for` loop. Otherwise when 'strict' mode is enabled we build the whole UDS payload and then exit anyway. Moving this before the loop, saves a bit of CPU time in the error case.

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -3610,6 +3620,28 @@ impl<S: SecurityPlugin> EcuManager<S> {
...
+            .iter()
+            .filter_map(|p| p.short_name())
+            .collect();
+        let unexpected: Vec<&str> = json_values
+            .keys()
+            .filter(|k| !expected_names.contains(k.as_str()))
+            .map(String::as_str)
+            .collect();
+        if !unexpected.is_empty() {
```
alexmohr: suggestion: Kind of debatable but I would skip logging the warning and only build the vectors when strict_parameter_validation is enabled.

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -3610,6 +3620,28 @@ impl<S: SecurityPlugin> EcuManager<S> {
...
+            tracing::warn!(
+                unexpected_parameters = ?unexpected,
+                "Request contains parameters not defined in the service"
+            );
+            if self.strict_parameter_validation {
```
alexmohr: question: More of a general nature and directed into @floroks and @theswiftfox direction. Do we want to have this as an extra configuration flag or have a more overarching 'strict' mode which this check is only part of?

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -2019,6 +2021,7 @@ impl<S: SecurityPlugin> EcuManager<S> {
         type_: EcuManagerType,
         func_description_config: &cda_interfaces::FunctionalDescriptionConfig,
         fallback_to_base_variant: bool,
+        strict_parameter_validation: bool,
```
alexmohr: suggestion: (or follow up)  A dedicated config struct (e.g., EcuManagerConfig { fallback_to_base_variant, strict_parameter_validation, type_ }) would be cleaner

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -8040,6 +8074,62 @@ mod tests {
...
+    async fn test_process_parameter_map_unexpected_params_strict() {
+        let (mut ecu_manager, service, _, _) = create_ecu_manager_with_struct_service(1);
+        ecu_manager.strict_parameter_validation = true;
+
+        let struct_value = json!({"param1": 0x1234u32, "param2": 1.0f32, "param3": "hello"});
+        // bogus_param is at the top-level service parameter map, not defined in the service
+        let payload_data = UdsPayloadData::ParameterMap(
+            [
+                ("main_param".to_string(), struct_value),
+                ("bogus_param".to_string(), json!("should be rejected")),
```
alexmohr, line 8174: question: does the implementation only check top level params? 
We should make sure things like 
```json
{"main_param": {"param1": 1, "bogus_nested": 2}}
```
are caught too.

alexmohr (commented):
Thanks for the contribution :)

shegazyy:
Hi @alexmohr, thanks again for the review. I've pushed updates addressing all your inline suggestions, the latest commits cover the requested changes

Regarding the broader question on whether this should stay as its own flag or fold into an overarching 'strict' mode, happy to refactor in whichever direction the team prefers once @floroks & @theswiftfox weigh in. 

Otherwise ready for another look whenever you have time 🙏

alexmohr:
@shegazyy Thanks, for future reference, at least for my review, please let me resolve the threads myself, makes re-reviewing things a lot easier on my end :)

shegazyy:
@alexmohr Got it, sorry about that. Will leave the threads for you to resolve going forward 🙏

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -3558,20 +3573,35 @@ impl<S: SecurityPlugin> EcuManager<S> {
...
-                let short_name = param.short_name().ok_or_else(|| {
-                    DiagServiceError::InvalidDatabase(
-                        "Unable to find short name for param".to_owned(),
-                    )
-                })?;
+            .collect();
 
-                self.map_param_to_uds(&param, value.get(short_name), payload, struct_byte_pos)
-            })
+        if self.strict_parameter_validation {
```
alexmohr, line 3642: suggestion: same as in process_parameter_map, please move this to a helper function instead.

Also there is still a gap in detecting these fields, mux parameters can still have bogus keys without being detected, here is a test to validate this. 

```rust
 /// Strict mode must reject unexpected keys inside a Mux parameter object.
    ///
    /// `map_mux_to_uds` only reads `"Selector"` and the matched case name
    /// any extra keys are silently ignored even when
    /// `strict_parameter_validation` is enabled.  This test documents that gap:
    /// it **should** fail with `BadPayload`, mirroring the behaviour of
    /// `map_struct_to_uds` and `process_parameter_map`.
    #[tokio::test]
    async fn test_map_mux_to_uds_unexpected_params_strict() {
        let (mut ecu_manager, service, _) =
            create_ecu_manager_with_mux_service(None, None, None);
        ecu_manager.strict_parameter_validation = true;

        // Valid mux payload with an extra key "bogus_mux_key" at the mux
        // object level.  "Selector" and "mux_1_case_1" are the only
        // legitimate keys for this request.
        let test_value = json!({
            "mux_1_param": {
                "Selector": 5,
                "mux_1_case_1": {
                    "mux_1_case_1_param_1": 13.37_f32,
                    "mux_1_case_1_param_2": 7
                },
                "bogus_mux_key": "should be rejected"
            }
        });

        let payload_data =
            UdsPayloadData::ParameterMap(serde_json::from_value(test_value).unwrap());

        let result = ecu_manager
            .create_uds_payload(&service, &skip_sec_plugin!(), Some(payload_data), None)
            .await;

        assert!(
            result.is_err(),
            "Expected strict mode to reject unexpected mux-level key 'bogus_mux_key', \
             but the request succeeded"
        );
        if let Err(e) = result {
            assert!(
                e.to_string().contains("Unexpected parameters in request"),
                "Expected 'Unexpected parameters in request' error, got: {e}"
            );
        }
    }
```

alexmohr:
Added 'changes requested' because it needs a rebase onto main first.

shegazyy:
Hi @alexmohr, thanks for the review. All three points are addressed in the latest commits:

- Extracted the validation logic into a helper function, matching the `process_parameter_map` pattern.
- Closed the mux parameter gap so bogus keys inside mux params are now detected too; added your test, and it passes.
- Rebased onto main.

Ready for another look whenever you have time .

cda-core/src/diag_kernel/ecumanager.rs
floroks: I'm thinking an extra strict section in the config with individual flags, but also an "set all to true" flag.

alexmohr: Makes sense, the change @shegazyy made follows the pattern I already established with the strict_ecu_config. So I would suggest we keep it as is for now and do the "set all to true" flag in a follow up.

alexmohr: https://github.com/eclipse-opensovd/classic-diagnostic-adapter/issues/376 follow up issue, closing this thread.

alexmohr:
@shegazyy Great work, thank you. Please squash the commits, then we are ready to go :)

shegazyy:
@alexmohr commits are squashed, let me know if there's anything else to do. And thank you so much for the continuous support 🙌

alexmohr (approved):
lgtm, thank you for the contribution and resolving all findings :)

PR #365, opened 2026-06-01
feat(plugin): add diagnostic database update plugin (runtimefiles) 

alexmohr:
## Summary

Introduce the `cda-plugin-runtime-update` crate and the foundational
bulk-data endpoints for managing diagnostic database lifecycle
(runtimefiles-current, runtimefiles-nextupdate, runtimefiles-backup).
## Checklist

## Notes for Reviewers


alexmohr:
Todo: create followup for Software Requirement: Diagnostic Database Update Plugin - Downgrade Protection [req~plugin-diagnostic-database-update-downgrade-protection](https://eclipse-opensovd.github.io/classic-diagnostic-adapter/02_requirements/index.html#req~plugin-diagnostic-database-update-downgrade-protection)

cda-comm-uds/src/lib.rs
```diff
@@ -145,6 +169,40 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsMana
...
+                handle.abort();
+            }
+        }
+        {
+            let mut transfers = self.data_transfers.lock().await;
+            for (_, transfer) in transfers.drain() {
+                transfer.task.abort();
+            }
+        }
+    }
```
SizifWork, line 191: I think there should be a way to do this in one loop.

cda-main/src/mdd.rs
```diff
@@ -299,6 +299,95 @@ async fn load_mdd_paths_from_storage(storage_dir: &str) -> Option<Vec<PathBuf>>
         .collect())
 }
 
+/// Seeds the `DiagnosticDatabase` storage collection from `database_path` when the collection
+/// is empty. This copies all `.mdd` files from the filesystem into storage so that the runtime
+/// update plugin has a populated baseline to work with.
+pub async fn seed_storage_from_database_path(storage_dir: &str, database_path: &str) {
```
alexmohr, line 299: Todo: also needed for cfg

integration-tests/tests/sovd/runtimefiles.rs
```diff
@@ -1528,25 +1563,51 @@ async fn runtimefiles_apply_removes_ecu_routes() -> Result<(), TestingError> {
...
     // The reload_databases path shuts down the old UDS/gateway and rebuilds routes;
     // 5 s is sufficient for the integration-test environment.
     cda_interfaces::util::tokio_ext::sleep_for(Duration::from_secs(5)).await;
 
-    // Only FLXC1000.mdd was applied → FLXCNG1000's route no longer exists.
+    // FLXC1000 was removed from staging → its route no longer exists.
     send_cda_request(
         &runtime.config,
-        "components/flxcng1000",
+        "components/flxc1000",
```
alexmohr: Todo use the constant for the path

alexmohr: Same in 1622 and more

cda-sovd/src/sovd/apps/sovd2uds/bulk_data/runtimefiles.rs
```diff
@@ -0,0 +1,504 @@
...
+        State(route_state): State<RuntimeUpdateRouteState<P, L>>,
+        Path(id): Path<String>,
+    ) -> impl IntoResponse {
+        match route_state.plugin.get_execution_status(&id).await {
+            Some(exec) => (
+                StatusCode::OK,
+                Json(serde_json::json!({
+                    "id": exec.id,
+                    "mode": format!("{:?}", exec.mode),
+                    "status": format!("{:?}", exec.status),
```
floroks: question: When using a debug format for this property, it would output Failed(<text>), which is not super easy to parse. Wouldn't it make sense to separate this into "status" and some additional Information/reason field?

cda-main/src/update/security.rs
```diff
@@ -0,0 +1,419 @@
...
+                    VerificationError(format!("Invalid UTF-8 path: {}", path.display()))
+                })?;
+                mmap_and_decode_mdd(path_str).map_err(|e| {
+                    VerificationError(format!("Failed to parse MDD '{}': {e}", path.display()))
+                })?;
+            }
+            UpdateFileType::Config => {
+                let content = std::fs::read_to_string(path).map_err(|e| {
+                    VerificationError(format!("Failed to read config '{}': {e}", path.display()))
+                })?;
```
floroks: nitpick: should probably use `tokio::fs::read_to_string`

cda-main/src/update/mod.rs
```diff
@@ -0,0 +1,392 @@
...
+
+/// Reads, parses, and applies a TOML [`Configuration`] from `config_path` into `config`.
+///
+/// Extracted as a free function so that it can be unit-tested without constructing
+/// the full [`DefaultRuntimeFileReloadHandler`].
+async fn reload_configuration_from_path(
+    config: &Arc<RwLock<crate::config::configfile::Configuration>>,
+    config_path: PathBuf,
+) -> Result<(), ReloadError> {
+    let content = std::fs::read_to_string(&config_path)
```
floroks: nitpick: should probably use `tokio::fs::read_to_string`

cda-main/src/update/security.rs
```diff
@@ -0,0 +1,419 @@
...
+                "No vehicle lock owned".to_owned(),
+            ))?;
+        if lock_state_provider.has_non_vehicle_locks().await {
+            return Err(RuntimeUpdateError::LockConflict(
+                "Non-vehicle locks are held, cannot apply update".to_owned(),
+            ));
+        }
+        if self.guard.has_active_operations() {
+            return Err(RuntimeUpdateError::OperationsInProgress(
+                "Another operation operation is running already (i.e. flash transfer)".to_owned(),
```
floroks: typo: "operation operation"

cda-plugin-runtime-update/src/storage.rs
```diff
@@ -0,0 +1,1747 @@
...
+///
+/// Returns [`RuntimeUpdateError`] if storage operations fail when reading collections.
+pub async fn compute_nextupdate_state(
+    storage: &impl Storage,
+    query: &RuntimeFilesQuery,
+) -> Result<BulkDataList, RuntimeUpdateError> {
+    let compute = async |collection_next: &CollectionName|
+           -> Result<Vec<BulkDataDescriptor>, RuntimeUpdateError> {
+        match storage.get_collection(collection_next).await {
+            Ok(_) => Ok(get_collection_items(storage, collection_next, query).await?),
```
floroks: question: wouldn't it make sense to use the connection handle from Ok() to list the items, instead of calling `get_collection` again inside of `get_collection_items`?

floroks:
general/question: Not sure how to handle it, but logically configuration files should be separate from the diagnostic database. Maybe adding a secondary bulk-data endpoint for them (e.g. `bulk-data/configuration-*`) could make sense? this could decouple the logic of "bulk-data" with preparation handling from the actual content of the files.

alexmohr:
> general/question: Not sure how to handle it, but logically configuration files should be separate from the diagnostic database. Maybe adding a secondary bulk-data endpoint for them (e.g. `bulk-data/configuration-*`) could make sense? this could decouple the logic of "bulk-data" with preparation handling from the actual content of the files.

What if a new mdd file requires a new configuration? When splitting the endpoints we cannot guarantee anymore that both are applied together

We could have a shared apply?

floroks:
> What if a new mdd file requires a new configuration? When splitting the endpoints we cannot guarantee anymore that both are applied together
> 
> We could have a shared apply?

That's a good point, it's probably very uncommon, but specifying a shared apply is probably more complex than it'd make sense, so we can keep it as is. We need to make sure to specify it in the documentation though.

cda-sovd/src/sovd/apps/sovd2uds/bulk_data/runtimefiles.rs
```diff
@@ -0,0 +1,504 @@
+/*
```
floroks, line 1: Shouldn't this implementation reside in the plugin, so that the paths are not available when the plugin is not added?

alexmohr:
> We need to make sure to specify it in the documentation though.

You mean in docs/03_architecture/05_plugins/02_diagnostic_database_update.rst?

cda-plugin-runtime-update/src/storage.rs
alexmohr: great hint, because this closure is bascially re-implementing `get_collection_item`

cda-main/src/update/security.rs
alexmohr: done

cda-main/src/update/mod.rs
alexmohr: done

cda-main/src/update/security.rs
alexmohr: done

cda-sovd/src/sovd/apps/sovd2uds/bulk_data/runtimefiles.rs
alexmohr: 100%, also the type should be defined in sovd interfaces instead of being build on the fly, fixed both.

cda-sovd/src/sovd/apps/sovd2uds/bulk_data/runtimefiles.rs
alexmohr: On purpose here, as this way all plugin can share the same sovd implementation and vehicle locking requirements. As the vehicle lock is specified by the requirement and is needed to make this safe, I think it's safe to assume that the sovd part is common enough to be shared among all plugin. 
It's not part of the 'normal' vehicle routes, you get by calling `cda_sovd::add_vehicle_routes` but a separate function `opensovd_cda_lib::update::add_runtime_update_routes` that only should be called if you want the plugin, if not it can be skipped. 
But I made the register function a bit more generic, so it can be used for all plugins and not only for the default plugin, of which the construction has been moved out into `opensovd_cda_lib::update::init_default_runtime_update_plugin`

floroks:
> > We need to make sure to specify it in the documentation though.
> 
> You mean in docs/03_architecture/05_plugins/02_diagnostic_database_update.rst?

yes, there should probably be a section on where the add for the http api is residing, but we could also move into a separate detailed design description later

alexmohr:
> yes, there should probably be a section on where the add for the http api is residing, but we could also move into a separate detailed design description later

done

cda-comm-uds/src/lib.rs
SizifWork: at  first glance  something like  that should  work :

```suggestion
    pub async fn shutdown(&self) {
    let mut tester_present_tasks = self.tester_present_tasks.write().await;
    let mut session_reset_tasks = self.session_reset_tasks.write().await;
    let mut security_reset_tasks = self.security_reset_tasks.write().await;
    let mut data_transfers = self.data_transfers.lock().await;

    for handle in tester_present_tasks
        .drain()
        .map(|(_, tp)| tp.task)
        .chain(session_reset_tasks.drain().map(|(_, h)| h))
        .chain(security_reset_tasks.drain().map(|(_, h)| h))
        .chain(data_transfers.drain().map(|(_, t)| t.task))
    {
        handle.abort();
    }
}

```

alexmohr: thank you, applied exactly as suggested.

cda-storage/src/local_storage.rs
```diff
@@ -92,10 +93,12 @@ impl LocalStorage {
 }
 
 impl Storage for LocalStorage {
+    type CollectionHandle = LocalCollection;
+
     async fn get_collection(
         &self,
         name: &CollectionName,
-    ) -> Result<Arc<impl Collection + 'static>, StorageError> {
+    ) -> Result<Arc<LocalCollection>, StorageError> {
```
theswiftfox, line 100: q: why did we put the concrete type here, but in 114 still using impl Bounds?

cda-sovd/src/sovd/execution_registry.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// update plugin are expected).
+    pub async fn replace(&self, other: &EcuExecutionRegistry) {
+        let replacement = other.inner.read().await.clone();
+        let mut inner = self.inner.write().await;
+        *inner = replacement;
+    }
+}
+
+impl cda_interfaces::runtime_update_api::ActiveOperationsGuard for EcuExecutionRegistry {
+    fn has_active_operations(&self) -> bool {
```
theswiftfox: question: the word `operations` made my mind go straight to the /operations endpoint logically. But if I understand this correctly we want to track if any `activity` is currently ongoing not only an operation right? Maybe we could choose a name that has less risk of being confused with the other?

edit: keeping this comment for completeness sake, but this is already the same discussion as within the update plugin code. (i just added it first here ^^)

cda-comm-uds/src/lib.rs
theswiftfox: if you want to go even shorter with full function chaining :D
```
        tester_present_tasks
            .drain()
            .map(|(_, tp)| tp.task)
            .chain(session_reset_tasks.drain().map(|(_, h)| h))
            .chain(security_reset_tasks.drain().map(|(_, h)| h))
            .chain(data_transfers.drain().map(|(_, t)| t.task))
            .for_each(|h| h.abort());
```

cda-interfaces/src/runtime_update_api/mod.rs
```diff
@@ -0,0 +1,436 @@
...
+            fn has_active_operations(&self) -> bool {
+                $(self.$idx.has_active_operations())||+
+            }
+        }
+    };
+}
+
+// Support up to 5 active operations guard. CDA internally currently is using up to two.
+// Defining more, so it can be used liberally when it's used as a library.
+// The alternative would be to use Vec<dyn ActiveOperationsGuard> instead of the type parameter
```
theswiftfox: opinion: I think we should either support exactly what we need in the CDA (eg. a single and a tuple of two guards) or we should allow full flexibility via using `Vec<Box<dyn ActiveOperationsGuard>>`
I'd prefer the second option. The Box<dyn ...> overhead should be neglible as this should not be a 'hot path' in the cda and allows for the most flexibility for the future.

cda-interfaces/src/runtime_update_api/mod.rs
```diff
@@ -0,0 +1,436 @@
...
+
+/// Provides read-only access to vehicle lock state for security validation.
+///
+/// Implemented by the SOVD server to expose lock information to plugins
+/// without creating a dependency on cda-sovd. OEMs may replace this
+/// implementation to integrate custom lock management systems.
+#[async_trait]
+pub trait LockStateProvider: Send + Sync + 'static {
+    /// Returns the `sub` claim of the vehicle lock owner, or `None` if no vehicle lock is held.
+    async fn is_vehicle_lock_owned(&self) -> Option<String>;
```
theswiftfox: nitpick: is_* in rust is usually reserved for functions returning a `bool`
I'd suggest using simply using `owned_vehicle_lock`

cda-interfaces/src/runtime_update_api/mod.rs
```diff
@@ -0,0 +1,436 @@
...
+    pub id: String,
+    pub mimetype: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub size: Option<u64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub hash: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub hash_algorithm: Option<HashAlgorithm>,
+    #[serde(
+        rename = "x-sovd2uds-OrigPath",
```
theswiftfox, line 218: question: OrigPath or OriginPath? OrigPath could also be read as 'original path'

cda-interfaces/src/storage_api/storage.rs
```diff
@@ -45,7 +48,9 @@ pub trait Storage: Send + Sync {
     fn get_or_create_collection(
         &self,
         name: &CollectionName,
-    ) -> impl Future<Output = Result<Arc<impl Collection + 'static>, StorageError>> + Send;
+    ) -> impl Future<
+        Output = Result<Arc<impl Collection + DirectFileAccess + 'static>, StorageError>,
```
theswiftfox: Q: in `get_collection` this was changed to Arc<Self::CollectionHandle>. Was it intentional to change it in one place but not the other? If we don't need the change to the concrete type having the impl allows for flexibility when changing the type around.

cda-main/src/config/mod.rs
```diff
@@ -42,3 +52,268 @@ pub fn load_config(config_file_path: Option<&str>) -> Result<configfile::Configu
...
+    Ok(())
+}
+
+/// Checks whether a configuration source is available.
+///
+/// # Errors
+/// Returns [`AppError::ConfigurationError`](crate::AppError::ConfigurationError) when no
+/// configuration file is found on disk or in storage.
+#[cfg(not(feature = "config-optional"))]
+pub fn require_config_source() -> Result<(), crate::AppError> {
```
theswiftfox: suggestion: to avoid the clippy thingy altogether you could use a `if cfg!(...)` pattern:

like this, clippy should not complain and you only have one function definition
```rust
/// Checks whether a configuration source is available.
///
/// # Errors
/// Returns [`AppError`](crate::AppError) when no configuration source is found
/// (only when the `config-optional` feature is disabled).
pub fn require_config_source() -> Result<(), crate::AppError> {
    if cfg!(feature = "config-optional") {
        println!("No configuration found on disk or in storage. Using default values.");
        Ok(())
    } else {
        Err(crate::AppError::ConfigurationError(
            "No configuration found. Provide a configuration file, store one via runtime update, \
             or build with the 'config-optional' feature to allow starting without one."
                .to_owned(),
        ))
    }
}
```

cda-main/src/update/mod.rs
```diff
@@ -0,0 +1,427 @@
...
+///
+/// Creates the reload handler, storage backend, and returns a configured
+/// [`cda_plugin_runtime_update::DefaultRuntimeFilesUpdatePlugin`] instance.
+/// The returned plugin is not wrapped in Arc; the caller should apply Arc wrapping
+/// if needed.
+///
+/// # Errors
+/// Returns [`AppError`] if storage initialization fails.
+pub async fn init_default_runtime_update_plugin<S, F, P>(
+    ctx: Box<
```
theswiftfox, line 257: q: was wondering about the boxing here - this is due to async and future size? If yes, can we add a comment why the box was used here so we remember in the future? :D
if no, can we drop the Box? ^^

cda-main/src/lib.rs
```diff
@@ -357,35 +311,12 @@ pub fn generate_config_cmd(output: Option<&PathBuf>) -> Result<(), AppError> {
...
-    })
-}
-
 /// Parse CLI arguments and start the CDA with the default startup flow.
 ///
 /// # Errors
 /// Returns [`AppError`] if configuration loading, validation, or startup fails.
 pub async fn run_from_cli() -> Result<(), AppError> {
-    run(AppArgs::parse()).await
+    Box::pin(run(AppArgs::parse())).await
```
theswiftfox, line 319: question: why the Box::pin here?

cda-main/src/lib.rs
```diff
@@ -557,641 +585,78 @@ pub async fn load_vehicle_data<
...
             tracing::warn!(error = %e, "Failed to register database health provider");
         }
-        Some(provider)
+        Some(HealthProviders { doip, database })
     } else {
         None
-    }
-}
-
-fn warn_unmatched_ecu_config_keys<S: SecurityPlugin>(
```
theswiftfox, line 805: q: did i miss it somewhere else, or did we loose this check?

cda-main/src/mdd.rs
theswiftfox, line 1: question: (maybe a dumb one) would this make sense to have this module exported via the cda-database crate instead of cda-main?

cda-storage/src/storage_seed.rs
theswiftfox, line 1: q: (similar to mdd.rs) does it make sense to have this as part of the storage_api crate?

cda-plugin-runtime-update/src/operations/mod.rs
theswiftfox, line 1: stupid nitpick from me / discussion point: i keep stumbling upon the word operations 🙈 would it make sense to use the terminology of 'actions' in the context of the update plugin? eg. actions::apply etc? I know that operations is the technically accurate term here though

cda-plugin-runtime-update/src/operations/executions.rs
```diff
@@ -0,0 +1,371 @@
...
+                return Err(e);
+            }
+            Ok(false) => {}
+        }
+    }
+
+    let execution_id = uuid::Uuid::new_v4().to_string();
+    {
+        let mut execs = params.executions.write().await;
+        execs.insert(
```
theswiftfox, line 116: question: where are we removing the exections that are completed / aborted? Or are we keeping them indefinitely for now?

theswiftfox (commented):
sorry for the many questions 🙈

cda-storage/src/local_storage.rs
alexmohr: using the `CollectionHandle` type now on all trait functions. It's a bit more restrictive than the previous approach though.

cda-comm-uds/src/lib.rs
alexmohr: done :)

cda-interfaces/src/runtime_update_api/mod.rs
alexmohr: makes sense, replaced with `Vec<Box<dyn...`

cda-interfaces/src/runtime_update_api/mod.rs
alexmohr: renamed to `vehicle_lock_owner_sub` to make it even clearer what the string contains :)

cda-interfaces/src/runtime_update_api/mod.rs
alexmohr: Origin path, but only moved that :/ See [main](https://github.com/eclipse-opensovd/classic-diagnostic-adapter/blob/main/cda-sovd-interfaces/src/lib.rs#L105) I am not sure if anyone downstream depends on that naming already. 

I do agree though that 'OriginPath' would be better.

cda-interfaces/src/storage_api/storage.rs
alexmohr: Well we only _need_ it in one place. I changed this, so we use the type in all places. Alternatively we could drop the type from the trait.

cda-main/src/config/mod.rs
alexmohr: Thank you. :)

cda-main/src/update/mod.rs
alexmohr: There is a comment in 284, but added 
```
    // Boxed to keep the async future size within Clippy's `large_futures` limit;
    // the context is heap-allocated so its fields don't bloat the state machine.
```
Before the `ctx` parameter.

cda-main/src/lib.rs
alexmohr: left over from refactoring or hallucination. It's not needed and removed.

cda-main/src/lib.rs
alexmohr: We did not lose it. I renamed it to `handle_ecu_config_keys` because I found it confusing that a 'warn' function can return an error. Feel free to suggest a better name though :)

cda-main/src/mdd.rs
alexmohr: Not really, the main issue with that is that it needs `Configuration`, `EcuConfig` `FileManagerMap` from cda-main. We could refactor all of these things into cda-database though, replace the `AppError` currently used with a different Error type and implement mapping from the new type to AppError.

`FileManagerMap` is just a new type pattern so it could move to database too.
The `seed_storage_collection.rs` then also must be moved so it can be accessed from the `database` crate.

So while it _could_ be moved I'd rather do that as a follow up if at all.

cda-storage/src/storage_seed.rs
alexmohr: In constrast to mdd.rs this can be moved without issue. Done :)

cda-plugin-runtime-update/src/operations/executions.rs
alexmohr: good catch, they where stored indefinitely. I changed this by only keeping the latest execution. As we cannot start a new execution while one is in progress.

I changed this, so only the latest execution is kept. I still use a map to store this though.
But while doing this I noticed that I forgot to implement `GET /apps/sovd2uds/bulk-data/runtimefiles-nextupdate/executions/` as well. 
This is now fixed and mirrors the other endpoints. 

The specification is updated to document that only the latest execution state is kept.

Also the endpoint for `GET /apps/sovd2uds/bulk-data/runtimefiles-nextupdate/executions/id` was missing include-schema query parameter which now is added as well. 

Great hint here :)

cda-plugin-runtime-update/src/operations/mod.rs
alexmohr: Well, as the endpoint so also available via `/apps/sovd2uds/operations/diagnostic-database-update` it kinda makes sense to name this operations. `executions` would be an alternative if we stick to the sovd naming. The enum is called ExuectionMode::Apply right now, we could rename all that to actions but imho keeping Executions is cleaner. We could replace 'operations' with 'executions' though if that helps.

cda-sovd/src/sovd/execution_registry.rs
alexmohr: Commented on the other comment as well, but I can agree with this one being confusing. Per your suggestion named this to `ActivityGuard`

alexmohr (commented):
thanks for your review, I hope I addressed all findings. The last discussion point still standing is the 'operations' rename.

cda-main/src/lib.rs
theswiftfox: suggestion: add a comment why the pin is needed? (large future, as clippy complained :D) I'm gonna forget and remove it again in the future otherwise or ask stupid questions again xD

alexmohr: I forgot myself that clippy already complained while I was implementing this. Added the comment :)

theswiftfox (approved):

Issue #54, opened 2025-09-18
schema_for_responses is not returning the MUX Selector


alexmohr:
For example 0x1904 returns data structured like this
```json
      "StandardEnvironmentData": {
        "Case_0x1": {
          "External_Tester_Present_Flag": "not present",
          "Frequency_Counter": 32,
          "Occurrence_Flag": "Occurrence",
          "Operation_Cycle_Counter": 96
        },
        "Selector": 1
      }
```
Generating the schema for this response returns 
```json
 "any-of": Array [
            Object {
                "properties": Object {
                    "External_Tester_Present_Flag": Object {
                        "default": String(""),
                        "type": String("string"),
                    },
                    "Frequency_Counter": Object {
                        "default": String(""),
                        "type": String("integer"),
                    },
                    "Occurrence_Flag": Object {
                        "default": String(""),
                        "type": String("string"),
                    },
                    "Operation_Cycle_Counter": Object {
                        "default": String(""),
                        "type": String("integer"),
                    },
                },
                "type": String("object"),
            },
        ],
    },
```

While this is correct from from the database perspective, in [ecumanager.rs#L1826](https://github.com/eclipse-opensovd/classic-diagnostic-adapter/blob/main/cda-core/src/diag_kernel/ecumanager.rs#L1826) a 'Selector' key is added. This should be reflected by the schema.

PR #371, opened 2026-06-08
refractor (cda-comm-uds): split lib into modules

theswiftfox:
## Summary

This is the first of a set of commits / PRs to split up the lib.rs into smaller testable (and tested) submodules.

## Checklist

## Notes for Reviewers

---

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -155,6 +155,7 @@ arithmetic_side_effects = "deny"
 separated_literal_suffix = "deny"
 disallowed_methods = "deny"
 string_slice = "deny"
+needless_pass_by_ref_mut = "deny"
 
 [workspace.lints.rust]
 unexpected_cfgs = { level = "warn", check-cfg = ['cfg(nightly)'] }
--- a/cda-comm-uds/src/lib.rs
+++ b/cda-comm-uds/src/lib.rs
@@ -1,6 +1,6 @@
 /*
  * SPDX-License-Identifier: Apache-2.0
- * SPDX-FileCopyrightText: 2025 The Contributors to Eclipse OpenSOVD (see CONTRIBUTORS)
+ * SPDX-FileCopyrightText: 2026 The Contributors to Eclipse OpenSOVD (see CONTRIBUTORS)
  *
  * See the NOTICE file(s) distributed with this work for additional
  * information regarding copyright ownership.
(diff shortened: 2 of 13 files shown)


cda-comm-uds/src/lib.rs
```diff
@@ -886,7 +368,7 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsMana
             0,
             &dtc_code.to_be_bytes(),
         )?;
-        raw_payload.push(0xFF); // record number, 0xFF means all records or all memory
+        raw_payload.push(0xFF);
```
alexmohr: suggestion: can we please keep the comment here. Imho it's useful to know why 0xFF is used. Alternatively I would be happy with a named constant instead of a magic number.

alexmohr: Similar in 929 and 1023 comments are lost that where useful when implementing #54 
In general you removed a bunch of comments that where useful in the next couple of diff lines. please bring them back.

cda-comm-uds/src/security.rs
```diff
@@ -0,0 +1,142 @@
...
+
+                    Ok((security_access, result))
+                }
+                DiagServiceResponseType::Negative => Ok((security_access, result)),
+            }
+        }
+    }
+}
+
+pub(crate) async fn do_get_send_key_param_name<
```
alexmohr: general question: While refactoring the diag_kernel (part 2) I was at the same point of adding these `do_` functions (just named differently. 
But I found this a bit cumbersome to read through so I opted to split the `EcuManager` ttrait into subtraits which the modules implement and EcuManager is just a composition of these traits. 
I.e. https://github.com/eclipse-opensovd/classic-diagnostic-adapter/pull/378/changes#diff-3baf67e6d009bdd2c9c862d365a9e0539ba9058dbc480cb0542107462f373bf1R587 

Can we align the approaches, so we have one way of doing things?

Cargo.toml
```diff
@@ -152,6 +152,7 @@ arithmetic_side_effects = "deny"
 separated_literal_suffix = "deny"
 disallowed_methods = "deny"
 string_slice = "deny"
+needless_pass_by_ref_mut = "deny"
```
alexmohr, line 158: nitpick: Refacatoring PRs are always hard to review, even though the changes are in separate commits, is it possible to split this into to PRs instead? (and also require this lint in cicd)

theswiftfox: thats why it is in a seperate commit. i can open a PR for the 4 places where it was changed, if you prefer. but the overhead is a bit much for the tiny change.
as for why I added it, its simply i noticed that functions have a mut ref that was not needed and i noticed the lint not being enabled. If we want to add it in CI/CD we would in theory need to align again with the project, as this was not part of the original set of lints that were agreed upon in the adr afaik.

cda-comm-uds/src/security.rs
theswiftfox: sure, we can go with that approach as well. I don't have a preference here :)

cda-comm-uds/src/lib.rs
theswiftfox: for the first one, I think a named const would be the way to go if we want to keep the info.
For the ones related to 54, I usually would rather have code references in the issue than todo comments that get never removed in the codebase that reference an issue. But I can add them back as well if you prefer to have them here. (and this should be a structural refractor, not a cleanup ^^)

theswiftfox: but thanks for the hint, some comments got lost when moving stuff around. will restore them

Cargo.toml
alexmohr: alright fine 😆 , can you take care of aligning with the project? maybe we want to collect even more 'deny' lints?

cda-comm-uds/src/security.rs
alexmohr: I prefer splitting it up instead of `do_` @mbfm any opinion here?

Cargo.toml
theswiftfox: I'll put it on the agenda for next weeks architecture board :)

cda-comm-uds/src/security.rs
theswiftfox: the more i think about it the more i lean on prefering the trait split

alexmohr: then I would argue for trait splitting 😆

mbfm: Hmm, is the `UdsManager` in the public API that users/OEMs will need to interact with?

Users will need to have the trait imported to be able to call the methods defined in that trait. So, we would need to expose those traits in the public API then.

cda-comm-uds/src/lib.rs
```diff
@@ -1815,8 +997,6 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsEcu
             ));
         }
 
-        // even if the data transfer job is done,
-        // data_transfer_exit must be called before starting a new one
```
alexmohr, line 1865: question: removed the comments here on purpose?
Some of them where useful imho (not all though)

This one here is kinda useful. 

It would be great if we could get the comments back that explain the variant detection. Like:

> No duplicated ECUs, proceed with normal variant detection

> Detect variants for all duplicated ECUs

> // No specific variant found despite online ECUs - mark all as undetected.
                // Falling back to base variant is only allowed when there are no duplicates.


 // Nothing to do --> this one can be removed though. But I don't think removing all of them is a good idea

cda-comm-uds/src/lib.rs
```diff
@@ -2577,16 +1639,16 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsEcu
         for (name, ecu) in self.ecus.iter() {
             let ecu_guard = ecu.read().await;
             if gateway_only && ecu_guard.logical_address() != ecu_guard.logical_gateway_address() {
-                continue; // skip non gateway ECUs
+                continue;
```
alexmohr: question: Nagging again about comments, but I think they where quite useful and I would like to have them back at the continue statements.

Can you please go through the code again and check if you really ment to remove all these comments. It's fine to remove useless comments like ' // skip functional description ecu for this' as that is quite obvious but not all of it is that obvious.

alexmohr: This is the last 'you lost comments' for this review round, but please double check the rest too.

cda-comm-uds/src/session.rs
```diff
@@ -0,0 +1,166 @@
...
+        ecu_name: &str,
+        expiration: Option<Duration>,
+        reset_type: ResetType,
+    ) {
+        let expiration = if let Some(expiration) = expiration
+            && expiration > Duration::ZERO
+        {
+            expiration
+        } else {
+            return;
```
alexmohr, line 38: question: Should we log a warning here?

cda-comm-uds/src/tester_present.rs
```diff
@@ -0,0 +1,255 @@
...
+        control_msg: &TesterPresentControlMessage,
+    ) -> Result<(), DiagServiceError> {
+        let payload = {
+            let ecu = self.ecu_manager(&control_msg.ecu)?;
+            let target_address = match &control_msg.type_ {
+                TesterPresentType::Functional(_) => ecu.read().await.logical_functional_address(),
+                TesterPresentType::Ecu(_) => ecu.read().await.logical_address(),
+            };
+            ServicePayload {
+                data: vec![service_ids::TESTER_PRESENT, 0x80],
```
alexmohr: nitpick: magic number 0x80

cda-comm-uds/src/lib.rs
theswiftfox: terribly sorry for that. I think i should have restored all relevant comments now (removed a few where it got replaced with a named constant)

cda-comm-uds/src/session.rs
theswiftfox: I don't think so. From my understanding this is here so that a `set_ecu_security_access` call without expiration (None) does not trigger a reset task. We could instead handle this before calling the reset task logic, but iirc we explicitly decided against that.

cda-comm-uds/src/lib.rs
```diff
@@ -1017,10 +447,10 @@ impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsMana
         scope: DtcReadInformationFunction,
     ) -> Result<(Option<ExtendedSnapshots>, Option<serde_json::Value>), DiagServiceError> {
         fn extract_schema_properties(schema_desc: &SchemaDescription) -> Option<serde_json::Value> {
+            // Todo when solving #54: We are missing the mux case name in the schema.
```
theswiftfox, line 497: note: i thought the comment was fitting on the start of the function. if you disagree i'll move it back

cda-comm-uds/src/tester_present.rs
theswiftfox: fixed by creating a constant for it

cda-comm-uds/src/lib.rs
theswiftfox: i think i finally got em all. 🤯 sorry for the comment chaos.

cda-comm-uds/src/lib.rs
theswiftfox: this one was replaced with a constant that has a docstring. the other comments that were lost should all be restored.

mbfm:
By the way, I noticed that the `UdsManager::_phantom` field isn't necessary anymore. Might be worth removing as part of this pull request.

These would be the changes necessary:
```diff
cda-comm-uds/src/lib.rs --- 1/3 --- Rust
 62     security_reset_tasks: Arc<RwLock<HashMap<EcuIdentifier, JoinHandle<()>>>>,
 63     functional_description_database: String,
 64     fault_config: FaultConfig,
-65     _phantom: std::marker::PhantomData<R>,
 66 }
 67
 68 impl<S: EcuGateway, R: DiagServiceResponse, T: EcuManager<Response = R>> UdsManager<S, R, T> {

cda-comm-uds/src/lib.rs --- 2/3 --- Rust
 85                 .description_database
 86                 .clone(),
 87             fault_config,
-88             _phantom: std::marker::PhantomData,
 89         };
 90
 91         let vd_uds_clone = manager.clone();

cda-comm-uds/src/lib.rs --- 3/3 --- Rust
 635             security_reset_tasks: Arc::clone(&self.security_reset_tasks),
 636             functional_description_database: self.functional_description_database.clone(),
 637             fault_config: self.fault_config.clone(),
-638             _phantom: self._phantom,
 639         }
 640     }
 641 }

```

cda-comm-uds/src/lib.rs
alexmohr: Looks like most comments are back, as a side effect spelling and upper case first was fixed in some cases 😆

cda-comm-uds/src/lib.rs
alexmohr: I don't see something anymore that is obviously lost. :) thank you

alexmohr:
PR looks good to me, approving once the std::marker is also removed :)

alexmohr (approved):

PR #373, opened 2026-06-09
chore: refactor sovd manager into modules.

alexmohr:
## Summary
This is the first commit in the diag_kernel refactoring, to split the module up.
It extracts test_utils module which is shared among the different modules to create ecumanager instances and mock mdd datatypes. Service lookups and comparams are also refactored into a new module.

## Checklist

## Notes for Reviewers


cda-core/src/diag_kernel/service_lookup.rs
```diff
@@ -0,0 +1,927 @@
...
+                .write()
+                .await
+                .insert(lookup_id, Some(location));
+            Some(service)
+        } else {
+            None
+        }
+    }
+
+    pub(crate) fn search_with_location<F>(
```
theswiftfox, line 220: q: is this pub(crate) as we expect this to be called from other modules in the future? Right now it seems to only be privately called.
(same goes for get_services_from_diag_layer_and_parent_refs)

theswiftfox (commented):
LGTM. Just one question about visibility of some methods.

cda-core/src/diag_kernel/service_lookup.rs
alexmohr: The idea was, to make the various lookup pub(crate), so they can be used later on in other modules. I could make them private though an we make them pub(crate) when (or rather if) we need it.

alexmohr:
@theswiftfox Question answered, PR rebased, pipeline happy. I would appreciate another review because this is a pain to rebase 🙈

cda-core/src/diag_kernel/service_lookup.rs
theswiftfox: we can keep it with the pub(crate) in that case as it will probably be used once the final refractor pass is done, but ideally we only export the ones used outside of the module, as it makes the glance value a bit better imo if i can directly know the fn is not used outside of the module :)

theswiftfox (approved):

PR #380, opened 2026-06-15
docs(error-handling): add requirements and architecture documentation

floroks:
## Checklist

### DIFF
--- a/.opencode/agents/AGENTS.md
+++ b/.opencode/agents/AGENTS.md
@@ -1,0 +1,226 @@
+---
+name: cda
+description: CDA development agent with project context, code style rules, and ASPICE documentation guidance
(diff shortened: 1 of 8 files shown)


.opencode/agents/AGENTS.md
```diff
@@ -0,0 +1,226 @@
+---
```
alexmohr, line 1: suggestion: While this `AGENTS.md` isn't huge (yet?), the bigger an agents file is the faster it will be ignored eventually by an LLM. What worked really well for me personally is to turn the agents.md into an index and use it to link to different skills instead of trying to store everything in a single agents file. 
Example: https://github.com/alexmohr/assimilate/blob/main/AGENTS.md

PR #390, opened 2026-06-18
feat(tests): add Tester Present integration tests

alexmohr:
### feat(tests): add Tester Present integration tests
Add ECU simulator disconnect/hard reset capabilities to simulate TCP connection drops and ECU reboots. Add comprehensive tests verifying Tester Present frames are maintained during active locks across session switches and network disrupts.

### fix(integration-tests): fsync testcontainer CDA configuraiton
When running the integration tests in a loop, to test for flaky tests, the tests break from time to time because they are unable to parse the CDA configuration. The parser points out the file is only partially written. This commit syncs the disk state for the test container directory (making sure meta data is up to date) and the configuration file itself to make sure this is fully flushed before usage too.

### fix(variant-detection): trigger variant detection after reconnection
To make sure an ECU is correctly detected as online after a network disconnect, this commit triggers a variant detection as soon as the connection is re-established. Aside from setting the correct ECU state, this makes also sure that the tester present task does not skip the ECU anymore, marking it as offline.

### fix(update-reload): fix an issue that ECUs do not come back online
The shutdown was not canceling several tasks which left them running with outdated references. This commit collects all tasks that must be shut down and does so cleanly during application shutdown or when applying an update. Update is now destroying all old instances first, then setting up new instances instead doing this one at a time. Fix an issue where the DOIP socket was not closed properly, leaving messages being sent to the wrong instance by re-using the existing socket instead of re-creating it.

### refactor(connection): add two structs containing connection setup
Solves clippy warning of too many parameters and makes usage a bit more ergonomic.

## Checklist

## Notes for Reviewers

Deeming the crap regression as acceptable because create_socket(_with_protocol) is actually tested through the integration tests and just not measured yet. 

### DIFF
--- a/cda-comm-doip/src/connections.rs
+++ b/cda-comm-doip/src/connections.rs
@@ -64,6 +64,28 @@ struct ConnectionSettings {
     alive_check_interval: Duration,
 }
 
+#[derive(Clone)]
+struct GatewayTarget {
+    ip: String,
+    name: String,
+    ecus: Vec<u16>,
+}
+
+#[derive(Clone)]
+struct GatewayHandlerConfig {
+    connection_config: ConnectionConfig,
+    doip_connection_config: DoIPConfig,
+    routing_activation_request: RoutingActivationRequest,
+    connection_settings: ConnectionSettings,
+    variant_detection: Option<(mpsc::Sender<Vec<String>>, Vec<String>)>,
+}
+
+struct GatewayConnectionHandles {
+    sender: mpsc::Sender<DoipPayload>,
+    receivers: HashMap<u16, broadcast::Receiver<Result<DiagnosticResponse, EcuError>>>,
+    task_handles: Vec<tokio::task::JoinHandle<()>>,
+}
+
 #[derive(Error, Debug, Clone)]
 pub enum EcuError {
     #[error("Resource not found: `{0}`")]
@@ -164,14 +186,16 @@ where
     let connection_timeout = gateway_ecu.connection_timeout();
     let connection_retry_attempts = gateway_ecu.connection_retry_attempts();
 
-    let (sender, receiver, task_handles) = match connection_handler(
-        config.connection.clone(),
-        gateway.ip.clone(),
-        gateway.ecu.clone(),
-        config.doip,
+    let target = GatewayTarget {
+        ip: gateway.ip.clone(),
+        name: gateway.ecu.clone(),
+        ecus: ecu_ids.clone(),
+    };
+    let config = GatewayHandlerConfig {
+        connection_config: connection_config.clone(),
+        doip_connection_config,
         routing_activation_request,
-        ecu_ids.clone(),
-        ConnectionSettings {
+        connection_settings: ConnectionSettings {
             routing_activation: routing_activation_timeout,
             retry_delay: connection_retry_delay,
             connect_timeout: connection_timeout,
@@ -180,10 +204,13 @@ where
             alive_check_interval: config.alive_check_interval,
         },
         variant_detection,
-    )
-    .await
-    {
-        Ok((sender, receiver, task_handles)) => (sender, receiver, task_handles),
+    };
+    let GatewayConnectionHandles {
+        sender,
+        receivers,
+        task_handles,
+    } = match connection_handler(target, config).await {
+        Ok(handles) => handles,
         Err(e) => {
             return Err(EcuError::EcuConnectionError(
                 ConnectionError::ConnectionFailed(format!(
@@ -194,7 +221,7 @@ where
         }
     };
 
-    let doip_ecus = create_ecu_receiver_map(ecu_ids, &sender, &receiver);
+    let doip_ecus = create_ecu_receiver_map(ecu_ids, &sender, &receivers);
 
     tracing::info!("Connected to gateway");
     state
@@ -243,36 +270,22 @@ fn create_ecu_receiver_map(
     doip_ecus
 }
 
-#[allow(clippy::type_complexity)]
 #[tracing::instrument(
-    skip(routing_activation_request, connection_settings, connection_config),
+    skip(target, config),
     fields(
-        tester_ip = connection_config.source_ip.clone(),
-        port = connection_config.port,
-        tls_port = connection_config.tls_port,
-        gateway_ip = %gateway_ip,
-        gateway_name = %gateway_name,
-        ecu_count = ecus.len(),
+        tester_ip = config.connection_config.source_ip.clone(),
+        port = config.connection_config.port,
+        tls_port = config.connection_config.tls_port,
+        gateway_ip = %target.ip,
+        gateway_name = %target.name,
+        ecu_count = target.ecus.len(),
         dlt_context = dlt_ctx!("DOIP"),
     )
 )]
 async fn connection_handler(
-    connection_config: ConnectionConfig,
-    gateway_ip: String,
-    gateway_name: String,
-    doip_connection_config: DoIPConfig,
-    routing_activation_request: RoutingActivationRequest,
-    ecus: Vec<u16>,
-    connection_settings: ConnectionSettings,
-    variant_detection: Option<(mpsc::Sender<Vec<String>>, Vec<String>)>,
-) -> Result<
-    (
-        mpsc::Sender<DoipPayload>,
-        HashMap<u16, broadcast::Receiver<Result<DiagnosticResponse, EcuError>>>,
-        Vec<tokio::task::JoinHandle<()>>,
-    ),
-    EcuError,
-> {
+    target: GatewayTarget,
+    config: GatewayHandlerConfig,
+) -> Result<GatewayConnectionHandles, EcuError> {
     // channel to send messages to the gateway / ecus
     let (intx, inrx) = mpsc::channel::<DoipPayload>(50);
 
@@ -285,7 +298,7 @@ async fn connection_handler(
         HashMap::new();
 
     // create ecu response channels
-    for ecu in ecus {
+    for &ecu in &target.ecus {
         let (tx, rx) = broadcast::channel::<Result<DiagnosticResponse, EcuError>>(10);
 
         outtx.insert(ecu, tx);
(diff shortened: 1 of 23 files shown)


cda-comm-doip/src/lib.rs
```diff
@@ -71,19 +73,91 @@ enum DiagnosticResponse {
...
+    pub(crate) ecus: Arc<HashMap<String, RwLock<T>>>,
+    pub(crate) socket: Arc<Mutex<DoIPUdpSocket>>,
+}
+
 pub struct DoipDiagGateway<T: EcuAddresses + DoipComParams> {
-    doip_connections: Arc<RwLock<Vec<Arc<DoipConnection>>>>,
-    logical_address_to_connection: Arc<RwLock<HashMap<u16, usize>>>,
-    ecus: Arc<HashMap<String, RwLock<T>>>,
-    socket: Arc<Mutex<DoIPUdpSocket>>,
+    state: Arc<DoipGatewayState<T>>,
```
theswiftfox: q: why is this an Arc<..> as it has internally only Arcs, we can simply make it 'Clone' and then use it without the Arc.

cda-comm-doip/src/connections.rs
```diff
@@ -425,147 +402,163 @@ fn spawn_connection_reset_task(
...
+            let alive_check_interval = gateway_connection
+                .gateway_doip_config
+                .transport
+                .alive_check_interval;
+            let send_timeout = gateway_connection
+                .gateway_doip_config
+                .transport
+                .send_timeout;
+            let alive_check_enabled = alive_check_interval > Duration::ZERO;
+            let effective_interval = if alive_check_enabled {
```
theswiftfox: suggestion:  we don't need to set a different interval if alive check is disabled (see [here](https://docs.rs/tokio/latest/tokio/macro.select.html))

Specifically:
Additionally, each branch may include an optional if precondition. If the precondition returns false, then the branch is disabled. The provided <async expression> is still evaluated but the resulting future is **never polled**. This capability is useful when using select! within a loop. (emphasis mine)

testcontainer/ecu-sim/src/main/kotlin/webserver/WebserverRoutes.kt
```diff
@@ -195,6 +195,46 @@ data class SimpleInterceptorDto(
...
+        var closedConnections = 0
+
+        // Verify the ECU exists
+        findByEcuName(ecuName)
+            ?: return@post call.respond(
+                HttpStatusCode.NotFound,
+                mapOf("message" to "ECU $ecuName not found"),
+            )
+
+        // Close all TCP connections on all DoIP entities in all network instances.
```
theswiftfox: q: in the description its stated that all TCP connections to **this** ECU are dropped. Here we close all connections on all DoIP entities? The comment seems that this is intentional, but then I'd suggest dropping the {ecu} path param and make it obvious that this kills all DoIP traffic from the endpoint already :)

cda-comm-doip/src/lib.rs
```diff
@@ -267,48 +344,68 @@ impl<T: EcuAddresses + DoipComParams> DoipDiagGateway<T> {
...
+            vam_listener_handle: Arc::new(vam_listener_handle),
+        })
     }
 
-    pub fn shutdown(&self) {
+    pub async fn shutdown(&mut self) {
         self.cancel_token.cancel();
+        // Abort and await the VAM listener task so it stops reading from the
+        // shared UDP socket before a new gateway reuses it.
+        self.vam_listener_handle.abort();
```
theswiftfox: suggestion: similar to the task_handles, we need to await after aborting.
for example like this, to ensure the task is not stuck doing something still.
```rust
    self.vam_listener_handle.abort();
    let _ = (&mut *self.vam_listener_handle).await;
```

cda-comm-doip/src/lib.rs
```diff
@@ -267,48 +344,68 @@ impl<T: EcuAddresses + DoipComParams> DoipDiagGateway<T> {
...
-        Ok(gateway)
+        Ok(DoipDiagGateway {
+            state,
+            cancel_token,
+            vam_listener_handle: Arc::new(vam_listener_handle),
+        })
     }
 
-    pub fn shutdown(&self) {
+    pub async fn shutdown(&mut self) {
```
theswiftfox, line 392: q (commenting this here as this is where i noticed):
while the tasks are `.abort()`ed here, they are never awaited. This can cause issues, if a consumer is expecting all tasks to not do anything after shutdown is being called. If it is intentional, we should document it.

I think the easiest solution would be to collect them on the DoipGatewayState into a `tokio::task::JoinSet` (if we don't need to access them individually for some reason?)

Then the shutdown could to this:
```rust
    let mut tasks = self.state.connection_tasks.lock().await;
    tasks.abort_all();
    while tasks.join_next().await.is_some() {}
    drop(tasks);
```
The other option would be to individually await each task in the existing structure I think.

cda-main/src/update/mod.rs
```diff
@@ -108,31 +108,34 @@ where
     async fn reload_databases(&self, mdd_paths: Vec<PathBuf>) -> Result<(), ReloadError> {
         let cfg = self.config.read().await.clone();
 
+        // Shut down old components BEFORE creating new ones to avoid port/socket
+        // conflicts. Reuse the existing UDP socket so that there is never a second
+        // socket bound to the same DoIP port (which would cause non-deterministic
+        // VAM delivery between old and new listeners).
+        self.variant_detection_handle.lock().await.abort();
```
theswiftfox: suggestion: as before, abort() only tells the task to cancel, we cannot be sure it is actually stopped after calling abort().

theswiftfox (commented):
Mostly minor nitpicks and comments about handling of `abort()`. While I don't think it would cause issues on  a regular basis, I fear we could run into some edgecases with bad timing on the aborts ^^

Sidenote, the docs on `abort()` for tokio are kind of bad. They only note [here](https://docs.rs/tokio/latest/tokio/task/index.html#cancellation) how canellation actually behaves..

cda-comm-doip/src/lib.rs
alexmohr: No hard reason here, Arc<...> didn't require implementing the 7 lines for  `impl<T: EcuAddresses + DoipComParams> Clone for DoipGatewayState<T>` but changed this and using clone now

cda-comm-doip/src/connections.rs
alexmohr: I'm not sure if I can follow here: 
```rust
    _ = alive_interval.tick(), if alive_check_enabled => {
```

Already contains a precondition, so the future is only polled when the alive tick is enabled. 
The separate value for the interval is used to prevent a panic, because Duration::Zero is not supported as an interval for select. 

We could make the interval an option though like below. I'm not sure if that would fulfill your suggestion. The suggestion below is probably a bit nicer.

```diff
diff --git a/cda-comm-doip/src/connections.rs b/cda-comm-doip/src/connections.rs
index 93493857..bc797dfd 100644
--- a/cda-comm-doip/src/connections.rs
+++ b/cda-comm-doip/src/connections.rs
@@ -444,21 +444,14 @@ where
                 .transport
                 .send_timeout;
             let alive_check_enabled = alive_check_interval > Duration::ZERO;
-            let effective_interval = if alive_check_enabled {
-                alive_check_interval
-            } else {
-                // Use a very large duration when disabled so the interval never fires.
-                // tokio::time::Instant is limited so we use ~136 years which is well
-                // within the representable range.
-                Duration::from_secs(u64::from(u32::MAX))
-            };
-            let mut alive_interval = tokio::time::interval_at(
-                tokio::time::Instant::now()
-                    .checked_add(effective_interval)
-                    .expect("interval start overflow"),
-                effective_interval,
-            );
-            alive_interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+            let mut alive_interval = alive_check_enabled.then(|| {
+                let mut interval = tokio::time::interval_at(
+                    tokio::time::Instant::now() + alive_check_interval,
+                    alive_check_interval,
+                );
+                interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+                interval
+            });
             loop {
                 tokio::select! {
                     // Per default tokio, randomized which branch is checked first to ensure fairness.
@@ -512,10 +505,15 @@ where
                         }
                         // Reset the alive check timer so it only fires after a full
                         // interval of silence - never during active communication.
-                        alive_interval.reset();
+                       alive_interval.as_mut().map(|interval| interval.reset());
                     },
 
-                    _ = alive_interval.tick(), if alive_check_enabled => {
+                   _ = async {
+                        match alive_interval.as_mut() {
+                                Some(interval) => interval.tick().await,
+                                None => std::future::pending().await,
+                            }
+                        }, if alive_check_enabled => {
                         if send_pending_status(&send_pending_tx, true).is_err() {
                             break;
                         }
```

testcontainer/ecu-sim/src/main/kotlin/webserver/WebserverRoutes.kt
alexmohr: Followed your suggestion. Even if this would only close the connection of a single ecu, it would close the gateway connection which might be shared between different ECUs which will be disconnected too, making this confusing at best. Closing all connections is clearly the better API design here. thank you.

cda-comm-doip/src/lib.rs
alexmohr: Thank you. Implemented excatly as suggested.

cda-comm-doip/src/lib.rs
alexmohr: Implementing it _exactly_ like this isn't possible because we have to take ownership of the future to await it. Changed the struct field to an option, taking ownership and then using `let...` as suggested to wait until the future is done. applies to the other similar comment too.

cda-comm-doip/src/connections.rs
theswiftfox: Oooh I see. Yeah this is perfect. thank you

alexmohr: weeelllll I didn't check the suggestion for clipyp warnings first.
So I had to keep this
```
tokio::time::Instant::now()
                        .checked_add(alive_check_interval)
                        .expect("Failed to set interval, alive check value is too big"),
```

But as I wasn't happy with the `expect` in a task with a value never checked before I moved the sanity checks into a config module within cda-interfaces and implemented the sanity check for `DoIPConfig`.

cda-comm-doip/src/connections.rs
```diff
@@ -425,147 +404,165 @@ fn spawn_connection_reset_task(
...
-                    }
+            let alive_check_interval = gateway_connection
+                .gateway_doip_config
+                .transport
+                .alive_check_interval;
+            let send_timeout = gateway_connection
+                .gateway_doip_config
+                .transport
+                .send_timeout;
+            let alive_check_enabled = alive_check_interval > Duration::ZERO;
```
theswiftfox, line 448: question: can we still set this to 0 to disable? in config.rs the validation throws an error if this is set to 0, right?
`validate_timeout(self.alive_check_interval_secs, "alive_check_interval_secs")?;`

alexmohr: whops, I thought I removed the validate_timeout for the alive check, thanks. 0 is now a valid alive check again.

theswiftfox (dismissed):
Not sure about the clippy comment - to me it looks like the function IS used for the other interval check 🤷‍♀️

Thanks for the upates, LGTM now!

theswiftfox (approved):

PR #391, opened 2026-06-19
fix: Allow running CDA with test data

mbfm:
## Summary

The test data contains an invalid ECU for testing purposes. This causes the CDA to fail during startup, when using the test data.

This changes the included configuration to be more suitable for development use.
For production use, it should be configured differently, like many other configuration values.

---
While implementing this, I also ran into an error message, because I had manually edited the reference configuration. I've made this error message more general and more helpful.  
Here's what the error message looks like in practice now: https://github.com/eclipse-opensovd/classic-diagnostic-adapter/actions/runs/27951065426/job/82707812599?pr=391

## Checklist

## Notes for Reviewers

As discussed with @alexmohr and @theswiftfox.

---

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,13 +63,28 @@ jobs:
             exit 1
           fi
           echo "Generated files are up to date."
-      - name: Verify reference config is up to date
+      - name: Verify reference config matches generated config
         run: |
           cargo run --locked --all-features -- generate-config --output /tmp/opensovd-cda.toml
           if ! diff -u opensovd-cda.toml /tmp/opensovd-cda.toml; then
             echo ""
-            echo "::error::opensovd-cda.toml is out of date."
-            echo "Regenerate with: cargo run --all-features -- generate-config --output opensovd-cda.toml"
+            echo "::error::opensovd-cda.toml does not match configuration generated from code."
+
+            cat <<EOF
+          The opensovd-cda.toml file is a reference generated from the source code.
+          You should not edit it directly and it needs to be regenerated when you
+          edit the configuration fields in the source code.
+
+          You can regenerate it with:
+          cargo run --all-features -- generate-config --output opensovd-cda.toml
+
+          You will need DLT installed as a library on your system.
+          You can install it like so:
+            * On Debian-based systems: apt install libdlt-dev
+            * On macOS with Homebrew:
+              brew tap COVESA/dlt-daemon https://github.com/COVESA/dlt-daemon
+              brew install COVESA/dlt-daemon/dlt-daemon
+          EOF
             exit 1
           fi
           echo "opensovd-cda.toml is up to date."
--- a/cda-database/src/lib.rs
+++ b/cda-database/src/lib.rs
@@ -58,10 +58,9 @@ pub struct DatabaseConfig {
     /// This helps catch typos in ECU names early. When `false` (default),
     /// unmatched per-ECU config entries only produce a warning.
     pub strict_ecu_config: bool,
-    /// When `true`, MDD files that fail to load are skipped and the remaining
-    /// databases continue loading. When `false` (the default), any MDD load
+    /// When `true` (the default), MDD files that fail to load are skipped and
+    /// the remaining databases continue loading. When `false`, any MDD load
     /// failure aborts the entire load/reload operation.
-    #[serde(default)]
     pub ignore_invalid_mdd: bool,
 }
 
@@ -75,7 +74,7 @@ impl Default for DatabaseConfig {
             ignore_protocol: false,
             strict_parameter_validation: false,
             strict_ecu_config: false,
-            ignore_invalid_mdd: false,
+            ignore_invalid_mdd: true,
         }
     }
 }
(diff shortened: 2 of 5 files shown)


alexmohr:
@mbfm The example configuration is generated, to make sure it always stays consistent with the implementation. The documentation comments come from the doc strings on the fields, so to make the build happy you have to update the docstring instead (or additionally). You can re-generate the configuration by running (requires dlt to be installed locally)
```
cargo run --all-features --package opensovd-cda --bin opensovd-cda -- generate-config  --output opensovd-cda.toml
```

alexmohr:
This solves #392

.github/workflows/build.yml
```diff
@@ -63,13 +63,24 @@ jobs:
...
+            cat <<EOF
+          The opensovd-cda.toml file is a reference generated from the source code.
+          You should not edit it directly and it needs to be regenerated when you
+          edit the configuration fields in the source code.
+
+          You can regenerate it with:
+          cargo run --all-features -- generate-config --output opensovd-cda.toml
+
+          You will need DLT installed as a library on your system.
+          On Debian-based systems, you can install it with: apt install libdlt-dev
```
alexmohr: awesome, thanks for adding this! 
nitpick here, could you add the documentation for MacOS too? 
```
brew tap COVESA/dlt-daemon https://github.com/COVESA/dlt-daemon
brew install COVESA/dlt-daemon/dlt-daemon
```

We will leave the windows users behind though, they will be on their own 🤷

cda-database/src/lib.rs
```diff
@@ -58,9 +58,13 @@ pub struct DatabaseConfig {
     /// This helps catch typos in ECU names early. When `false` (default),
     /// unmatched per-ECU config entries only produce a warning.
     pub strict_ecu_config: bool,
-    /// When `true`, MDD files that fail to load are skipped and the remaining
-    /// databases continue loading. When `false` (the default), any MDD load
+    /// When `true` (the default), MDD files that fail to load are skipped and
+    /// the remaining databases continue loading. When `false`, any MDD load
     /// failure aborts the entire load/reload operation.
-    #[serde(default)]
+    #[serde(default = "default_ignore_invalid_mdd")]
```
alexmohr: question: Should we remove the serde defautl attribute here? It's the only parameter that has it and we implement `default` for the object anyway or should we create a follow up, that the other fields get a serde default too, which would have the advantage that a user can leave fields out of the config and the serde default will be used (based on who you're asking this could be a disadvantage too). 

It would make sense that we follow the same principle for all config parameters (and yes I am aware that I introduced this 🙈 )

.github/workflows/build.yml
mbfm: Thanks, I've added the macOS instructions. 🙂

And yeah, I wouldn't want to provide Windows instructions, when we don't have anyone to try them out. Wrong instructions can be a lot worse than no instructions...

theswiftfox: If you have an idea for the windows instructions, i can verify them, as i have a windows machine setup for the FOSS cda development (needed that to fix the openssl import on windows xD). But I think the bigger concern there is if DLT can be built on windows. I can test that later this afternoon.

alexmohr: @theswiftfox You have to at least use something like mingw, dlt-daemon is heavily relying on the posix api and does not have special windows treatment right now.

mbfm: They do have documentation for compiling in Cygwin, which is probably the best we're going to get: https://covesa.github.io/dlt-daemon/doc/dlt_on_Cygwin.html
I really don't know, though, if this allows using the library for development outside of Cygwin.

Of course, we could also tell Windows devs to do their development in WSL or similar. I don't know the limitations of that, but it should allow building the library and using it like on Linux.

theswiftfox: Yeah, it works to build with cygwin, but i had to disable -Werror as the build otherwise failed. The main problem is, that you then need to build the CDA through cygwin as dlt.h depends on other headers not present in windows and is used when compiling dlt-wrapper.c

otherwise that built dll should be perfectly usable as long as cygwin dll is found in the path (and we probably would need to link cygdlt-3.dll):
```
Dump of file ..\dlt-daemon\build\src\lib\libdlt.dll.a

File Type: LIBRARY

  Summary

          14 .idata$2
         C78 .idata$4
         C78 .idata$5
        2D3C .idata$6
         648 .idata$7
         BD8 .text
```
```
Dump of file ..\dlt-daemon\build\src\lib\cygdlt-3.dll

File Type: DLL

  Image has the following dependencies:

    cygwin1.dll
    KERNEL32.dll
```
Not sure if we can find a solution / workaround for the dlt-wrapper.c maybe @alexmohr has an idea?

alexmohr: Not really, I have a PR open on the dlt-wrapper which might makes that even slightly worse, because it always generates the code bindings based on the installed dlt library. I don't think there is a good solution here because we will need the DLT header at link time. I personally wouldn't overcomplicate a scenario almost no one is going to use anyway. Even if you are on windows and made a configuration change, the pipeline logs the diff you worst case is that you have to apply that diff to the config yourself.

theswiftfox: Yeah thats reasonable. Though its a bit frustrating that something like generating the full config depends on external libraries being installed 🙈

cda-database/src/lib.rs
mbfm: I've removed the `#[serde(default)]` annotation and instead changed it in the `Default` trait implementation. I also fixed that the actual default values being used, were set in `Configuration::default()`.

alexmohr (approved):

PR #395, opened 2026-06-23
docs(sovd-api): add locks requirements and architecture

floroks:
## Summary

Add 9 requirements and  architecture elements covering the SOVD lock API: CRUD endpoints, exclusivity, expiration, ECU and functional-group enforcement, vehicle-level blocking, priority preemption, defunct lock lifecycle, and HTTP 409 error semantics.

Also improve plantuml auto-detection in conf.py by falling back to shutil.which() before the hardcoded JAR path,
and fixes an error in the embedded files requirements (wrong path).

--

Adds three new global lock configuration options, each documented as a requirement
and matching architecture entry, enum-typed for future extensibility:

- lock_requirement_policy (require_for_write_operations | require_for_all_operations):
  controls whether a lock is required only for write endpoints, or for all endpoints
  including reads.
- lock_acquisition_policy (unrestricted | vehicle_lock_required): controls whether an
  active vehicle lock must exist before an ECU or functional group lock can be
  acquired.
- lock_exclusivity_policy (exclusive_by_default | non_exclusive_by_default): controls
  the value substituted for x_sovd2uds_isexclusive when a lock request omits it.

Updates req\~sovd-api-lock-api and req\~sovd-api-lock-exclusivity (and their
architecture counterparts) to reference the new configurable exclusivity default.

Also some fixes for mdd-signatures.

## Checklist

### DIFF
--- a/docs/02_requirements/02_sovd.rst
+++ b/docs/02_requirements/02_sovd.rst
@@ -512,6 +512,324 @@ mdd-files, or to prepare for operations which affect the whole vehicle (e.g. dis
          - Functions affecting the whole vehicle (i.e. communication disable/enable)
 
 
+.. _requirements_sovd_api_locks:
+
+Locks
+^^^^^
+
+Locks allow clients to claim exclusive or non-exclusive access to a diagnostic entity
+(vehicle, ECU, or functional group) and to coordinate diagnostic communication across
+multiple concurrent clients. The same CRUD API and enforcement semantics apply to all
+three lock scopes.
+
+.. req:: Lock API
+    :id: req~sovd-api-lock-api
+    :links: arch~sovd-api-lock-api
+    :status: draft
+
+    All three lock scopes -- vehicle (``/locks``), ECU (``/components/{ecu-name}/locks``),
+    and functional group (``/functions/functionalgroups/{group-name}/locks``) -- must expose
+    the following operations:
+
+    .. list-table:: Lock endpoints
+       :header-rows: 1
+       :widths: 10 40 50
+
+       * - Method
+         - Path
+         - Description
+       * - POST
+         - ``.../locks``
+         - Acquire a lock. If the caller already owns the active lock, extends its expiration
+           instead.
+       * - GET
+         - ``.../locks``
+         - List active and defunct locks for the entity. The ``owned`` flag is relative to
+           the caller.
+       * - GET
+         - ``.../locks/{id}``
+         - Return expiration details for a specific lock.
+       * - PUT
+         - ``.../locks/{id}``
+         - Extend the expiration of an existing lock. Only the lock owner may call this.
+       * - DELETE
+         - ``.../locks/{id}``
+         - Release the lock. Only the lock owner may call this.
+
+    The request body for POST and PUT must support the following fields:
+
+    .. list-table:: Lock request fields
+       :header-rows: 1
+       :widths: 25 20 55
+
+       * - Field
+         - Type
+         - Description
+       * - ``lock_expiration``
+         - unsigned integer
+         - Duration in seconds from the time of the request after which the lock
+           automatically expires.
+       * - ``break_lock``
+         - boolean (optional, default ``false``)
+         - Whether an existing lock shall be broken/preempted. See :need:`req~sovd-api-lock-priority`.
+       * - ``x_sov2uds_isexclusive``
+         - boolean (optional, default ``false``)
+         - Whether the lock is exclusive. See :need:`req~sovd-api-lock-exclusivity`.
+       * - any
+         - any (optional)
+         - Vendor-specific data passed to the vendor specific lock preemption mechanism.
+           See :need:`req~sovd-api-lock-priority`.
+
+    The POST response must include ``id`` (UUID string) and ``owned`` (boolean, always
+    ``true`` for the creating client). The GET ``/locks/{id}`` response must include
+    ``lock_expiration`` as an ISO 8601 string.
+
+    Keys and Values must be case-insensitive.
+
+    **Rationale**
+
+    A unified lock API across all three scopes reduces client implementation complexity and
+    ensures consistent lock management regardless of the diagnostic entity type.
+
+
+.. req:: Lock Exclusivity
+    :id: req~sovd-api-lock-exclusivity
+    :links: arch~sovd-api-lock-exclusivity
+    :status: draft
+
+    A lock must be either exclusive or non-exclusive, controlled by the ``x_sovd2uds_isexclusive`` field
+    in the POST/PUT request body (default: ``false``).
+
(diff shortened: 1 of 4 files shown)


docs/02_requirements/02_sovd.rst
```diff
@@ -512,6 +512,324 @@ mdd-files, or to prepare for operations which affect the whole vehicle (e.g. dis
...
+    ``/functions/functionalgroups/{group-name}/operations``, and
+    ``/functions/functionalgroups/{group-name}/modes``.
+
+
+.. req:: Vehicle Lock Blocks Child Lock Acquisition
+    :id: req~sovd-api-lock-vehicle-blocking
+    :links: arch~sovd-api-lock-vehicle-blocking
+    :status: draft
+
+    When a vehicle lock is held by a client, any other client's attempt to acquire an ECU
```
alexmohr:
```suggestion
    When a vehicle lock is held by a client, any other clients attempt to acquire an ECU
```

docs/02_requirements/02_sovd.rst
```diff
@@ -512,6 +512,324 @@ mdd-files, or to prepare for operations which affect the whole vehicle (e.g. dis
...
+    The vehicle lock owner may still acquire ECU or functional group locks.
+
+    **Rationale**
+
+    The vehicle lock represents a claim over the complete set of diagnostic entities on the
+    vehicle. Permitting other clients to acquire child locks while the vehicle lock is held
+    would violate that claim.
+
+
+.. req:: Lock Priority Preemption
```
alexmohr, line 871: q: Is this an optional mechanism that can be disabled or will we make this mandatory? Not sure this is wanted by every OEM.

docs/02_requirements/02_sovd.rst
```diff
@@ -512,6 +512,324 @@ mdd-files, or to prepare for operations which affect the whole vehicle (e.g. dis
...
+    another client holds the lock. Delegating the preemption decision to a vendor mechanism
+    allows vendors to implement any priority scheme without modifying the CDA core.
+
+
+.. req:: Defunct Lock State
+    :id: req~sovd-api-lock-defunct
+    :links: arch~sovd-api-lock-defunct
+    :status: draft
+
+    When a lock is preempted by a higher-priority client it must transition to a
```
alexmohr, line 916: q: I do understand that certain operations like flashing should be done by clients with a high prio, so they won't get pre-empted. But what if a low prio client starts flashing too and then gets pre-empted. Do we abort to flash transfer or do we deny the pre-emption in this case? (I.e. the 'apply' of an runtime update is rejected when a flash transfer is running)
An alternative could be that certain operations are bound to a given lock priority, but that's probably something to implement in OEM side but OpenSOVD would have to provide the interfaces for this.

docs/02_requirements/02_sovd.rst
```diff
@@ -512,6 +512,324 @@ mdd-files, or to prepare for operations which affect the whole vehicle (e.g. dis
...
+      at preemption time.
+    - The defunct lock must remain visible in ``GET /locks`` until its original expiration
+      time elapses, or it is deleted.
+    - The defunct lock response must include the standard lock fields (``id``, ``owned``,
+      ``lock_expiration``) together with the following vendor-prefixed extension fields:
+
+      - ``x_sovd2uds_broken_by`` -- the identity of the preempting client, as returned
+        by the vendor mechanism.
+      - ``x_sovd2uds_broken_at`` -- the ISO 8601 timestamp at which preemption occurred.
+      - ``x_sovd2uds_current_holder`` -- identity of the current lock holder
```
alexmohr, line 929: q: What's the usecase for broken by **and** current holder, in most cases they will be equal, except the  In general, wouldn't it be nice then to return the current lock holder when locking a resource is rejected?

docs/03_architecture/02_sovd-api/02_sovd-api.rst
```diff
@@ -755,6 +755,226 @@ Faults -- SID 14\ :sub:`16` & 19\ :sub:`16`
...
+    A ``validate_fg_lock`` check is applied at the entry of every functional group
+    communication endpoint. It follows the same decision logic as
+    :need:`arch~sovd-api-lock-ecu-enforcement`, operating on the functional group lock
+    and the functional group endpoints (``/data``, ``/operations``, ``/modes``).
+
+    See also :need:`arch~sovd-api-functional-communication-locks` for the Tester Present
+    side-effects associated with acquiring and releasing a functional group lock.
+
+
+.. arch:: Vehicle Lock Blocks Child Lock Acquisition
```
alexmohr, line 978: question: If a low prio client is holding a vehicle lock and the high prio client wants an ECU lock, will the low prio client be pre-empted anyway? Same question for functional groups, when the high prio client wants an ECU that is part of a low prio locked functional group.

alexmohr (commented):
couple of questions, mainly about the lock pre-emption :)

docs/02_requirements/02_sovd.rst
floroks: yes, you get the option to implement this in a plugin (e.g. security), if you don't implement it, the standard is for the lock request to be denied

see `req~sovd-api-lock-priority:

> If no vendor mechanism is configured, or if the mechanism does not grant preemption, conflicting POST requests are rejected with HTTP 423 as normal.

docs/02_requirements/02_sovd.rst
floroks: In practice it'll not just be a client, but a client + additional vendor specific additional information, which can be used to distinguish between different use-cases from the same client. Usually only a single client with a single specific additional data would be granted this type of prioritization which should then be the highest, but it's up to individual OEMs to define that.

docs/02_requirements/02_sovd.rst
floroks: In most cases yes, but when a lock was only broken for a short time, and there is no current holder, the client can directly reattempt to get a lock. this would mostly be the case when someone is doing manual diagnostics (e.g. low frequency of request), and a quick higher prioritizes automated client doing a quick read from data.

docs/03_architecture/02_sovd-api/02_sovd-api.rst
floroks: lock preemption is only defined for vehicle locks, because it requires additional data to the client id. So a client can't request a higher prio lock for the ecu only, so it'll be rejected.

docs/02_requirements/02_sovd.rst
floroks: fixed

docs/02_requirements/02_sovd.rst
floroks: returning the current lockholder when locking is rejected can also be added, but it would then be an additional parameter in the locking response error parameters for the regular lock request

theswiftfox (approved):
lgtm :)

PR #399, opened 2026-06-25
feat(config): consolidate strict flags into dedicated [strict] section

### LINKED ISSUE #377: Improvement: add a sctrict config
## Description

Introduce a dedicated `[strict]` configuration section that groups all individual strict-mode flags (e.g. `strict_parameter_validation`, `strict_ecu_config`) under a single namespace, and add a top-level `enabled` flag within that section that sets all individual flags to `true` at once.

Currently each strict check is a standalone field scattered across different config sections (e.g. `[database].strict_parameter_validation`, `[database].strict_ecu_config`). As the number of such checks grows, configuration becomes harder to discover and manage.

## Motivation

This was raised during the review of PR #342, where the question was whether `strict_parameter_validation` should remain a per-feature flag or be part of a broader strict mode. The agreed direction is to keep individual flags for granular control, but also provide a master switch to enable all of them at once.

Benefits:
- Easier discoverability: all strict checks are co-located in config.
- Simpler onboarding: users who want fully strict behavior set one flag instead of hunting down each individual option.
- Consistent pattern for future strict checks to follow.

## Proposed Solution

Add a `[strict]` section to `opensovd-cda.toml` with:
- An `enabled` boolean that, when `true`, overrides all individual strict flags to `true`.
- Individual flags mirroring the existing ones (e.g. `parameter_validation`, `ecu_config`) so users can still opt into specific checks without enabling everything.

Existing flags in their original sections should be removed, so settings are not duplicated.

## Additional Context

> [!IMPORTANT]

shegazyy:
Move strict_parameter_validation and strict_ecu_config out of [database] into a new top-level [strict] config section with a master enabled switch that activates all checks at once. Individual flags remain available for granular control.

Closes #377

## Summary

Consolidates scattered strict-mode flags (`strict_parameter_validation`, `strict_ecu_config`) from the `[database]` config section into a new top-level `[strict]` section. Adds a master `enabled` switch that activates all strict checks at once, while individual flags (`parameter_validation`, `ecu_config`) remain available for granular control. This improves discoverability and provides a consistent pattern for future strict checks.

## Checklist

## Related

Follow-up to #342 (strict_parameter_validation feature)

## Notes for Reviewers

- `StrictConfig` resolution: `config.strict.parameter_validation()` returns `self.enabled || self.parameter_validation`, so the master switch OR individual flags activate each check.
- All fields use `#[serde(default)]`, so existing configs without a `[strict]` section continue to work (all checks default to disabled).
- The reference config (`opensovd-cda.toml`) was regenerated via `cargo run --all-features -- generate-config`.
- 4 new unit tests cover: master switch, individual flags, defaults, and TOML parsing.
- Error message in `handle_ecu_config_keys` updated to reference the new `[strict] ecu_config` path.

### DIFF
--- a/cda-main/src/config/configfile.rs
+++ b/cda-main/src/config/configfile.rs
@@ -131,6 +131,35 @@ impl schemars::JsonSchema for EcuComParams {
     }
 }
 
+/// Strict-mode flags that opt in to stricter runtime validation.
+///
+/// When `enabled` is `true`, every individual check is activated regardless
+/// of its own value.  Individual flags can also be turned on independently
+/// for more granular control.
+#[derive(Deserialize, Serialize, Clone, Debug, Default, schemars::JsonSchema)]
+pub struct StrictConfig {
+    /// Master switch - when `true`, all individual strict checks are enabled.
+    pub enabled: bool,
+    /// Reject requests containing parameters not defined in the diagnostic
+    /// service with a `BadPayload` error (HTTP 400).
+    pub parameter_validation: bool,
+    /// Exit with an error if any key under `[ecu.<name>]` does not match a
+    /// loaded MDD database.
+    pub ecu_config: bool,
+}
+
+impl StrictConfig {
+    #[must_use]
+    pub fn parameter_validation(&self) -> bool {
+        self.enabled || self.parameter_validation
+    }
+
+    #[must_use]
+    pub fn ecu_config(&self) -> bool {
+        self.enabled || self.ecu_config
+    }
+}
+
 /// Top-level application configuration.
 #[derive(Deserialize, Serialize, Clone, Debug, schemars::JsonSchema)]
 pub struct Configuration {
@@ -159,10 +188,11 @@ pub struct Configuration {
     pub faults: FaultConfig,
     /// Per-ECU configuration blocks keyed by ECU short name (case-insensitive
     /// match against the MDD short name returned by `load_proto_data`).
-    #[serde(default)]
     pub ecu: HashMap<String, EcuConfig>,
     /// Configuration for update plugin, i.e. storage paths
     pub runtime_update_config: RuntimeUpdateConfig,
+    /// Strict-mode validation flags.
+    pub strict: StrictConfig,
 }
 
 /// Per-ECU configuration block.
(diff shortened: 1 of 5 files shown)


shegazyy:
Hey @alexmohr 🙏 PR is ready for review. Let me know if anything else needs adding!

cda-main/src/config/configfile.rs
```diff
@@ -131,6 +131,38 @@ impl schemars::JsonSchema for EcuComParams {
     }
 }
 
+/// Strict-mode flags that opt in to stricter runtime validation.
+///
+/// When `enabled` is `true`, every individual check is activated regardless
+/// of its own value.  Individual flags can also be turned on independently
+/// for more granular control.
+#[derive(Deserialize, Serialize, Clone, Debug, Default, schemars::JsonSchema)]
+pub struct StrictConfig {
```
alexmohr, line 140: suggestion: In https://github.com/eclipse-opensovd/classic-diagnostic-adapter/commit/7ee82be6bcbbbf0bb2776ffcd4c953ace943aaf9 @mbfm unified the configuration subtypes and removed `serde(default)` from them. As you implement `Default` anyway, you can ommit the serde default like Frank did on the linked commit

alexmohr (commented):
minor thing for consistency with other types, everything else looks great. thank you.

cda-main/src/config/configfile.rs
```diff
@@ -163,6 +192,9 @@ pub struct Configuration {
     pub ecu: HashMap<String, EcuConfig>,
     /// Configuration for update plugin, i.e. storage paths
     pub runtime_update_config: RuntimeUpdateConfig,
+    /// Strict-mode validation flags.
+    #[serde(default)]
+    pub strict: StrictConfig,
```
alexmohr, line 195: question: Do we need this serde default here? @mbfm any reason line 191 still has the 'default', I'm not sure if it was missed or if it is actually needed.

mbfm: I only removed `#[serde(default)]` from `DatabaseConfig::ignore_invalid_mdd` and separately moved the `Default` implementations from `Configuration` to the subtypes.

I did not generally remove the `#[serde(default)]` from the configuration types, although I guess, this could probably be done. The behavior did not seem different to just implementing the `Default` trait.

alexmohr: thank you, in that case @shegazyy, could you remove the serde default here, because we derive default anyway. You can remove it from 191 too if you want :) (if there are remaining serde default we can open a follow up ticket)

alexmohr:
@shegazyy thank you :) lgtm

alexmohr (approved):

PR #401, opened 2026-06-26
improve and fix configuration and communication related issues

alexmohr:
## Summary
* [improve(logging): improve the error message during comparam lookup](https://github.com/eclipse-opensovd/classic-diagnostic-adapter/commit/b4b05b38eaba527af0f028d01f256f6283cc9815)
* [fix(doip): accept an response without prio ack](https://github.com/eclipse-opensovd/classic-diagnostic-adapter/commit/ac4f2ff8e77709ae5385b627e5b5621844244e9a)
* [fix(payload): fix encoding text table in coded const](https://github.com/eclipse-opensovd/classic-diagnostic-adapter/commit/bd9d028e62994fc5eb8c3141c398369547636efe)
* [feat(config): Add configuration for diag comm service_lookup](https://github.com/eclipse-opensovd/classic-diagnostic-adapter/commit/38649b1b959fb4430ffa111ba1789fabfc20158b)

## Checklist

## Notes for Reviewers

These fixes issues found during vehicle test. 

### DIFF
--- a/cda-comm-doip/src/lib.rs
+++ b/cda-comm-doip/src/lib.rs
@@ -597,18 +597,35 @@ impl<T: EcuAddresses + DoipComParams> EcuGateway for DoipDiagGateway<T> {
                                             )
                                             .await;
                                         }
+                                        Ok(
+                                            response @ (DiagnosticResponse::Msg(_)
+                                            | DiagnosticResponse::Pending(_)
+                                            | DiagnosticResponse::BusyRepeatRequest(_)
+                                            | DiagnosticResponse::TemporarilyNotAvailable(
+                                                _,
+                                            )),
+                                        ) => {
+                                            // Some ECUs do not send an ACK before the
+                                            // diagnostic response. Treat a valid diagnostic
+                                            // response as an implicit ACK and forward it
+                                            // directly to the response channel.
+                                            tracing::debug!(
+                                                "Received diagnostic response before ACK, \
+                                                 treating as implicit ACK"
+                                            );
+                                            try_send_uds_response(
+                                                &response_sender,
+                                                response.try_into(),
+                                            )
+                                            .await;
+                                            break 'ack_waiting true;
+                                        }
                                         Ok(msg) => {
                                             tracing::warn!(
                                                 ecu_name = %transmission_params.ecu_name,
                                                 "Expected ACK/NACK but received unexpected message: {:?}",
                                                 msg
                                             );
-                                            // any response but ACK/NACK is unexpected because
-                                            // every sent message should be answered with
-                                            // ACK or NACK before sending anything else.
-                                            // however, we should still continue waiting
-                                            // for ACK/NACK in case we get something unexpected,
-                                            // as some ECUs might not follow the spec properly.
                                             continue 'ack_waiting;
                                         }
                                         Err(e) => {
--- a/cda-comm-doip/src/lib.rs
+++ b/cda-comm-doip/src/lib.rs
@@ -19,8 +19,8 @@ use std::{
 
 use cda_interfaces::{
     DiagServiceError, DoipComParams, DoipGatewaySetupError, EcuAddresses, EcuConnectivityHandler,
-    EcuGateway, HashMap, HashMapExtensions, ServicePayload, TransmissionParameters, UdsResponse,
-    dlt_ctx,
+    EcuGateway, HashMap, HashMapExtensions, ServicePayload, TransmissionParameters,
+    UDS_ID_RESPONSE_BITMASK, UdsResponse, dlt_ctx,
     util::{self, tokio_ext},
 };
 use doip_definitions::{
@@ -69,14 +69,43 @@ const NRC_TEMPORARILY_NOT_AVAILABLE: u8 = 0x94;
 #[derive(Debug, Clone)]
 enum DiagnosticResponse {
     Msg(DiagnosticMessage),
-    Pending(u16),
+    Pending {
+        source_address: u16,
+        request_sid: u8,
+    },
     Ack((u16, Vec<u8>)),
     Nack(DiagnosticMessageNack),
     AliveCheckResponse,
     TesterPresentNRC(u8),
     GenericNack(GenericNack), // todo #22 -> we need the address of the ECU that sent the nack
-    BusyRepeatRequest(u16),
-    TemporarilyNotAvailable(u16),
+    BusyRepeatRequest {
+        source_address: u16,
+        request_sid: u8,
+    },
+    TemporarilyNotAvailable {
+        source_address: u16,
+        request_sid: u8,
+    },
+}
+
+impl DiagnosticResponse {
+    fn matches_request(&self, request: &[u8]) -> bool {
+        let Some(sid) = request.first() else {
+            return false;
+        };
+
+        match self {
+            Self::Ack((_, previous)) => request.starts_with(previous),
+            Self::Msg(msg) => msg
+                .message
+                .first()
+                .is_some_and(|response_sid| *response_sid == (*sid | UDS_ID_RESPONSE_BITMASK)),
+            Self::Pending { request_sid, .. }
+            | Self::BusyRepeatRequest { request_sid, .. }
+            | Self::TemporarilyNotAvailable { request_sid, .. } => *request_sid == *sid,
+            _ => false,
+        }
+    }
 }
 
 pub(crate) struct DoipGatewayState<T: EcuAddresses + DoipComParams> {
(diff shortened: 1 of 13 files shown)


cda-comm-doip/src/lib.rs
```diff
@@ -601,18 +601,35 @@ impl<T: EcuAddresses + DoipComParams> EcuGateway for DoipDiagGateway<T> {
...
+                                            tracing::debug!(
+                                                "Received diagnostic response before ACK, \
+                                                 treating as implicit ACK"
+                                            );
+                                            try_send_uds_response(
+                                                &response_sender,
+                                                response.try_into(),
+                                            )
+                                            .await;
+                                            break 'ack_waiting true;
```
theswiftfox: question: should that not be a return instead of a break? as we already return the message here? could this cause us to send the reply double through the channel if we use break with true here?

cda-core/src/diag_kernel/payload_encode.rs
```diff
@@ -706,6 +681,54 @@ impl<S: SecurityPlugin> EcuManager<S> {
...
+            "Required parameter '{param_name}' missing",
+        ))),
+    }
+}
+
+fn phys_value_str_to_json(
+    value: &str,
+    param_short_name: &str,
+) -> Result<serde_json::Value, DiagServiceError> {
+    if value.chars().all(char::is_numeric) {
```
theswiftfox: question: even though i suggested to use is_numeric when we were testing, i realized that this will not work for negative numbers. Not sure if we should instead check if the first character is '-' and allow that and then check the rest for 'is_numeric' or if it fallls within the ascii range for the characters '0'-'9' (should be `.all(|c| c.is_ascii_digit())`)

alexmohr: Good point, instead of dealing with the parsing manually, I opted for this instead:

```rust
if let Ok(number) = value.parse::<i64>() {
    Ok(serde_json::Value::Number(number.into()))
} else if let Ok(number) = value.parse::<u64>() {
    Ok(serde_json::Value::Number(number.into()))
} else {
    Ok(serde_json::Value::from(value))
}

```

cda-comm-doip/src/lib.rs
alexmohr: Return and break are pretty much the same here, both set ack_recevied to `true`. It's a bit confusing to read imho, I will add a test for this and extract the wait for ack into a function that returns an Enum with the 3 possible states. 
We do have to run the second loop because the first response could be a NRC like busy repeat, in which case we do want the receiver of the uds frames to give the chance to sent the message again instead of closing the channel right away by returning from the entire function.

theswiftfox: Yeah my concern is rather that we both set the ack_waiting to true *and* already send the response via channel and then continue the regular behaviour. I'm mostly thinking about edge cases where we get something unexpected then and simply hand that to the caller as a response.
But the suggestion with the enum sounds good!

alexmohr: Refactored into 

```rust 
async fn wait_for_ack_or_response(
    receiver: &mut broadcast::Receiver<Result<DiagnosticResponse, EcuError>>,
    ecu_name: &str,
    sent_message: &DiagnosticMessage,
    timeout: Duration,
) -> Result<Option<DiagnosticResponse>, DiagServiceError> 
```

instead, so `read_ecu_responses` is the only part that is dealing with responses. 
I see your concern about `something unexpected then and simply hand that to the caller as a response.`. The only case where this would happen where it didn't happen before was if we get a random response that does not match the request sid. 
Before accepting a response this was guarded by the ACKs previous message check. To make sure we do not regress here, the ack check is now also validating the response id by checking `first byte | UDS_ID_RESPONSE_BITMASK` on responses and the request sid from nacks

cda-core/src/diag_kernel/payload_encode.rs
```diff
@@ -708,6 +683,48 @@ impl<S: SecurityPlugin> EcuManager<S> {
...
+            })?;
+            Ok(phys_value_str_to_json(str_val.as_ref()))
+        }
+        _ => Err(DiagServiceError::InvalidRequest(format!(
+            "Required parameter '{param_name}' missing",
+        ))),
+    }
+}
+
+fn phys_value_str_to_json(value: &str) -> serde_json::Value {
```
theswiftfox, line 716: q: should we also handle float values here? or make the name reflect that its only integers?

cda-core/src/diag_kernel/payload_encode.rs
```diff
@@ -732,7 +749,21 @@ fn process_coded_constants(
                         "Param '{}' is missing coded value",
                         param.short_name().unwrap_or_default()
                     )))?;
-            let const_json_value = str_to_json_value(coded_const_value, diag_type.base_datatype())?;
+
+            let const_json_value = if coded_const_value.chars().all(char::is_numeric) {
```
theswiftfox: q: should we do the same here as with `phys_value_str_to_json`? eg. try to parse to different integer (and maybe float) variants? this would then also handle the case were the coded const is negative. as of now the is_numeric would not fire for negative numbers ig.

theswiftfox:
Not sure where to put this, so I'll add the comment here:
PR descriptions states '[fix(connection): do not send a diagnostic ack when it is disabled](https://github.com/eclipse-opensovd/classic-diagnostic-adapter/commit/2943dca164d8e65eb3ffbd89b3a6d469ddce68cb)' but either I was just really bad at reading the code or this now got completely removed? If the latter, can we document a reason for that as per spec we theoretically should support it?

cda-comm-doip/src/lib.rs
```diff
@@ -66,14 +67,43 @@ const NRC_TEMPORARILY_NOT_AVAILABLE: u8 = 0x94;
...
+}
+
+impl DiagnosticResponse {
+    fn matches_request(&self, request: &[u8]) -> bool {
+        let Some(sid) = request.first() else {
+            return false;
+        };
+
+        match self {
+            Self::Ack((_, previous)) => previous.is_empty() || request.starts_with(previous),
```
theswiftfox: q: does an ack without a 'previous' set make sense? we don't have any way to know what was actually acked then, right?
But i assume there was a reason to explicitly allow `is_empty()`, so can we maybe add a comment why?

cda-core/src/diag_kernel/payload_encode.rs
alexmohr: added a branch for try parse f64

alexmohr:
> Not sure where to put this, so I'll add the comment here: PR descriptions states '[fix(connection): do not send a diagnostic ack when it is disabled](https://github.com/eclipse-opensovd/classic-diagnostic-adapter/commit/2943dca164d8e65eb3ffbd89b3a6d469ddce68cb)' but either I was just really bad at reading the code or this now got completely removed? If the latter, can we document a reason for that as per spec we theoretically should support it?

I will remove the commit during the final squashing after approval. This logic is removed because the UDS specification in section 9.5 documents that a client should not sent any ACK. 

> s per spec we theoretically should support it

So no per spec we should not support it.

cda-core/src/diag_kernel/payload_encode.rs
alexmohr: fixed, extracted logic from `phys_value_str_to_json` and reuse here.

cda-comm-doip/src/lib.rs
alexmohr: well no actually that wasn't something we discovered during vehicle testing, it was just a precaution, but now that you're mentioning it, I'd rather remove this.

theswiftfox:
> > Not sure where to put this, so I'll add the comment here: PR descriptions states '[fix(connection): do not send a diagnostic ack when it is disabled](https://github.com/eclipse-opensovd/classic-diagnostic-adapter/commit/2943dca164d8e65eb3ffbd89b3a6d469ddce68cb)' but either I was just really bad at reading the code or this now got completely removed? If the latter, can we document a reason for that as per spec we theoretically should support it?
> 
> I will remove the commit during the final squashing after approval. This logic is removed because the UDS specification in section 9.5 documents that a client should not sent any ACK.
> 
> > s per spec we theoretically should support it
> 
> So no per spec we should not support it.

But, this was not a UDS ack, but a DoIP ack? 
Maybe I'm missreading here, but it states:
> Each DoIP entity shall support the diagnostic message positive acknowledgement as specified in Table 23.

and
> Each DoIP entity shall send the diagnostic message positive acknowledgement with ACK code set to 0016 (see
Table 24) immediately after the diagnostic message has been correctly processed and copied into the
destination network transmission buffer.

So to me this reads like we should support sending an Ack to a gateway? Or are we not a DoIP entity as tester? -> i'd be fine with making it configurable, but I could see gateways that expect / enforce acks not being happy if we completely remove it?

alexmohr:
> So to me this reads like we should support sending an Ack to a gateway? Or are we not a DoIP entity as tester? -> i'd be fine with making it configurable, but I could see gateways that expect / enforce acks not being happy if we completely remove it?

> Diagnostic messages can also be sent by the DoIP entities, for
example, when transmitting a diagnostic response or an unsolicited message (e.g. response on event) from
a server DoIP entity to the client DoIP entity. In this case, the diagnostic messages are not acknowledged by
the client DoIP entity.

As discussed the spec expliclity mentioned that responses shall not be ACKd

theswiftfox (approved):

Issue #30, opened 2025-08-21
Publish documentation to github pages


floroks:
Only main, or how to handle releases/tags?

floroks:
this is implemented

PR #405, opened 2026-06-29
feat(server): add remote_address config for user-facing URLs

### LINKED ISSUE #397: Improvement: CDA should know remote address used to reach it
## Description

Currently, CDA only has the concept of a bind address, typically set to `0.0.0.0`.
However, when users want to reach CDA, they need to use the remote address, typically `127.0.0.1`.
CDA should be made aware of this remote address to allow providing it to users.

## Motivation

Having the remote address available would allow setting it as the default address in the Swagger-UI, which allows using the Swagger-UI to send requests to the CDA.

We could also log these addresses during the CDA startup, to document where this Swagger-UI can be reached, as well as potentially other interfaces that we're listening on.

## Proposed Solution

A configuration option should be added to allow specifying the remote address, with a default of `127.0.0.1`.

This remote address should be passed into the Swagger-UI and be logged during startup.

## Additional Context

The Swagger-UI currently tries to make requests to `0.0.0.0`, which cannot work:

Image

> [!IMPORTANT]

shegazyy:
The bind address (typically 0.0.0.0) is not reachable from a browser, so Swagger-UI was generating unusable server URLs. Add a remote_address field to ServerConfig (default 127.0.0.1) and use it for the OpenAPI server URL and startup log.

Closes #397

## Summary

Adds a `remote_address` field to `[server]` config (default `127.0.0.1`) used for user-facing URLs. The bind address (`0.0.0.0`) was being passed to Swagger-UI as the OpenAPI server URL, making API calls from the browser fail. Now Swagger-UI points to the remote address, and the startup log includes the full Swagger-UI URL for discoverability.

## Checklist

## Related

Closes #397

## Notes for Reviewers

- Bind address (`address`) is unchanged - still controls the TCP listener
- `remote_address` only affects the OpenAPI server URL (Swagger-UI) and the startup log message
- Component `href` links are unaffected - they already use the HTTP `Host` header from the incoming request
- Users can override `remote_address` in their config for remote deployments (e.g. `remote_address = "192.168.1.50"`)
- Reference config (`opensovd-cda.toml`) regenerated via `cargo run --all-features -- generate-config`

### DIFF
--- a/cda-main/src/lib.rs
+++ b/cda-main/src/lib.rs
@@ -482,16 +482,30 @@ where
     register_version_endpoints(&dynamic_router).await;
     pre_load_hook(dynamic_router.clone()).await?;
 
+    let display_address = if config.server.address == "0.0.0.0" {
+        "127.0.0.1"
+    } else {
+        &config.server.address
+    };
+    let swagger_ui_url = format!(
+        "http://{}:{}{}",
+        display_address,
+        config.server.port,
+        cda_sovd::SWAGGER_UI_ROUTE
+    );
+
     setup_vehicle_and_routes::<SP, SL>(
         config,
         &dynamic_router,
-        &webserver_config,
         health_state.as_ref(),
         clonable_shutdown_signal.clone(),
     )
     .await?;
 
-    tracing::info!("CDA fully initialized and ready to serve requests");
+    tracing::info!(
+        "CDA fully initialized and ready to serve requests.\nYou can access the REST API \
+         documentation at: {swagger_ui_url}"
+    );
     if let Some(provider) = main_health_provider {
         provider.update_status(cda_health::Status::Up).await;
     }
@@ -530,7 +544,6 @@ pub async fn run_with_config(config: Configuration) -> Result<(), AppError> {
 pub async fn setup_vehicle_and_routes<SP: SecurityPlugin, SL: SecurityPluginLoader>(
     config: Configuration,
     dynamic_router: &cda_sovd::dynamic_router::DynamicRouter,
-    webserver_config: &cda_sovd::WebServerConfig,
     health_state: Option<&cda_health::HealthState>,
     clonable_shutdown_signal: futures::future::Shared<
         impl std::future::Future<Output = ()> + Send + 'static,
@@ -616,8 +629,7 @@ pub async fn setup_vehicle_and_routes<SP: SecurityPlugin, SL: SecurityPluginLoad
     )
     .await;
 
-    cda_sovd::add_openapi_routes(dynamic_router, &vehicle_data.update_guard, webserver_config)
-        .await;
+    cda_sovd::add_openapi_routes(dynamic_router, &vehicle_data.update_guard).await;
 
     // SAFETY: Must be applied AFTER all routes are registered (layer only covers existing routes).
     cda_sovd::install_update_guard(dynamic_router, vehicle_data.update_guard.clone()).await;
--- a/cda-sovd/src/lib.rs
+++ b/cda-sovd/src/lib.rs
@@ -29,6 +29,7 @@ use cda_plugin_security::SecurityPluginLoader;
 use dynamic_router::DynamicRouter;
 pub use dynamic_router::{RouteGroupNotFound, RouteHandle};
 pub use http::Method;
+use opensovd_axum_extra::ExtractHost;
 use tokio::net::TcpListener;
 use tower::{Layer, ServiceExt as TowerServiceExt};
 use tower_http::{normalize_path::NormalizePathLayer, trace::TraceLayer};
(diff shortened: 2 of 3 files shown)


lh-sag:
When bind to `0.0.0.0` should we not return the ip / local address the request is coming from?

mbfm:
> When bind to `0.0.0.0` should we not return the ip / local address the request is coming from?

I don't think, this is possible here, as the Swagger UI / OpenAPI endpoint comes from a third-party library and we need to set its server URL when registering its route in the web server: https://github.com/eclipse-opensovd/classic-diagnostic-adapter/blob/68ab6c39efa53bf1150f78db4a3e968fd4deddc1/cda-sovd/src/lib.rs#L233

But just returning the address with which the request was sent will also stop working, if CDA is running behind a reverse proxy. Then we will see the address under which the reverse proxy reaches out to us, which will not be useful for end users. In particular, this is also the case, when running in a Docker container.  
At least in past projects, I've always found that because of this, you want a configurable remote address sooner or later...

cda-main/src/lib.rs
```diff
@@ -484,7 +492,7 @@ where
     )
     .await?;
 
-    tracing::info!("CDA fully initialized and ready to serve requests");
+    tracing::info!(swagger_ui = %swagger_ui_url, "CDA fully initialized and ready to serve requests");
```
mbfm: I think, it would be better to write this in prose, since this log message is especially helpful for new users. Something like this maybe?

```suggestion
    tracing::info!("CDA fully initialized and ready to serve requests.\nYou can access the REST API documentation at: {swagger_ui_url}");
```

mbfm (approved):
Something seems to have gone wrong with your rebase, as now the changes from #399 seem to also be included here. Pretty sure, I was able to review the right changes regardless, just as a heads-up.

As for the changes, I kind of expected a `remote_port` field to be implemented alongside, but I guess, we can still implement that, if someone has a need for it. It is relatively unusual that the port needs to be reconfigured.

Otherwise looks good. I tried out the changes locally and everything worked as expected. 👍

lh-sag:
> I don't think, this is possible here, as the Swagger UI / OpenAPI endpoint comes from a third-party library and we need to set its server URL when registering its route in the web server:

@mbfm I was not aware that the PR is primary about swagger (had to check the issue). My suggestion was about the href in some SOVD payloads e.g. 
```
curl -s http://127.0.0.1:7690/sovd/v1/components | jq
...
      "href": "http://127.0.0.1:7690/sovd/v1/components/ecu",
...
```
If you bind to `0.0.0.0` what will you return as part of the href host/ip? It should match the request e.g. `Host: ` otherwise you get a href which is not reachable. I have never used swagger-ui, but cant one add the address in openapi.json?
Of course this will not work if an offline openapi description does not match the current network.  

Access via reverse proxy is another topic but can be handled like this https://github.com/eclipse-opensovd/opensovd/issues/30

mbfm:
@lh-sag Well, thanks a lot for the input. I didn't know that context and I had opened that issue expecting this kind of feedback, not for @shegazyy to start implementing it right away. 😅

I'll answer your questions, so that hopefully you can judge, whether there's any merit to the approach in this PR (perhaps in addition to #30). Personally, I'm certainly not deep enough into the project context to make a shout.

> If you bind to `0.0.0.0` what will you return as part of the href host/ip? It should match the request e.g. `Host: ` otherwise you get a href which is not reachable.

The idea with the configured remote address was that you'd always return the configured remote address. It makes the assumption that there is a single address under which users are supposed to reach the CDA server. This might make more sense in a traditional client-server communication, rather than the federated approach of OpenSOVD.

> I have never used swagger-ui, but cant one add the address in openapi.json? Of course this will not work if an offline openapi description does not match the current network.

The server URL is part of the `openapi.json`, yes. And I had to look into it myself, whether it can be changed dynamically, but I do think now that it should be possible.

It is part of an Axum router:

https://github.com/eclipse-opensovd/classic-diagnostic-adapter/blob/68ab6c39efa53bf1150f78db4a3e968fd4deddc1/cda-sovd/src/lib.rs#L230-L235

...which should be able to use an [extractor](https://docs.rs/axum/latest/axum/extract/index.html) to get the HTTP headers for a given request and then return a modified `openapi.json` by setting that as the `server_url` here.

> Access via reverse proxy is another topic but can be handled like this [eclipse-opensovd/opensovd#30](https://github.com/eclipse-opensovd/opensovd/issues/30)

Would be interesting, if Docker sets the correct headers for the forwarded header approach. Then again, I don't know, if we really target running in a Docker container. The `ecu-sim` in the CDA repository runs in Docker, but it wouldn't really be a problem there, if the Swagger-UI uses the wrong URL...

lh-sag:
> @lh-sag Well, thanks a lot for the input. I didn't know that context and I had opened that issue expecting this kind of feedback, not for @shegazyy to start implementing it right away. 😅
> 

This exchange will for sure help us to chose a good implementation for sure.

> I'll answer your questions, so that hopefully you can judge, whether there's any merit to the approach in this PR (perhaps in addition to #30). Personally, I'm certainly not deep enough into the project context to make a shout.
> 
> > If you bind to `0.0.0.0` what will you return as part of the href host/ip? It should match the request e.g. `Host: ` otherwise you get a href which is not reachable.
> 
> The idea with the configured remote address was that you'd always return the configured remote address. It makes the assumption that there is a single address under which users are supposed to reach the CDA server. This might make more sense in a traditional client-server communication, rather than the federated approach of OpenSOVD.
> 
> > I have never used swagger-ui, but cant one add the address in openapi.json? Of course this will not work if an offline openapi description does not match the current network.
> 
> The server URL is part of the `openapi.json`, yes. And I had to look into it myself, whether it can be changed dynamically, but I do think now that it should be possible.
> 
> It is part of an Axum router:
> 
> https://github.com/eclipse-opensovd/classic-diagnostic-adapter/blob/68ab6c39efa53bf1150f78db4a3e968fd4deddc1/cda-sovd/src/lib.rs#L230-L235
> 
> ...which should be able to use an [extractor](https://docs.rs/axum/latest/axum/extract/index.html) to get the HTTP headers for a given request and then return a modified `openapi.json` by setting that as the `server_url` here.
> 

And use the ip/host in SOVD payloads with href semantic. But that can be implemented in a separate issue/PR.
 
> > Access via reverse proxy is another topic but can be handled like this [eclipse-opensovd/opensovd#30](https://github.com/eclipse-opensovd/opensovd/issues/30)
> 
> Would be interesting, if Docker sets the correct headers for the forwarded header approach. Then again, I don't know, if we really target running in a Docker container. The `ecu-sim` in the CDA repository runs in Docker, but it wouldn't really be a problem there, if the Swagger-UI uses the wrong URL...

We have this case in SCORE. If you check https://github.com/eclipse-opensovd/opensovd/issues/108 opensovd-gateway and opensovd-cda are federated. A request to a classic ECU via opensovd-gateway will be routed to opensovd-cda and all href returned by the latter needs to point back to opensovd-gateway. At least thats how it was designed.

mbfm:
@lh-sag This PR has been changed to reflect what we discussed, at least as a first step towards that.

So, @floroks' and my suggestion would be that we merge this PR (after proper review) and then create two issues:
* Support absolute URLs like prescribed in the standard, with proxy headers and to reflect https://github.com/eclipse-opensovd/opensovd/issues/30
* Optionally support relative URLs (could also wait with implementation until the SOVD standard is updated)

floroks (approved):

PR #408, opened 2026-06-30
fix(doip): Detect DoIP version automatically

alexmohr:
## Summary

When creating the UDP socket to send the VIR, the
doip version is now set to 0xFF.
After receiving a VAM from a gateway, the gateway connection uses the DoIP version from the VAM instead of being hard coded to a fixed version instead.
This solves two issues:
1. The protocol version does not have to be configured anymore
2. The CDA is now compatible with networks of mixed DoIP versions.

## Checklist

## Notes for Reviewers

**The base branch for this PR is tests/tester-present (#390) which should be merged first. This is the case because #390 is refactoring a bunch of doip related communication already, using #390 as base branch makes this a lot easier to merge later and prevents duplicating similar work**

### DIFF
--- a/cda-comm-doip/src/connections.rs
+++ b/cda-comm-doip/src/connections.rs
@@ -957,7 +957,7 @@ mod tests {
     ) {
         let (client, server) = tokio::io::duplex(1024);
         let config = DoipSocketConfig {
-            protocol_version: ProtocolVersion::Iso13400_2010,
+            protocol_version: ProtocolVersion::Iso13400_2012,
             send_diagnostic_message_ack: false,
         };
         let server_conn = DoIPConnection::new(server, config);
--- a/cda-comm-doip/src/connections.rs
+++ b/cda-comm-doip/src/connections.rs
@@ -160,6 +160,7 @@ where
                 gateway_ip: discovered_gateway.ip.clone(),
                 name: discovered_gateway.ecu_name.clone(),
                 tester_address: gateway_ecu.tester_address().to_be_bytes(),
+                protocol_version: discovered_gateway.doip_protocol_version,
                 transport: transport.clone(),
             },
             routing_activation_request,
@@ -956,7 +957,7 @@ mod tests {
     ) {
         let (client, server) = tokio::io::duplex(1024);
         let config = DoipSocketConfig {
-            protocol_version: ProtocolVersion::Iso13400_2012,
+            protocol_version: ProtocolVersion::Iso13400_2010,
             send_diagnostic_message_ack: false,
         };
         let server_conn = DoIPConnection::new(server, config);
@@ -968,11 +969,12 @@ mod tests {
                 gateway_ip: "127.0.0.1".to_owned(),
                 name: String::new(),
                 tester_address: [0, 0],
+                protocol_version: ProtocolVersion::Iso13400_2010,
                 transport: DoipTransportConfig {
                     tester_ip: "127.0.0.1".to_owned(),
                     port: 13400,
                     tls_port: 0,
-                    socket: config,
+                    send_diagnostic_message_ack: false,
                     send_timeout: Duration::from_secs(5),
                     alive_check_interval,
                 },
--- a/cda-comm-doip/src/ecu_connection.rs
+++ b/cda-comm-doip/src/ecu_connection.rs
@@ -293,6 +293,13 @@ async fn connect_to_gateway(
     Ok(stream)
 }
 
+fn socket_config(config: &GatewayConnectionConfig) -> DoipSocketConfig {
+    DoipSocketConfig {
+        protocol_version: config.doip.protocol_version,
+        send_diagnostic_message_ack: config.doip.transport.send_diagnostic_message_ack,
+    }
+}
+
 #[tracing::instrument(
     skip_all,
     fields(
@@ -318,13 +325,9 @@ pub(crate) async fn establish_ecu_connection(
     )
     .await
     {
-        Ok(Ok(stream)) => EcuConnectionVariant::Plain(DoIPConnection::new(
-            stream,
-            DoipSocketConfig {
-                protocol_version: config.doip.protocol_version,
-                send_diagnostic_message_ack: config.doip.transport.send_diagnostic_message_ack,
-            },
-        )),
+        Ok(Ok(stream)) => {
+            EcuConnectionVariant::Plain(DoIPConnection::new(stream, socket_config(config)))
+        }
         Ok(Err(e)) => return Err(e),
         Err(_) => {
             return Err(ConnectionError::Timeout(
@@ -413,16 +416,7 @@ pub(crate) async fn establish_tls_ecu_connection(
     )
     .await
     {
-        Ok(Ok(stream)) => {
-            create_tls_stream(
-                stream,
-                DoipSocketConfig {
-                    protocol_version: config.doip.protocol_version,
-                    send_diagnostic_message_ack: config.doip.transport.send_diagnostic_message_ack,
-                },
-            )
-            .await?
-        }
+        Ok(Ok(stream)) => create_tls_stream(stream, socket_config(config)).await?,
         Ok(Err(e)) => {
             return Err(ConnectionError::ConnectionFailed(format!(
                 "Connect failed: {e:?}"
--- a/cda-comm-doip/src/ecu_connection.rs
+++ b/cda-comm-doip/src/ecu_connection.rs
@@ -318,9 +318,13 @@ pub(crate) async fn establish_ecu_connection(
     )
     .await
     {
-        Ok(Ok(stream)) => {
-            EcuConnectionVariant::Plain(DoIPConnection::new(stream, config.doip.transport.socket))
-        }
+        Ok(Ok(stream)) => EcuConnectionVariant::Plain(DoIPConnection::new(
+            stream,
+            DoipSocketConfig {
+                protocol_version: config.doip.protocol_version,
+                send_diagnostic_message_ack: config.doip.transport.send_diagnostic_message_ack,
+            },
+        )),
         Ok(Err(e)) => return Err(e),
         Err(_) => {
             return Err(ConnectionError::Timeout(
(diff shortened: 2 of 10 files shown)


cda-comm-doip/src/connections.rs
```diff
@@ -956,7 +957,7 @@ mod tests {
     ) {
         let (client, server) = tokio::io::duplex(1024);
         let config = DoipSocketConfig {
-            protocol_version: ProtocolVersion::Iso13400_2012,
+            protocol_version: ProtocolVersion::Iso13400_2010,
```
theswiftfox: q: I assume the change here is due to something in the spec, or compatibility reasons? if so, can we add a short docstring why we chose this version here?

cda-comm-doip/src/ecu_connection.rs
```diff
@@ -318,9 +318,13 @@ pub(crate) async fn establish_ecu_connection(
     )
     .await
     {
-        Ok(Ok(stream)) => {
-            EcuConnectionVariant::Plain(DoIPConnection::new(stream, config.doip.transport.socket))
-        }
+        Ok(Ok(stream)) => EcuConnectionVariant::Plain(DoIPConnection::new(
+            stream,
+            DoipSocketConfig {
```
theswiftfox: question: could we create that config outside and pass in the reference? as we are duplicating this in `establish_tls_ecu_connection`

cda-main/src/lib.rs
```diff
@@ -221,6 +222,12 @@ impl From<DoipGatewaySetupError> for AppError {
             }
             DoipGatewaySetupError::ResourceError(_) => Self::ResourceError(value.to_string()),
             DoipGatewaySetupError::ServerError(_) => Self::ServerError(value.to_string()),
+            DoipGatewaySetupError::UnknownECU {
+                logical_address,
+                protocol_version: u8,
```
theswiftfox: suggestion: drop the `: u8` binding and use `protocol_version` in the format string directly.

cda-interfaces/src/lib.rs
```diff
@@ -378,6 +378,16 @@ pub enum DiagServiceError {
 pub enum DoipGatewaySetupError {
     #[error("Invalid address: `{0}`")]
     InvalidAddress(String),
+    #[error("Invalid protocol: `{0}`")]
+    InvalidProtocol(String),
```
theswiftfox: q: this seems not used anywhere right now. is this expected to be used in the future? Or are we missing a check that the protocol the ecu responded with is a valid protocol?

cda-comm-doip/src/vir_vam.rs
```diff
@@ -359,21 +356,25 @@ where
...
                     source_ip = %source_addr.ip(),
                     logical_address = %format!("{:#06x}", logical_address),
+                    protocol_version = ?doip_msg.header.protocol_version,
                     "Matching ECU found"
                 );
                 Ok(Some(DiscoveredGateway {
                     ip: source_addr.ip().to_string(),
                     ecu_name: ecu.clone(),
                     logical_address,
+                    doip_protocol_version: doip_msg.header.protocol_version,
```
theswiftfox, line 366: q: (related to comment on InvalidProtocol error enum) do we need to verify the received protocol version is a valid doip protocol version?

cda-comm-doip/src/lib.rs
```diff
@@ -960,24 +956,26 @@ fn create_netmask(tester_ip: &str, tester_subnet: &str) -> Result<u32, DoipGatew
...
+/// * The underlying system call to create a new socket fails,
+///   resulting in a `DoipGatewaySetupError::SocketCreationFailed`.
+/// * Setting the `SO_REUSEADDR` socket option fails,
+///   resulting in a `DoipGatewaySetupError::InvalidAddress`.
+/// * On Unix-like systems, setting the `SO_REUSEPORT` socket option fails,
+///   resulting in a `DoipGatewaySetupError::PortBindFailed`.
+/// * Setting the `SO_BROADCAST` socket option fails,
+///   resulting in a `DoipGatewaySetupError::SocketCreationFailed`.
+/// * Setting the socket to non-blocking mode fails,
+///   resulting in a `DoipGatewaySetupError::InvalidConfiguration`.
```
theswiftfox: nitpick: i think this part of the docstring can be removed as the code producing the error was removed?

cda-comm-doip/src/lib.rs
```diff
@@ -1030,7 +1028,18 @@ fn create_socket_with_protocol(
     })?;
 
     let std_sock: std::net::UdpSocket = socket.into();
-    DoIPUdpSocket::new(std_sock, protocol_version).map_err(|e| {
+    DoIPUdpSocket::new(
+        std_sock,
+        DoipSocketConfig {
+            // Using ::DefaultValue here, because this socket is used to send the VIR, where
```
theswiftfox: nitpick: maybe `Using ::DefaultValue (0xFF) here, ...`
as per spec we need to send 0xFF iirc which conveniently is accessible via ::DefaultValue

cda-comm-doip/src/lib.rs
```diff
@@ -143,6 +144,9 @@ pub(crate) struct GatewayDoipConfig {
     pub(crate) name: String,
     /// UDS address of the tester.
     pub(crate) tester_address: [u8; 2],
+    /// The `DoIp` protocol version to use for this gateway connection.
+    /// Setup from the value in the VAM.
```
theswiftfox: nitpick: `Set from the protocol version field in the VAM`

cda-comm-doip/src/socket.rs
```diff
@@ -87,19 +87,19 @@ pub(crate) struct DoIPConnectionReadHalf<T: AsyncRead + Unpin> {
 }
 pub(crate) struct DoIPConnectionWriteHalf<T: AsyncWrite + Unpin> {
     io: FramedWrite<WriteHalf<T>, DoipCodec>,
-    protocol_version: ProtocolVersion,
+    config: DoipSocketConfig,
 }
 
 impl<T: AsyncWrite + Unpin> DoIPConnectionWriteHalf<T> {
-    pub fn new(io: WriteHalf<T>, protocol_version: ProtocolVersion) -> Self {
+    pub fn new(io: WriteHalf<T>, config: DoipSocketConfig) -> Self {
```
theswiftfox: q: i don't fully understand this change. why is the full config stored now when we still only access the protocol_version field of it?

theswiftfox (commented):
just a few findings

cda-comm-doip/src/vir_vam.rs
alexmohr: The doip library is already doing this when constructing the doip header, so it's not handled again here.

```rust
impl TryFrom<[u8; 8]> for DoipHeader {
    type Error = Error;

    fn try_from(value: [u8; 8]) -> Result<Self> {
        let protocol_version_slice = value.first().ok_or(Error::OutOfBounds {
            source: "DoIP Header",
            variable: "Protocol Version",
        })?;
        let protocol_version = ProtocolVersion::try_from(protocol_version_slice)?;
```

```rust
impl TryFrom<&u8> for ProtocolVersion {
    type Error = Error;

    fn try_from(value: &u8) -> Result<Self, Self::Error> {
        let val = *value;

        match val {
            v if v == ProtocolVersion::ReservedVer as u8 => Ok(ProtocolVersion::ReservedVer),
            v if v == ProtocolVersion::Iso13400_2010 as u8 => Ok(ProtocolVersion::Iso13400_2010),
            v if v == ProtocolVersion::Iso13400_2012 as u8 => Ok(ProtocolVersion::Iso13400_2012),
            v if v == ProtocolVersion::Iso13400_2019 as u8 => Ok(ProtocolVersion::Iso13400_2019),
            v if v == ProtocolVersion::Iso13400_2019Amd1 as u8 => {
                Ok(ProtocolVersion::Iso13400_2019Amd1)
            }
            v if v == ProtocolVersion::DefaultValue as u8 => Ok(ProtocolVersion::DefaultValue),
            v => Err(Error::InvalidProtocolVersion { value: v }),
        }
    }
}
```

cda-interfaces/src/lib.rs
alexmohr: forgot to remove this after making sure that the doip library is already handling this.

cda-comm-doip/src/vir_vam.rs
theswiftfox: Ah okay. then we can probably remove the unused error enum?
(and maybe a docsting at this place)

cda-comm-doip/src/socket.rs
alexmohr: Not needed, reverted.

cda-main/src/lib.rs
alexmohr: done

cda-comm-doip/src/ecu_connection.rs
alexmohr: Done

cda-comm-doip/src/lib.rs
alexmohr: Removed

cda-comm-doip/src/lib.rs
alexmohr: Done

cda-comm-doip/src/lib.rs
alexmohr: Done

cda-comm-doip/src/vir_vam.rs
alexmohr: removed the enum and adjusted the docstring

cda-comm-doip/src/connections.rs
alexmohr: not needed in this test, reverted it.

cda-comm-doip/src/socket.rs
theswiftfox: this is a super nitpick, but why don't we just pass in the protocol version, but instead pass the config by value still?

cda-comm-doip/src/socket.rs
```diff
@@ -115,18 +115,18 @@ impl<T: AsyncRead + Unpin> DoIPConnectionReadHalf<T> {
...
     io: UdpFramed<DoipCodec, tokio::net::UdpSocket>,
-    protocol_version: ProtocolVersion,
+    config: DoipSocketConfig,
 }
 
 impl DoIPUdpSocket {
     pub(crate) fn new(
         socket: std::net::UdpSocket,
-        protocol_version: ProtocolVersion,
+        config: DoipSocketConfig,
```
theswiftfox: same nitpick here as with write half, wouldnt passing in the protocol_version be enough?

cda-interfaces/src/lib.rs
theswiftfox: maybe i'm being stupid here, but then we can delete this enum variant right? or am I missing something?

alexmohr: Seems like I am stupid because I thought I removed this already ... thanks for double checking it's gone now. (for real this time 😆 )

cda-comm-doip/src/socket.rs
alexmohr: fixed :)

cda-comm-doip/src/socket.rs
alexmohr: fixed :)

theswiftfox (approved):

PR #411, opened 2026-07-01
refactor: improve EcuState handling

theswiftfox:
## Summary

This PR aims to improve the handling of EcuState and decouple it from the ecumanager itself.
The core changes are:
 
#### Single flat EcuState enum (Online, Offline, NotTested, Duplicate, Disconnected, NoVariantDetected) decoupled into 
- Connectivity — Online | Offline (transport reachability)
- VariantState — NotTested | Detected{name,is_base,is_fallback} | Duplicate | NotDetected

-> On disconnect, only Connectivity flips to Offline. Variant identity is preserved — enabling reconnect without re-running detection from scratch.

#### Variant Detection is now triggered when trying to send data to ECU that had no variant detection running yet
Before every UDS send, transport.rs checks if variant is NotDetected or ECU is Offline+Detected. If so, and ECU is reachable, triggers synchronous detect_variant first.

## Checklist

## Notes for Reviewers

---

### DIFF
--- a/cda-comm-uds/src/coordinator.rs
+++ b/cda-comm-uds/src/coordinator.rs
@@ -248,14 +248,14 @@ impl Message<EcuConnected> for EcuCoordinator {
 ///
 /// Uses `ask` (request-response) to guarantee the suppression is active
 /// before variant detection sends begin.
-pub struct PrepareVariantDetection;
+pub struct SuppressDisconnectHandling;
 
-impl Message<PrepareVariantDetection> for EcuCoordinator {
+impl Message<SuppressDisconnectHandling> for EcuCoordinator {
     type Reply = ();
 
     async fn handle(
         &mut self,
-        _msg: PrepareVariantDetection,
+        _msg: SuppressDisconnectHandling,
         _ctx: &mut Context<Self, Self::Reply>,
     ) -> Self::Reply {
         tracing::debug!(
@@ -267,14 +267,14 @@ impl Message<PrepareVariantDetection> for EcuCoordinator {
 }
 
 /// Re-enable disconnect events after variant detection completes.
-pub struct FinishVariantDetection;
+pub struct RestoreDisconnectHandling;
 
-impl Message<FinishVariantDetection> for EcuCoordinator {
+impl Message<RestoreDisconnectHandling> for EcuCoordinator {
     type Reply = ();
 
     async fn handle(
         &mut self,
-        _msg: FinishVariantDetection,
+        _msg: RestoreDisconnectHandling,
         _ctx: &mut Context<Self, Self::Reply>,
     ) -> Self::Reply {
         tracing::debug!(
@@ -282,27 +282,6 @@ impl Message<FinishVariantDetection> for EcuCoordinator {
             "Variant detection finished. Re-enabling disconnect events"
         );
         self.suppress_disconnect = false;
-    }
-}
-
-/// Clear the variant to trigger re-detection on the next UDS send.
-pub struct ClearVariantForRedetect;
-
-impl Message<ClearVariantForRedetect> for EcuCoordinator {
-    type Reply = ();
-
-    async fn handle(
-        &mut self,
-        _msg: ClearVariantForRedetect,
-        _ctx: &mut Context<Self, Self::Reply>,
-    ) -> Self::Reply {
-        tracing::info!(
-            ecu = %self.ecu_name,
-            "Clearing variant for re-detection"
-        );
-        let mut ecu_state = std_ext::lock_write(&self.state.ecu_state);
-        ecu_state.variant_state = VariantState::NotTested;
-        ecu_state.variant_index = None;
     }
 }
 
@@ -514,34 +493,6 @@ mod tests {
             }
         );
         assert_eq!(status.variant_index, Some(3));
-    }
-
-    #[tokio::test]
-    async fn clear_variant_for_redetect_sets_not_tested() {
-        let handle = spawn_test_coordinator("TestECU");
-        {
-            let mut ecu_state = handle.state.ecu_state.write().unwrap();
-            ecu_state.connectivity = Connectivity::Online;
-            ecu_state.variant_state = VariantState::Detected {
-                name: "V1".to_owned(),
-                is_base_variant: true,
-                is_fallback: false,
-            };
-            ecu_state.variant_index = Some(1);
-        }
-
-        handle
-            .actor_ref
-            .tell(ClearVariantForRedetect)
-            .await
-            .expect("Actor should be alive");
-        tokio::task::yield_now().await;
-
-        let status = handle.ecu_status();
-        assert_eq!(status.variant_state, VariantState::NotTested);
-        assert_eq!(status.variant_index, None);
-        // Connectivity preserved
-        assert_eq!(status.connectivity, Connectivity::Online);
     }
 
     #[tokio::test]
--- a/cda-comm-uds/src/coordinator.rs
+++ b/cda-comm-uds/src/coordinator.rs
@@ -1,0 +1,631 @@
+/*
+ * SPDX-FileCopyrightText: 2026 Copyright (c) Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Per-ECU state coordinator implemented as a kameo actor.
+//!
+//! The [`EcuCoordinator`] actor serializes connectivity state mutations for a single ECU.
+//! Variant detection writes happen directly via `std::sync::RwLock` (from `EcuManager`),
+//! while disconnect events go through the actor to avoid races.
+//!
+//! Reads go directly through [`EcuCoordinatorHandle::state`] (a `std::sync::RwLock`)
+//! without touching the actor mailbox.
+
+use cda_interfaces::{
+    Connectivity, EcuRuntimeState, EcuState, VariantState, dlt_ctx, util::std_ext,
+};
+use kameo::{
+    Actor,
+    actor::{ActorRef, Spawn},
+    message::{Context, Message},
+};
+
+/// The kameo actor that serializes connectivity state mutations for one ECU.
+#[derive(Actor)]
+pub struct EcuCoordinator {
+    /// Shared runtime state, also held by [`EcuCoordinatorHandle`] for direct synchronous reads.
+    state: EcuRuntimeState,
+    ecu_name: String,
+    /// When `true`, `EcuDisconnected` messages are suppressed.
+    /// Set during variant detection to prevent timeout-triggered disconnects
+    /// from racing with the detection result.
+    suppress_disconnect: bool,
+}
+
+/// Cloneable handle providing read access and actor messaging for one ECU.
+///
+/// Distributed to all components that need to read or mutate ECU runtime state.
+/// Direct reads via [`Self::state`] bypass the mailbox; mutations go through
+/// [`Self::actor_ref`].
+#[derive(Clone)]
+pub struct EcuCoordinatorHandle {
+    /// Actor reference for sending mutation messages.
+    pub actor_ref: ActorRef<EcuCoordinator>,
+    /// Shared state for direct synchronous reads. The actor is the sole writer
+    /// for connectivity; variant detection writes directly.
+    pub state: EcuRuntimeState,
+}
+
+impl EcuCoordinatorHandle {
+    /// Spawn a new coordinator actor for the given ECU.
+    ///
+    /// Returns a handle that can be cloned and distributed to all consumers.
+    #[must_use]
+    pub fn spawn(ecu_name: String) -> Self {
+        let state = EcuRuntimeState::new();
+        Self::spawn_with_state(ecu_name, state)
+    }
+
+    /// Spawn a coordinator actor sharing an existing [`EcuRuntimeState`].
+    ///
+    /// Use this when the `EcuManager` already owns the `EcuRuntimeState`
+    /// and the coordinator should operate on the same instance.
+    #[must_use]
+    pub fn spawn_with_state(ecu_name: String, state: EcuRuntimeState) -> Self {
+        let actor_ref = EcuCoordinator::spawn(EcuCoordinator {
+            state: state.clone(),
+            ecu_name,
+            suppress_disconnect: false,
+        });
+        Self { actor_ref, state }
+    }
+
+    /// Read the current ECU status (synchronous, no mailbox round-trip).
+    #[must_use]
+    pub fn ecu_status(&self) -> EcuState {
+        self.state.status()
+    }
+
+    /// Read the current connectivity (synchronous).
+    #[must_use]
+    pub fn connectivity(&self) -> Connectivity {
+        std_ext::lock_read(&self.state.ecu_state).connectivity
+    }
+
+    /// Read a service state by SID (synchronous).
+    #[must_use]
+    pub fn get_service_state(&self, sid: u8) -> Option<String> {
+        std_ext::lock_read(&self.state.service_states)
+            .get(&sid)
+            .cloned()
+    }
+}
+
+/// Set a single service state entry (e.g. session or security level after UDS response).
+pub struct SetServiceState {
+    /// UDS service identifier (SID).
+    pub sid: u8,
+    /// New state value (e.g. "extended", "`level_42`").
+    pub value: String,
+}
+
+impl Message<SetServiceState> for EcuCoordinator {
+    type Reply = ();
+
+    async fn handle(
+        &mut self,
+        msg: SetServiceState,
+        _ctx: &mut Context<Self, Self::Reply>,
+    ) -> Self::Reply {
+        tracing::debug!(
+            ecu = %self.ecu_name,
+            sid = msg.sid,
+            value = %msg.value,
+            "Setting service state"
+        );
+        std_ext::lock_write(&self.state.service_states).insert(msg.sid, msg.value);
+    }
+}
+
+/// Set default service states (session, security, DTC, comm control) if not already present.
+pub struct SetDefaultStates {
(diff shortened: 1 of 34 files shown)


cda-comm-uds/src/coordinator.rs
```diff
@@ -0,0 +1,535 @@
...
+
+    /// Read the current ECU status (synchronous, no mailbox round-trip).
+    #[must_use]
+    pub fn ecu_status(&self) -> EcuState {
+        self.state.status()
+    }
+
+    /// Read the current connectivity (synchronous).
+    #[must_use]
+    pub fn connectivity(&self) -> Connectivity {
```
alexmohr, line 90: question: Why is this not using tokio locks and async instead of this being synchronous and blocking the task calling the function? Basically same question for all functions that use lock_write

cda-comm-uds/src/tester_present.rs
```diff
@@ -75,19 +75,6 @@ impl<S: EcuGateway, T: EcuManager> UdsManager<S, T> {
                         schedule.set_missed_tick_behavior(MissedTickBehavior::Delay);
                         loop {
                             let _ = schedule.tick().await;
-                            // Skip sending if the ECU is not online; the loop will
```
theswiftfox, line 78: fixme: restore this but use connection state instead of variant state.

cda-comm-uds/src/transport.rs
```diff
@@ -43,15 +44,66 @@ impl<S: EcuGateway, T: EcuManager> UdsManager<S, T> {
...
+                    tracing::warn!(
+                        ecu_name,
+                        error = %e,
+                        "Pre-send variant detection failed"
+                    );
+                }
+            }
+        }
+
+        self.send_without_variant_guard(
```
alexmohr, line 79: question: If the ecu is online but needs detection we try variant detection first, if the ecu is offline we don't care? Why do we skip variant detection with an offline ecu instead of sending it as well and trying to update our internal state, as we send the uds frame in any case too.

cda-comm-uds/src/transport.rs
```diff
@@ -43,15 +44,66 @@ impl<S: EcuGateway, T: EcuManager> UdsManager<S, T> {
...
+        // Pre-send: if variant is not set or ECU was disconnected, attempt synchronous
+        // detection before proceeding. This handles the case where the ECU was previously
+        // unreachable or had its variant cleared after a session change.
+        {
+            let status = ecu.read().await.ecu_status();
+            let needs_detection = matches!(status.variant_state, VariantState::NotDetected)
+                || (status.connectivity == Connectivity::Offline
+                    && matches!(
+                        status.variant_state,
+                        VariantState::Detected { .. } | VariantState::NotDetected
```
alexmohr: question: Why is 'NotTested' a valid case for need_detection too?
I assume NotDetected is not a case because we don't want to retry with every frame we send on an undetectedable ecu?

cda-core/src/diag_kernel/ecumanager.rs
```diff
@@ -676,28 +666,30 @@ impl<S: SecurityPlugin> EcuManager<S> {
...
-        self.variant = EcuVariant {
-            name: Some(variant.name.clone()),
-            is_base_variant: variant.is_base_variant,
-            is_fallback: variant.is_fallback,
-            state,
-            logical_address: self.logical_address,
-        };
+        tracing::debug!("Setting variant to '{variant_name}'");
+        {
+            let mut es = std_ext::lock_write(&self.runtime_state.ecu_state);
```
alexmohr: nitpick: I would prefer ecu_state as variable name, but it's consistently 'es' in the PR.

cda-core/src/diag_kernel/security.rs
```diff
@@ -175,17 +177,17 @@ impl<S: SecurityPlugin> EcuManager<S> {
 
         // Get current ECU states
         let (ecu_session, ecu_security_level) = {
-            let ecu_states = self.ecu_service_states.read().await;
+            let ss = std_ext::lock_read(&self.runtime_state.service_states);
```
alexmohr: nitpick: personal preference here, but imho service_states would be the better named compared to `ss` (but at least it is consistent through this PR)

alexmohr (commented):
Mostly questions and nitpick, great change, makes handling the state much better. 
One nitpick I didn't have a place for: there is a typo in the commit message: ref_r_actor instead of refactor

cda-comm-uds/src/coordinator.rs
theswiftfox: mostly because it is only a short read / write of the state and not intended to be held for a longer time. Using sync locks made it a bit simpler than putting async everywhere, though after the last commit that would have gotten simpler again, as i could remove the need for reading the lock in the constructor.

I'm not specifically attached to using the sync lock, can also use async if you prefer that. :)

cda-comm-uds/src/transport.rs
theswiftfox: Oh I think on line it 55 should be NotTested, instead of NotDetected. Thanks!
Then it should make more sense.
NotTested -> Always try to detect once. Will transition into Detected or NotDetected or Duplicate

cda-comm-uds/src/transport.rs
theswiftfox: I think this is due to ecu_online being a bit missleading. If I'm not reading that code completely wrong it only returns if the ECU is 'reachable' under the current gateway from a logical address standpoint, not if the ECU is actually online 🥴
So it should be sent in both cases if the ecu belongs to this gateway. Though this makes me think the whole check is a bit stupid and we could just always send without the if ecu_online ... 😅

Ideally the ecu_online would probe via a placeholder uds request and then let us know if its actually online or something.

Should i just remove the clause ...ecu_online()... from the if?

cda-comm-uds/src/coordinator.rs
alexmohr: Should be fine, but can we document this once? maybe at the lock definition, so the next person reading this isn't asking the same question again :)

cda-comm-uds/src/transport.rs
alexmohr: Yes, having the ecu in a needs detection state should be sufficient to try to detect its variant

cda-comm-uds/src/transport.rs
alexmohr: Sounds good 👍

theswiftfox:
Rebased and ready to review again now.

cda-comm-uds/src/coordinator.rs
theswiftfox: added a doc comment on the lock

cda-comm-uds/src/transport.rs
theswiftfox: changed. now the NotTested is the trigger for automatic re-detection before a request. when an ecu reconnects after being disconnected before it gets put to NotTested as we will have to recheck the variant in that case (ECU might have rebooted into bootloader for example)

cda-core/src/diag_kernel/ecumanager.rs
theswiftfox: i changed all occurances (at least i hope i got all ^^)

cda-core/src/diag_kernel/security.rs
theswiftfox: Same, renamed all as per suggetion

cda-interfaces/src/ecumanager.rs
```diff
@@ -617,13 +765,27 @@ pub trait VariantDetection: Send + Sync + 'static {
...
+
+    /// Clear the current variant to trigger re-detection on the next send.
+    /// Sets variant state to [`VariantState::NotDetected`] and clears the variant name.
+    fn clear_variant_for_redetect(&mut self);
+}
+
+/// Callback interface for transport-level ECU connectivity changes.
+#[async_trait]
+pub trait EcuConnectivityHandler: Send + Sync + 'static {
+    /// Called when a connection is established or re-established for an ECU.
```
alexmohr: nitpick: The docstring is confusing, " re-established for an ECU." but then takes an array of ECUs. It would be helpful if the docstring would state that this is updating all ECUs behind (and including) the gateway and if the function name could reflect this too.

cda-interfaces/src/ecuuds.rs
```diff
@@ -574,16 +574,20 @@ pub trait UdsVariant {
     /// does not exist or no service for variant detection is available.
     async fn detect_variant(&self, ecu_name: &str) -> Result<(), DiagServiceError>;
 
-    /// Get the name of the variant for the given ECU.
+    /// Get the ECU status (connectivity + variant state) for the given ECU.
     /// # Errors
     /// Will return Err if the ECU does not exist.
-    /// If the variant is cannot be resolved, "Unknown" will be returned.
-    async fn get_variant(&self, ecu_name: &str) -> Result<EcuVariant, DiagServiceError>;
+    async fn get_ecu_status(&self, ecu_name: &str) -> Result<EcuState, DiagServiceError>;
```
alexmohr: nitpick: can we match the naming between the return type EcuState and the function name ecu_status?

cda-interfaces/src/util.rs
```diff
@@ -170,6 +170,30 @@ pub mod serde_ext {
...
 }
 
+pub mod std_ext {
+    #[inline]
+    pub fn lock_read<T>(lock: &std::sync::RwLock<T>) -> std::sync::RwLockReadGuard<'_, T> {
+        match lock.read() {
+            Ok(guard) => guard,
+            Err(poisoned) => {
+                tracing::warn!("RwLock poisoned, proceeding with poisoned read lock");
+                poisoned.into_inner()
```
alexmohr, line 181: suggestion: Once a lock is poisoned this state is never cleared automatically, which means we would log this warning for the rest of the application lifetime. As we chose to continue, we probably should clear the poisoned state either here or after we wrote the state again. 

Same for `lock_write`

See https://doc.rust-lang.org/std/sync/struct.RwLock.html#method.clear_poison

cda-core/src/diag_kernel/test_utils/ecu_manager_builder.rs
```diff
@@ -51,14 +52,16 @@ pub(crate) fn new_ecu_manager(
...
-    manager.variant = EcuVariant {
-        name: Some(TEST_DIAG_LAYER.to_owned()),
-        is_base_variant: true,
-        is_fallback: false,
-        state: EcuState::Online,
-        logical_address: 0,
-    };
-    manager.variant_index = Some(0);
+    {
+        let mut es = std_ext::lock_write(&manager.runtime_state.ecu_state);
```
alexmohr: nitpick: one ecu_state that is still named `es`

cda-core/src/diag_kernel/ecumanager.rs
alexmohr: There are a few left 

```
grep -rnE 'let (mut )?es = ' cda-*
cda-comm-uds/src/state_coordinator.rs:137:            let mut es = runtime_state.ecu_state.write().unwrap();
cda-core/src/diag_kernel/test_utils/ecu_manager_builder.rs:56:        let mut es = std_ext::lock_write(&manager.runtime_state.ecu_state);
cda-core/src/diag_kernel/variant.rs:60:                    let mut es = std_ext::lock_write(&self.runtime_state.ecu_state);
cda-core/src/diag_kernel/variant.rs:77:                        let mut es = std_ext::lock_write(&self.runtime_state.ecu_state);
cda-core/src/diag_kernel/variant.rs:101:        let mut es = std_ext::lock_write(&self.runtime_state.ecu_state);
cda-core/src/diag_kernel/variant.rs:110:        let mut es = std_ext::lock_write(&self.runtime_state.ecu_state);
cda-core/src/diag_kernel/variant.rs:123:        let mut es = std_ext::lock_write(&self.runtime_state.ecu_state);
```

cda-core/src/diag_kernel/security.rs
alexmohr: two left:

```
 grep -rnE 'let (mut )?ss = ' cda-*
cda-core/src/diag_kernel/state.rs:85:        let mut ss = std_ext::lock_write(&self.runtime_state.service_states);
cda-core/src/diag_kernel/state.rs:173:        let ss = std_ext::lock_read(&self.runtime_state.service_states);
```

alexmohr (commented):
Mostly nitpick with one real suggestion :)

cda-interfaces/src/ecuuds.rs
theswiftfox: oh sure. thanks :)

cda-comm-uds/src/transport.rs
```diff
@@ -300,6 +346,17 @@ impl<S: EcuGateway, T: UdsEcuDb> UdsManager<S, T> {
         drop(response_rx);
         drop(ecu_sem);
 
+        // Post-send: if a service send (not tester present) timed out,
+        // the ECU is unreachable - notify the coordinator.
+        // The coordinator will suppress this if variant detection is in progress.
+        if matches!(response, Err(DiagServiceError::Timeout))
+            && sent_sid != service_ids::TESTER_PRESENT
+        {
+            self.state_coordinator
```
alexmohr, line 364: question: Do we intend to have an automatic recovery mechanism here or is it enough to require a variant detection to get an ecu back online?

cda-comm-uds/src/coordinator.rs
```diff
@@ -0,0 +1,631 @@
...
+            sid = msg.sid,
+            value = %msg.value,
+            "Setting service state"
+        );
+        std_ext::lock_write(&self.state.service_states).insert(msg.sid, msg.value);
+    }
+}
+
+/// Set default service states (session, security, DTC, comm control) if not already present.
+pub struct SetDefaultStates {
```
alexmohr, line 130: suggestion: A bunch of these states / messages are never sent and not used so far outside of unit tests, it probably makes sense to have a follow up ticket to close this gap or remove the unused messages.

 SetServiceState, SetDefaultStates, ClearServiceStates, MarkAsDuplicate, MarkAsNoVariantDetected, ClearVariantForRedetect, plus handle readers ecu_status(), connectivity(), get_service_state(), get_handle().

Follow up ticket could describe that these things must go through the coordinator too instead of bypassing the actor?

cda-core/src/diag_kernel/variant.rs
```diff
@@ -115,12 +98,30 @@ impl<S: SecurityPlugin> VariantDetection for EcuManager<S> {
...
-        self.variant.state = EcuState::NoVariantDetected;
+        let mut es = std_ext::lock_write(&self.runtime_state.ecu_state);
+        es.variant_state = VariantState::NotDetected;
+        es.variant_index = None;
+        drop(es);
+        self.db_cache.reset_sync();
         self.diag_database.unload();
     }
+
+    fn clear_variant_for_redetect(&mut self) {
```
alexmohr: todo: This is never called.

cda-comm-uds/src/transport.rs
theswiftfox: oh crap i think we might have a gap here for ecus that are behind a gateway.
for gateway ecus the recovery is that on reconnect we set them back to online, but this would not happen for ecus that are behind a gateway as they do not trigger a reconnect.
We probably need to use a manual variant detection for those as recovery, but I have to double check if that works. And ideally I'd like a cleaner solution that works similar to the doip.

cda-comm-uds/src/variant.rs
```diff
@@ -111,6 +114,46 @@ impl<S: EcuGateway, T: EcuManager> UdsVariant for UdsManager<S, T> {
                 service_responses.insert(name, response);
             }
         }
+        self.state_coordinator
```
alexmohr, line 120: suggestion: Only set the the finish_variant detection after the variants are fully updated., because `detect_variant` is called again within this block.

cda-comm-uds/src/variant.rs
```diff
@@ -111,6 +114,46 @@ impl<S: EcuGateway, T: EcuManager> UdsVariant for UdsManager<S, T> {
...
+            if let Some(duplicates) = ecu
+                .read()
+                .await
+                .duplicating_ecu_names()
+                .cloned()
+                .filter(|d| !d.is_empty())
+            {
+                for dup_name in &duplicates {
+                    if let Some(dup_ecu) = self.ecus.get(dup_name) {
+                        let _ = dup_ecu
```
alexmohr, line 149: question: why is this not going through the ecu coordinator and using MarkAsDuplicate?

cda-comm-uds/src/variant.rs
theswiftfox: the finish_variant_detection only takes care of making the state_coordinator not triggering the ecu offline path in the uds send as this would cause a looping effect of retriggering variant detections.
I can maybe rename it to `suppress_ecu_timeout_handling_on_send` and `reenable_ecu_timeout_handling_on_send` to make this more clear.

cda-comm-uds/src/variant.rs
theswiftfox: this is only for the case that we did not get any responses for the ecu.
if we never reached any of the duplicates we do not know which the 'real' one is and which the duplicates are, so i think setting all to offline is correct here. but let me know if you disagree :)

cda-comm-uds/src/coordinator.rs
theswiftfox: Yes, I didn't want to blow up this PR more by moving all those already. Follow up ticket sounds good. I'll create one.

cda-core/src/diag_kernel/ecumanager.rs
theswiftfox: I think now i got them all! Somehow i managed to restore a few when rebasing I think. because I'm pretty sure i edited those in variant.rs before.. xD or i forgot to commit them xD

cda-interfaces/src/util.rs
theswiftfox: I would for now simply clear them and try to continue.

cda-comm-uds/src/coordinator.rs
alexmohr: For sake of completeness: https://github.com/eclipse-opensovd/classic-diagnostic-adapter/issues/424

cda-comm-uds/src/variant.rs
alexmohr: oh, ya makes sense, thx.

alexmohr (dismissed):

alexmohr (approved):

PR #413, opened 2026-07-02
Configurable Semantics and Hex config values

alexmohr:
## Summary
[feat(config): make semantics configurable](https://github.com/eclipse-opensovd/classic-diagnostic-adapter/commit/8481f4263e9fc04201c94c75cca9056fc8a69cf1)

Some OEM deviate from the semantics values the standard
suggests, so making the configurable makes the CDA usable
for these use cases too.

[feat(config): update the generate command write hex](https://github.com/eclipse-opensovd/classic-diagnostic-adapter/commit/cf35e06c774529905181a51b561e1843c03c695a)

The configuration generation command is now writing
selected values as hex.
* Makes it clear that hex is supported
* Makes configuration simpler as values do not have to converted to
  decimal

## Checklist


cda-interfaces/src/datatypes/database_naming_convention.rs
```diff
@@ -61,12 +62,40 @@ pub struct DatabaseNamingConvention {
...
+    pub semantics: Semantics,
+}
+
+/// Defines the name for the database semantics. Although they should follow the labels
+/// defined in the standards, in practice some OEM deviate from this.
+#[derive(Deserialize, Serialize, Clone, Debug, schemars::JsonSchema)]
+pub struct Semantics {
+    /// Semantic for `data` services
+    pub data: String,
+    /// Semantic for `service_id_rq` services
```
theswiftfox: q: maybe i'm missing something here but isn't SERVICEIDRQ the semantic of the parameter containing the serviceid in the request? not the id of a service itself?
edit: same for data

alexmohr: Fair point, given the docstrings here I could have left them out entirely 🤦  Added proper docstrings instead.

theswiftfox (approved):

PR #323, opened 2026-05-04
feat: add CAN bus transport and multi-transport gateway

MarkSchmitt:
Adds CAN (ISO-TP / ISO 15765-2) as a second diagnostic transport next to DoIP, behind the optional `can` cargo feature. Closes #195, closes #415.

## Structure

6 commits, each compiling on its own: four standalone diag-kernel/DoIP fixes flushed out during CAN bring-up, one squashed feature commit carrying the entire CAN transport (gateway, MDD-based addressing, duplicate groups, rediscovery, keep-alive, containerized CAN test environment), and the coverage CI wiring.

## Features

- Three interchangeable ISO-TP backends, selected by the interface string: kernel `can-isotp` sockets (`can0`), userspace ISO-TP over CAN_RAW (`rawcan:vcan0`, no `can_isotp` module needed), and CAN frames via a socketcand daemon (`socketcand:host:port:bus`, works off-Linux).
- **CAN addressing from the MDD com-params** (#415): per-ECU `CP_UniqueRespIdTable` entries first, scalar `CP_CanPhysReqId`/`CP_CanRespUSDTId` second; `[[can.ecu_mappings]]` and com-param config overrides win. ECUs sharing a derived ID pair form a duplicate group (candidate models of one physical node), resolved by variant detection like DoIP address twins.
- 11-bit standard and 29-bit extended CAN IDs (ISO 15765-4 normal fixed addressing), validated once via a shared `CanId` type.
- ECU discovery by probing (TesterPresent by default, configurable probe sequence incl. opt-out via `default_probes`), background rediscovery of absent/slept ECUs, fail-fast transport init.
- Functional-broadcast TesterPresent keep-alive with configurable interval; `keepalive_interval_ms = 0` disables it for resident/on-board deployments so network management can put ECUs to sleep - sleeping ECUs are then recovered autonomously by rediscovery on wake.
- `MultiTransportGateway`: explicit per-ECU pins via `[[can.transport_overrides]]` win; otherwise sticky first-detection binding (DoIP preferred). DoIP-only, CAN-only (`doip.enabled = false`, no DoIP socket bound), and mixed operation.
- KWP2000 `READ_DATA_BY_LOCAL_IDENTIFIER` (0x21) alongside UDS 0x22.

## Testing

- Three CI integration jobs (DoIP, pure CAN, mixed DoIP+CAN), all coverage-instrumented and merged into one report. All green.
[...]

PR #414, opened 2026-07-05
fix(diag-kernel): type compu-method results by the declared type

MarkSchmitt:
A DOP's DIAG-CODED-TYPE describes the raw bytes on the wire; the value produced by its compu method is described by the PHYSICAL-TYPE. The decode path dropped the physical type entirely and typed conversion results with the coded type, so every linear/scale-linear parameter with a fractional factor lost its fraction: a battery SOC coded as 16-bit unsigned with factor 1/16 and a float physical type decoded to 56 instead of 56.8125. The encode path already used the physical type; decode now carries it through DiagDataTypeContainerRaw and uses it to type linear and scale-linear results, falling back to the coded type when the MDD declares no physical type - databases that genuinely declare integer physical values keep integer results.

The JSON schema generation had the same flaw and reported such parameters as "integer"; it now derives the schema type from the physical type where declared.

Verified against a real BMS over CAN: SOC_Value now returns 56.8125 where a reference implementation reading the same DID computes 56.81 (raw 909 / 16); previously CDA returned 56.

## Summary

I was surprised to see the State of Charge returned by the CDA on my Smart EQ to truncate to integers while my handcrafted python script returned higher resolution. This is a change stemming from PR #323 but since it's not directly related to canbus, i've extracted it to here.

## Checklist

- [x] I have added or updated documentation (there isn't really any documentation as this is simply a bug)
<-- this could been done, but it's also not smth that _has_ been tested before properly.

## Related
I don't think there is any open issue

### DIFF
--- a/cda-core/src/diag_kernel/operations.rs
+++ b/cda-core/src/diag_kernel/operations.rs
@@ -31,6 +31,7 @@ use crate::diag_kernel::{
 
 pub(in crate::diag_kernel) fn uds_data_to_serializable(
     diag_type: datatypes::DataType,
+    physical_type: Option<datatypes::DataType>,
     compu_method: Option<&datatypes::CompuMethod>,
     is_negative_response: bool,
     data: &[u8],
@@ -47,6 +48,7 @@ pub(in crate::diag_kernel) fn uds_data_to_serializable(
                 category => {
                     return compu_lookup(
                         diag_type,
+                        physical_type,
                         compu_method,
                         category,
                         is_negative_response,
@@ -131,6 +133,7 @@ fn apply_linear_conversion(
 
 fn compu_lookup(
     diag_type: DataType,
+    physical_type: Option<DataType>,
     compu_method: &datatypes::CompuMethod,
     category: datatypes::CompuCategory,
     is_negative_response: bool,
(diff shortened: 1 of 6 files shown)


cda-core/src/diag_kernel/operations.rs
```diff
@@ -1729,12 +1742,71 @@ mod tests {
...
+            Some(&compu_method),
+            false,
+            &data,
+        );
         // Should fail because value doesn't match any scale (no scales exist)
         assert!(result.is_err());
     }
 
+    #[test]
+    #[allow(clippy::float_cmp)]
```
alexmohr, line 1757: Todo: as per code style, a reason should be given when disabling clippy warnings

floroks: @alexmohr can we enforce that through checks in the pipeline?

alexmohr: we can and I want to eventually it's the clippy lint https://rust-lang.github.io/rust-clippy/master/index.html#allow_attributes_without_reason

But I didn't have the time yet to enable it and fix the existing lints. Should be fairly easy though.

PR #425, opened 2026-07-08
fix(update-plugin): make the update plugin actually plugable.

alexmohr:
## Summary
Prior to this commit the update plugin was not plugable without rewriting the app init logic, as setup_vehicle_and_routes hard coded the default plugin.

## Checklist

## Notes for Reviewers


cda-interfaces/src/ecugateway.rs
```diff
@@ -127,3 +127,19 @@ pub trait EcuGateway: Clone + Send + Sync + 'static {
...
+/// during a database reload.
+///
+/// The associated type `UdpSocket` is deliberately opaque in `cda-interfaces` so that this
+/// crate stays free of any dependency on `cda-comm-doip`.
+pub trait EcuGatewaySockets {
+    /// Opaque socket handle type (e.g. `cda_comm_doip::socket::DoIPUdpSocket`).
+    type Socket: Send + Sync + 'static;
+
+    /// Returns a shared, cloneable handle to the underlying UDP socket.
+    fn upd_socket(&self) -> Arc<Mutex<Self::Socket>>;
```
theswiftfox: nit: `udp_` or `upd_`? sidenote: is it important if this is a tcp or udp socket for the interface? or should this rather just be `socket` or similar?

cda-interfaces/src/lib.rs
```diff
@@ -13,9 +13,13 @@
 
 use std::{
     fmt::{Display, Formatter},
+    future::Future,
```
theswiftfox: nit: shouldn't this be in prelude in edition 2024?

cda-interfaces/src/lib.rs
```diff
@@ -426,3 +431,26 @@ impl Display for DiagCommAction {
         }
     }
 }
+
+/// Type alias for the boxed shared shutdown signal.
+/// This provides a concrete named type for use in generic bounds.
+pub type ShutdownSignal =
```
theswiftfox: q: do we need + Sync here? Otherwise this could be reduced to `futures::future::Shared<futures::future::BoxFuture<()>` as [BoxFuture](https://docs.rs/futures/latest/futures/future/type.BoxFuture.html) is an alias for the Pin used here, just without the Sync. But I'd assume the Shared should take care of the sync?

theswiftfox (commented):
just a few first comments. need to check the second half more in depth still.

cda-interfaces/src/ecugateway.rs
alexmohr: whops, should have been udp, but I agree the protocol is not important, so i remove the prefix

cda-interfaces/src/lib.rs
alexmohr: I keep adding this import because RustRover complains about `Use of an undeclared crate or module `futures` [E0433]` every time and eventually it gets me to click 'add use' every time 😵‍💫 
I'm 99% sure you will have to comment this a couple more times until this settles in my brain. 
What do you think about adding https://github.com/trailofbits/dylint, so we can check for such an import automatically?

cda-interfaces/src/lib.rs
alexmohr: thanks for the hint. Makes the impl of shutdown_signal way neater. :)

alexmohr:
Closing this in favor of #465  which is rebased onto #323 because I want the CAN PR to be merged first. #465 is kept as draft until #323 is merged. Using a new PR because I want to keep the original branch around in case of rebase mishaps.

PR #427, opened 2026-07-09
feat: add proc-macro for vendor overrides

theswiftfox:
## Summary

Introduces a `vendor_overridable` annotation macro that allows to replace a given function with a vendor override by impelemting it with the `vendor_override` annotation. This allows to have the possibility to override functions at compile time without the need of using traits everywhere.

## Checklist

## Notes for Reviewers

---


lh-sag:
@theswiftfox for the sake of curiosity why did you chose to introduce compile-time overrides?

override-macros/src/lib.rs
```diff
@@ -0,0 +1,846 @@
...
+                the crate root so external crates can register overrides")]
+            #[doc = #hook_doc]
+            pub static #hook_ident: [fn(#(#hook_param_types),*) -> ::core::option::Option<#ret_ty>];
+
+            #(#doc_attrs)*
+            #[doc = ""]
+            #[doc = #dispatcher_doc]
+            #[deny(unreachable_pub, reason = "vendor_overridable items must be reachable from \
+                the crate root so external crates can register overrides")]
+            pub fn #public_name #generics (#inputs) -> #ret_ty #where_clause {
```
mbfm: When I've changed up a function like this before, I found it much more robust, when instead of taking out the individual parts (like `generics`, `inputs` etc.) and piecing them back together afterwards, I would instead clone the `ItemFn` object of the original function and just mutate the fields that I wanted to adjust. The `ItemFn` object can later be turned into a `TokenStream` again.

For example, currently this code is not preserving whether the original function is `async`, which I'm not sure, is something we won't need in the future.

Using mutation should also help reduce code duplication here, between the `any_erased` case and the else-case.

override-macros/tests/duplicate.rs
```diff
@@ -0,0 +1,46 @@
...
+fn double_vendor_b(value: u32) -> u32 {
+    value.saturating_mul(2).saturating_add(2000)
+}
+
+#[test]
+fn validation_reports_duplicate_override() {
+    let result = validate_vendor_overrides();
+    let errors = result.expect_err("expected duplicate override registration to be reported");
+    assert_eq!(errors.len(), 1);
+    let message = errors.first().expect("checked len() == 1 above");
```
mbfm, line 43: Could check the length + unpack in one go with a let-else:

```suggestion
    let [message] = errors.as_slice() else {
        panic!("Expected one error message!");
    };
```

I'll leave it up to you, though, if you actually want to use that syntax. It's certainly not a very widely known... 😅

override-macros/src/lib.rs
```diff
@@ -0,0 +1,846 @@
...
+///
+/// At most one override may be registered per overridable function (generic
+/// or not). This is enforced two ways:
+/// - the crate that defines this function must call
+///   [`macro@declare_vendor_override_registry`] once at its crate root.
+///   Every [`macro@vendor_override`] attachment then self-registers into
+///   that crate's registry with no manual bookkeeping required, and the
+///   final binary calls the generated `<crate>::validate_vendor_overrides()`
+///   once at startup, which fails loudly if more than one override is
+///   linked in for the same hook (`linkme` slice order is otherwise
```
mbfm: This reads to me like the final binary calls `validate_vendor_overrides()` automatically, when the code using it always calls it manually...

override-macros/src/lib.rs
```diff
@@ -0,0 +1,846 @@
...
+        .filter_map(|arg| match arg {
+            syn::FnArg::Typed(pat_type) => match pat_type.pat.as_ref() {
+                syn::Pat::Ident(pi) => Some(pi.ident.to_string()),
+                _ => None,
+            },
+            syn::FnArg::Receiver(_) => None,
+        })
+        .collect();
+    for name in erased {
+        if !param_names.iter().any(|p| name == p) {
```
mbfm: Can compare the `Ident`s directly here:

```suggestion
    let param_names: Vec<&Ident> = func
        .sig
        .inputs
        .iter()
        .filter_map(|arg| match arg {
            syn::FnArg::Typed(pat_type) => match pat_type.pat.as_ref() {
                syn::Pat::Ident(pi) => Some(&pi.ident),
                _ => None,
            },
            syn::FnArg::Receiver(_) => None,
        })
        .collect();
    for name in erased {
        if !param_names.iter().any(|p| name == *p) {
```

Side-note: Could also consider using a `HashSet` here, especially for finding out the difference between the two sets...

mbfm (commented):
Initial review comments. I still have to work my way through the rest of `override-macros/src/lib.rs`.

override-macros/src/lib.rs
```diff
@@ -0,0 +1,846 @@
...
+
+        // Compile-time check: the override signature must be a valid
+        // instantiation of the overridable function.
+        const _: fn(#(#assert_param_types),*) -> #ret_ty = #public_fn_path;
+
+        #[::linkme::distributed_slice(#hook_path)]
+        fn #shim_ident(#(#shim_params),*) -> ::core::option::Option<#ret_ty> {
+            #(#shim_bindings)*
+            ::core::option::Option::Some(#fn_ident(#(#shim_call_args),*))
+        }
```
mbfm: Similar as above, might want to consider using mutation of `ItemFn` object rather than putting the tokens back together manually.

override-macros/src/lib.rs
```diff
@@ -0,0 +1,846 @@
...
+
+    // Adapter generation: emit the override function unchanged plus a shim
+    // that downcasts erased parameters and is registered in the hook slice.
+    let fn_name = fn_ident.to_string();
+    let shim_ident = Ident::new(
+        &format!("__{fn_name}_vendor_override_shim"),
+        fn_ident.span(),
+    );
+    let output = &func.sig.output;
+    let ret_ty = match output {
```
mbfm:
```suggestion
    let ret_ty = match &func.sig.output {
```

override-macros/src/lib.rs
```diff
@@ -0,0 +1,846 @@
...
+    let mut hook_path = args.hook_path;
+
+    // Replace the last path segment with the derived hook static name.
+    // e.g. `do_something` -> `DO_SOMETHING_HOOK`
+    let Some(last_seg) = hook_path.segments.last_mut() else {
+        return syn::Error::new_spanned(&hook_path, "vendor_override: expected a non-empty path")
+            .to_compile_error()
+            .into();
+    };
+    last_seg.ident = hook_ident_for(&last_seg.ident);
```
mbfm: Might be worth visually grouping, as I did not realize right away that this is still adjusting `hook_path`. I generally do it this way:

```suggestion
    let hook_path = {
        let mut hook_path = args.hook_path;

        // Replace the last path segment with the derived hook static name.
        // e.g. `do_something` -> `DO_SOMETHING_HOOK`
        let Some(last_seg) = hook_path.segments.last_mut() else {
            return syn::Error::new_spanned(&hook_path, "vendor_override: expected a non-empty path")
                .to_compile_error()
                .into();
        };
        last_seg.ident = hook_ident_for(&last_seg.ident);
        hook_path
    };
```

That way, `hook_path` is also not anymore mutable afterwards, making it more obvious that this is the only place where it is mutated.

override-macros/src/lib.rs
mbfm, line 1: Would prefer for this file to be split up. It is very long and complex. 😅

In the past, I've found creating a submodule for each proc-macro works quite well. The function annotated with `#[proc_macro]` still needs to be in the `lib.rs`, due to the way proc-macros work, but then even the code inside, I would just move into an implementation-function in the submodule.

I liked this split in particular, because it meant that everything in the submodule could consistently return `syn::Result<_>` and then only in the `lib.rs`, you'd do one `.to_compile_error()` to render out the actual error.

mbfm (commented):
Alright, done with the review.

theswiftfox:
> @theswiftfox for the sake of curiosity why did you chose to introduce compile-time overrides?

@lh-sag  oh, sorry that question somehow completely went under my radar.
Uhm, we noticed that for example the security access service lookup can be quite different for different vendors. When we had a discussion about it it came down to either introduce a trait where a vendor can then insert a plugin that needs to be handed through from main (and adds yet another generic) or to try and fit every possible way to do the lookup in the supplied function (can get quite complex and break easily).

As both options didn't sound too great from a maintenance standpoint I came up with an idea that allows to supply an override method while providing a reference implementation in the CDA. This seemed to me to offer the greatest flexibility with the least effort down the line, as we can simply extend the list of overridable logic over time whenever we notice that some vendor might need a slightly different flavor that does not make sense to build into the open source variant, without it changing any API contract by doing so.

(I hope that makes sense, the idea was floating for quite some time, so its not easy to pinpoint a single exact reason, but I tried to condense the reasoning ^^)

override-macros/src/shared.rs
mbfm, line 1: Would be good to sort the visitor declarations.
There's always a type, e.g. `NameElidedLifetimes`, then an `impl VisitMut` for that type and a function that applies the visitor, e.g. `with_lifetime()`. These seem to be in an almost random order in this file, which makes it really difficult to see that pattern.

Maybe those functions could also be associated with the type to make this more obvious, e.g. `with_lifetime()` → `NameElidedLifetimes::with_lifetime()`.

override-macros/src/vendor_override.rs
```diff
@@ -109,17 +117,36 @@ fn build_shim_parts(func: &ItemFn, erased: &[Ident], fn_name: &str) -> syn::Resu
...
                     quote! { downcast_mut },
                 )
             } else {
-                (quote! { &dyn ::core::any::Any }, quote! { downcast_ref })
+                (
+                    async_lifetime.map_or_else(
+                        || quote! { &dyn ::core::any::Any },
+                        |lifetime| {
+                            quote! {
+                                &#lifetime (dyn ::core::any::Any + ::core::marker::Sync)
```
mbfm, line 139: Is it intentional that we only require `Sync`, not `Send + Sync`, as is normally done?

From what I can find online, this is only really a thing in niche cases: https://users.rust-lang.org/t/sync-but-not-send/21551  
Although, if we do only need `Sync`, then we should not restrict it further.

Just wondering, if this is an intentional choice...

override-macros/src/vendor_override.rs
```diff
@@ -195,16 +230,23 @@ pub(crate) fn expand(
...
+    let ret_type = match &func.sig.output {
+        ReturnType::Default => parse_quote! { () },
+        ReturnType::Type(_, ty) => ty.as_ref().clone(),
     };
+    let ret_ty = quote! { #ret_type };
+    let async_lifetime: syn::Lifetime = parse_quote!('__vendor_override);
+    let has_borrowed_input = func.sig.inputs.iter().any(|arg| match arg {
+        syn::FnArg::Typed(pat_type) => contains_reference(&pat_type.ty),
+        syn::FnArg::Receiver(_) => false,
+    });
```
mbfm: This code is duplicated between `vendor_override.rs` and `vendor_overridable.rs`. And `contains_reference()` is only used in the duplicated code.

So, instead of the `contains_reference()` function, the `shared` module could expose a `has_borrowed_input()` function.

override-macros/src/vendor_override.rs
theswiftfox: so, we are only passing a reference here, not an object. As the boxed future might be polled from another thread it needs to be sync. But as we never send the whole future in this snippet we do not need to constraint it to send.
In practice I'd assume almost any type that gets in here that fulfills sync to also be send. But in the idea of constraining only the necessary bounds I'd rather leave out enforcing send.

small edit: and `T: Sync` implies `&T: Send` as a shared reference to a sync object is always send.

override-macros/tests/duplicate.rs
theswiftfox: While I like the pattern, I'd keep it as is for now. I think everywhere else in the codebase of the CDA we are already using the assert -> .expect pattern. So I'd keep it in line.

override-macros/src/vendor_override.rs
mbfm: Makes sense, thanks for the explainer. 🙂

mbfm (approved):

theswiftfox:
@mbfm sry, had to force push again, as i mixed up the commit emails 🙈️
I'll ignore the conventional commits check failure in pre-commit, as i plan to squash the changes anyways. just added as extra commit to have it easier to review.
The last commit should address your remaining findings :)

mbfm (approved):

floroks (approved):

PR #428, opened 2026-07-10
fix: Install `reuse` via `uv`.

mbfm:
## Summary

This change installs `reuse` via `uv` to ensure devs all use the same version.

It also simplifies the project setup, since devs don't need to manually install `reuse` on their PC.

## Checklist

## Notes for Reviewers

I don't think, it was documented before that the project setup needs `reuse`, so there was no documentation to update.

---

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,4 +15,8 @@ version = "0.1.0"
 requires-python = ">=3.13"
 
 [dependency-groups]
-tools = ["prek==0.3.13", "ruff==0.15.12"]
+tools = [
+    "prek==0.3.13",
+    "reuse==6.2.0",
+    "ruff==0.15.12",
+]
--- a/uv.lock
+++ b/uv.lock
@@ -3,6 +3,24 @@ revision = 3
 requires-python = ">=3.13"
 
 [[package]]
+name = "attrs"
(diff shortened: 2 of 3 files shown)


mbfm:
I just realized that in the CI pipeline, it needs to have been automatically installed somehow before this change, too, which seems to be done by this line: https://github.com/eclipse-opensovd/cicd-workflows/blob/f4d36f3f68ace4e01c6681e8d4ab89e109e6b818/pre-commit-action/action.yml#L48

Do we need to unify these somehow? Should I pin `reuse` to the same version as in that pre-commit-action?

alexmohr:
> I just realized that in the CI pipeline, it needs to have been automatically installed somehow before this change, too, which seems to be done by this line: https://github.com/eclipse-opensovd/cicd-workflows/blob/f4d36f3f68ace4e01c6681e8d4ab89e109e6b818/pre-commit-action/action.yml#L48
> 
> Do we need to unify these somehow? Should I pin `reuse` to the same version as in that pre-commit-action?

Although I authored action you linked I think we have to 
1. Pin to an exact version, but should be through a SHA, I really don't want to get pwned because we're using latest and greatest of a tool. 
2. If we can unify we should

Issue #325, opened 2026-05-05
Follow-up: systemd_notify improvements from PR #322 review


theswiftfox:
## Context

Post-merge review comments on #322 surfaced a few improvements for the `cda-extra` systemd_notify implementation. This issue tracks implementing them in a follow-up PR.

### 1. Signal `Stopping` to systemd on shutdown

When the shutdown signal is received in the `tokio::select!`, the task currently exits without notifying systemd. We should send `NotifyState::Stopping` before exiting so systemd knows the service is shutting down intentionally rather than crashing.

### 2. Use `timeout / 2` for watchdog notify interval (alignmen with systemd recommendation)

The current implementation uses `timeout - 5s` (floored to 1s), which works correctly but deviates from the [systemd-recommended convention](https://www.freedesktop.org/software/systemd/man/latest/sd_event_set_watchdog.html) of notifying at half the watchdog timeout. Switching to `timeout / 2` aligns with the documented best practice and scales more predictably across different timeout configurations.

### 3. Set `default-features = false` on `sd-notify` workspace dependency

In the root `Cargo.toml`, the `sd-notify` dependency should use `default-features = false` to follow the project's convention of minimizing pulled-in features. This has no functional change to the current implementation, as the `sd-notify` crate does not enable any features per default, but aligns with the other dependencies.

## References

- Original PR: #322
- Review comments: https://github.com/eclipse-opensovd/classic-diagnostic-adapter/pull/322#pullrequestreview-4219198399

mohammadp1001:
Hi Elena

I would like to work on this issue if possible.

theswiftfox:
Hey! :)
Sounds great! I assigned the issue to you

PR #433, opened 2026-07-11
fix(sd_notify): send Stopping on shutdown and use timeout/2 for watchdog interval

mohammadp1001:
## Summary

Follow-up to PR #322's post-merge review, addressing items 1, 2, and 3 from #325 in
`cda-extra`'s `systemd_notify` implementation:

- Send `NotifyState::Stopping` to systemd when the shutdown signal fires, so an
  intentional shutdown isn't mistaken for a watchdog failure.
- Replace the `timeout - 5s` (floored to 1s) watchdog notify interval with
  `timeout / 2`, matching systemd's documented convention
  ([`sd_watchdog_enabled(3)`](https://www.freedesktop.org/software/systemd/man/latest/sd_event_set_watchdog.html)).
  The previous formula collapsed to "always notify every 1s" for any realistic
  watchdog timeout, since `.min(1s)` was acting as a cap rather than a floor.
- Renamed `determine_interval` to `watchdog_notify_interval` (item 2.1) to make
  its purpose clearer, and updated the doc comments accordingly.
- Set `default-features = false` on the workspace `sd-notify` dependency in
  the root `Cargo.toml`. No functional change: `sd-notify` enables no
  features by default today.

## Checklist

## Related

Follow-up to #322


cda-extra/src/systemd_notify.rs
```diff
@@ -73,16 +81,11 @@ where
...
-            interval.min(Duration::from_secs(1))
-        } else {
-            duration
-        }
-    })
+/// Interval at which watchdog notifications should be sent: half of the
+/// systemd-configured watchdog timeout, per the systemd-recommended convention
+/// (see `sd_watchdog_enabled(3)`).
+fn watchdog_notify_interval() -> Option<Duration> {
+    sd_notify::watchdog_enabled().map(|timeout| timeout / 2)
```
alexmohr: suggestion: We should keep some safe guard, so Duration != Duration::Zero because tokio will panic when using Duration::Zero as tick interval

theswiftfox:
@mohammadp1001 can you check the pipeline failure and fix the reported clippy lint?

mohammadp1001:
Sure, I will check.

cda-extra/src/systemd_notify.rs
alexmohr: @mohammadp1001 Please do not resolve my threads yourself. I will resolve them if I agree with the solution. It helps tracking what I commented and what I reviewed, instead of going through the closed threads.

mohammadp1001: Hi, Anything else missing?

alexmohr: 1ms interval will lead to considerable system load. It's legal from the systemd configuration as it supports intervals starting a 1 micro second. But imho it's not a sane configuration for the CDA, we should log a warning once.

mohammadp1001: Thank you for pointing this out. I set a threshold of 1 second anything below will be discarded.

cda-extra/src/systemd_notify.rs
```diff
@@ -62,26 +74,49 @@ where
...
+/// Clamped to at least 1ms so a very small (or zero) `WatchdogSec` can never
+/// produce a zero-length interval, which `tokio::time::interval` panics on.
+///
+/// systemd accepts watchdog timeouts down to microsecond resolution, which is a
+/// legal but implausible configuration for the CDA: it would wake the notify task
+/// hundreds or thousands of times per second for no benefit. Such a configuration
+/// is honoured, but warned about once at startup.
+fn watchdog_notify_interval() -> Option<Duration> {
+    sd_notify::watchdog_enabled().map(|timeout| {
+        // `checked_div` instead of `/` to satisfy clippy::arithmetic_side_effects;
```
theswiftfox, line 105: nit: please remove `checked_div` instead of `/` to satisfy clippy::arithmetic_side_effects;

cda-extra/src/systemd_notify.rs
```diff
@@ -62,26 +74,49 @@ where
     let sd_notify_future = spawn_named!("sd_notify", async move {
         tokio::select! {
             _ = watchdog_future => {},
-            () = shutdown_signal => {},
+            () = shutdown_signal => {
```
theswiftfox, line 77: thanks for this, was completely missing. 
related question: I think we are also missing the send of the NotifyState::Running after startup? Do you think it would make sense to add this in addition to the clean shutdown?

mohammadp1001: Is there any Running state at all?

alexmohr: Systemd watchdog has

```
       READY=1
           Tells the service manager that service startup is finished, or
           the service finished re-loading its configuration. This is
           only used by systemd if the service definition file has
           Type=notify or Type=notify-reload set. Since there is little
           value in signaling non-readiness, the only value services
           should send is "READY=1" (i.e.  "READY=0" is not defined).
```

Which maps to `NotifyState::Ready`. It could be set in cda-main once the loading all things is done. I.e. in `run_with_ext_from_config`

alexmohr: @mohammadp1001 Do you want to add the change in here or should we create a follow up?

mohammadp1001: I am almost done. I will push my code.

alexmohr (approved):
lgtm, I stil want to give @theswiftfox the chance to take a look at this, as she raised the notify ready point.

theswiftfox (approved):

PR #434, opened 2026-07-13
feat(diag-kernel): dop fixes

floroks:
## Summary

feat(diag-kernel): implement TABLE-KEY and TABLE-STRUCT parameter encode/decode

Add support for encoding and decoding TABLE-KEY and TABLE-STRUCT ODX
parameter types, previously unimplemented (logged as an error).

- Decode: resolve TABLE-KEY via its key DOP's compu method to obtain
  the row key, then look up the selected row's structure to decode
  TABLE-STRUCT contents.
- Encode: resolve the key string to its wire representation via the
  key DOP, validate it against available table rows, and encode the
  companion TABLE-STRUCT using the selected row's structure DOP.
- Update integration tests and test ECU simulator (Kotlin) with
  routine functionalities and ODX/mdd fixtures covering these
  parameter types.

---

fix(diag-kernel): fix EndOfPdu request arrays overwriting each other

When encoding a request parameter backed by an EndOfPdu DOP with more
than one repeated item (e.g. a fixed min==max item count), each item
was encoded at the exact same absolute byte offset instead of being
appended after the previous one.

map_param_value_to_uds computed the struct's byte position once,
before the loop over the JSON array, and reused that fixed offset for
every item. DiagCodedType::encode writes each field at an absolute
byte offset in the payload buffer (resizing and overwriting a byte
range), rather than appending. As a result every item after the first
clobbered the bytes of the previous item at the same offset, and the
final payload only retained whichever item was encoded last.

This affects any EndOfPdu array parameter with more than one item in a
request, such as bidirectional-authentication-style requests carrying
multiple leading-length-prefixed challenge/proof structures - the
first entries were silently dropped from the wire payload while the
JSON call appeared to succeed.

Fix by advancing the write position for each item to the current end
of the payload (bounded below by the field's own start offset), so
items are appended sequentially instead of overwriting one another.

Add create_ecu_manager_with_end_pdu_request_service test-builder plus
a regression test exercising a request with a fixed 2-item EndOfPdu
array of leading-length-prefixed fields, verifying both items are
present in the encoded payload.

## Checklist

## Notes for Reviewers


cda-core/src/diag_kernel/payload_decode.rs
```diff
@@ -610,6 +622,195 @@ impl<S: SecurityPlugin> EcuManager<S> {
         Ok(())
     }
 
+    /// Decode a TABLE-KEY parameter from a UDS response payload.
+    /// Reads the key byte(s), resolves through the key DOP's compu method
+    /// to the physical representation (row key string), and stores it in `data`.
+    fn map_table_key_from_uds(
+        _mapped_service: &datatypes::DiagService,
```
theswiftfox: suggestion: if we don't need the parameter we can remove it to keep the argument list shorter :)

floroks: removed the parameter

theswiftfox (approved):

PR #437, opened 2026-07-14
fix(storage): resolve mixed-case filenames consistently across Collection ops & nextupdate behavior

floroks:
## Summary

fix(storage): resolve mixed-case filenames consistently across Collection ops

Collection::list() returned on-disk filenames verbatim, while metadata(), read(), delete(), and file_path() normalize the key to lowercase before resolving it to a path. Files placed on disk with mixed-case names (e.g. OEM-style MDD files provisioned outside of Collection::write) would appear in list() but then fail every subsequent per-key lookup with KeyNotFound, surfacing as spurious "key vanished during iteration" warnings when listing runtime files.

- list_keys_in_dir() now returns normalized (lowercase) keys, consistent with the rest of the Collection API.
- Added LocalCollection::ensure_normalized_on_disk(), called from read, metadata, delete, and file_path: if a case-insensitive match exists but the exact lowercase-named file doesn't, it is renamed in place so the on-disk state matches the key-normalization invariant.
- Added a regression test that places a mixed-case file directly on disk and verifies list()/metadata()/read() stay consistent.

---

fix(runtime-update): mirror runtimefiles-current in nextupdate when no update is pending

compute_nextupdate_state previously returned an empty list per category
(MDD/Configuration) whenever the corresponding NextUpdate collection did
not exist yet, instead of falling back to the currently active database.
This contradicted the documented spec, which states that
runtimefiles-nextupdate must initially mirror runtimefiles-current.

- compute_nextupdate_state now falls back to the current collection per
  category when its NextUpdate collection doesn't exist, while still
  treating an existing-but-empty NextUpdate collection as an explicit
  "delete all" target state.
- delete_nextupdate_file now lazily initializes NextUpdate from current
  before deleting, so files that only exist in current can be removed
  from the pending next-update state instead of returning FileNotFound.
- Updated unit tests to assert the mirrored-fallback behavior.
- Updated integration test assertions/helpers that previously expected
  nextupdate to be empty after apply/rollback/cleanup/delete-all to
  instead expect it to mirror runtimefiles-current.

## Checklist

## Notes for Reviewers

### DIFF
--- a/cda-storage/src/local_collection.rs
+++ b/cda-storage/src/local_collection.rs
@@ -62,17 +62,74 @@ impl LocalCollection {
         paths::sanitize_path_segment(key)?;
         Ok(self.dir.join(key))
     }
+
+    /// Ensures that a file matching `key` (already normalized to lowercase) exists on disk
+    /// under exactly that (lowercase) name.
+    ///
+    /// All keys are normalized to lowercase (see [`normalize_key`]), and [`list`](Collection::list)
+    /// returns keys in that same normalized form. However, a file can end up on disk with a
+    /// different case than its normalized key - for example when files are placed directly in
+    /// the collection directory outside of [`Collection::write`] (such as during initial
+    /// provisioning of OEM-supplied files, which often use mixed-case names). If left
+    /// unaddressed, such a file would be listed (via its normalized key) but every other
+    /// operation (`metadata`, `read`, `delete`, `file_path`) would fail with `KeyNotFound`,
+    /// since they resolve the path via the lowercase key directly.
+    ///
+    /// This self-heals that situation: if no file exists at the exact lowercase path but one
+    /// exists under a different case, it is renamed in place to the normalized name (a metadata-
+    /// only, same-filesystem operation) so that the on-disk state matches the key-normalization
+    /// invariant relied on everywhere else. No-op if the normalized path already exists or no
+    /// case-insensitive match is found.
+    ///
+    /// This is a best-effort, synchronous operation (no lock is taken): the rename is atomic
+    /// and idempotent, so a benign race with a concurrent caller doing the same self-heal (or
+    /// with a commit) simply results in a harmless `rename` failure that is ignored, since the
+    /// target either already exists in the desired state or will shortly.
+    fn ensure_normalized_on_disk(&self, key: &str) {
+        let target = self.dir.join(key);
+        if target.exists() {
+            return;
+        }
+
+        let Ok(entries) = std::fs::read_dir(&self.dir) else {
+            return;
+        };
+
+        for entry in entries.flatten() {
+            let path = entry.path();
+            let Some(name) = path.file_name().and_then(|n| n.to_str()) else {
+                continue;
+            };
+            // For valid files which would be listed, check if the name deviates, and normalize
+            if is_valid_key(&path) && name != key && name.eq_ignore_ascii_case(key) {
+                // Best-effort: ignore errors (e.g. lost the race to another renamer, or the
+                // target now exists).
+                let _ = std::fs::rename(&path, &target);
+                break;
+            }
+        }
+    }
 }
 
 /// Normalize a key to lowercase for case-insensitive storage.
 fn normalize_key(key: &str) -> String {
-    key.to_lowercase()
+    key.to_ascii_lowercase()
+}
+
+/// Checks if a key should be listed in a collection
+fn is_valid_key(path: &Path) -> bool {
+    let bak_ext = std::ffi::OsStr::new(crate::recovery::BACKUP_EXTENSION);
+    let tmp_ext = std::ffi::OsStr::new(crate::recovery::STAGING_EXTENSION);
+
+    // Only files are valid, when they are not backup or staging files.
+    path.is_file() && path.extension() != Some(bak_ext) && path.extension() != Some(tmp_ext)
 }
 
 impl Collection for LocalCollection {
     async fn read(&self, key: &str) -> Result<Arc<impl RandomAccessData + 'static>, StorageError> {
         let key = normalize_key(key);
         let path = self.key_path(&key)?;
+        self.ensure_normalized_on_disk(&key);
 
         let guard = io::acquire_read_lock(&self.data_lock).await;
 
@@ -87,6 +144,7 @@ impl Collection for LocalCollection {
     async fn metadata(&self, key: &str) -> Result<Metadata, StorageError> {
         let key = normalize_key(key);
         let path = self.key_path(&key)?;
+        self.ensure_normalized_on_disk(&key);
 
         let _guard = io::acquire_read_lock(&self.data_lock).await;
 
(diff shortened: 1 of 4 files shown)


cda-storage/src/local_collection.rs
```diff
@@ -62,6 +62,52 @@ impl LocalCollection {
         paths::sanitize_path_segment(key)?;
         Ok(self.dir.join(key))
     }
+
+    /// Ensures that a file matching `key` (already normalized to lowercase) exists on disk
+    /// under exactly that (lowercase) name.
+    ///
+    /// All keys are normalized to lowercase (see [`normalize_key`]), and [`list`](Collection::list)
+    /// returns keys in that same normalized form. However, a file can end up on disk with a
+    /// different case than its normalized key -- for example when files are placed directly in
```
alexmohr:
```suggestion
    /// different case than its normalized key - for example when files are placed directly in
```

cda-storage/src/local_collection.rs
```diff
@@ -62,6 +62,52 @@ impl LocalCollection {
...
+        let Ok(entries) = std::fs::read_dir(&self.dir) else {
+            return;
+        };
+
+        for entry in entries.flatten() {
+            let path = entry.path();
+            let Some(name) = path.file_name().and_then(|n| n.to_str()) else {
+                continue;
+            };
+            if path.is_file() && name != key && name.eq_ignore_ascii_case(key) {
```
alexmohr: question: why are we ignoring directories here? Right now our use case only supports files, so it's fine but it seems like an oversight or at least something worth a comment.

cda-storage/tests/local_storage_tests.rs
```diff
@@ -830,3 +830,52 @@ async fn recovery_create_collection_with_write_removes_orphaned_dir() {
...
+    // Every key returned by `list()` must be resolvable via `metadata()` without a
+    // `KeyNotFound` error.
+    for key in &keys {
+        let meta = collection.metadata(key).await.unwrap();
+        assert_eq!(meta.data_size, "mixed case data".len() as u64);
+    }
+
+    // Every key returned by `list()` must also be resolvable via `read()`.
+    let Some(key) = keys.first() else {
+        panic!("expected one key");
```
alexmohr, line 875: question: What happens if we have `FooBar` and `foobar` somehow on disk? Should we guard against that case as well (or at least log something?)

cda-storage/src/local_collection.rs
```diff
@@ -206,6 +256,14 @@ impl DirectFileAccess for LocalCollection {
...
 
 /// List all file names (keys) in a directory, excluding backup and staging files.
+///
+/// Returned keys are normalized to lowercase to match the invariant used by
+/// [`Collection::read`], [`Collection::metadata`], and [`DirectFileAccess::file_path`], all of
+/// which normalize the key before resolving it to a path. Without this normalization, a file
+/// that ended up on disk with a mixed-case name (e.g. placed there outside of
+/// [`Collection::write`], such as during initial provisioning) would be listed under its
+/// original case but then fail to resolve via `metadata`/`read`/`file_path`, which look up the
+/// lowercased path -- causing spurious "key not found" errors on case-sensitive filesystems.
```
alexmohr:
```suggestion
/// lowercased path - causing spurious "key not found" errors on case-sensitive filesystems.
```

alexmohr: There are a few more `--` instead of `-`, I won't comment the rest though :)

cda-storage/src/local_collection.rs
floroks: fixed

cda-storage/src/local_collection.rs
floroks: I added a `is_valid_key` function, used when listing keys, and for the normalization, and added comments.

cda-storage/tests/local_storage_tests.rs
floroks: Not sure what the best resolution for that would be, in practice it can only happen when a user or an application writes same name files with different cases manually into the directories. 

The fixed case with the upper-case names, which triggered this PR could reasonably happen on an initial database, when naming was just in upper-case instead of lower-case like expected.

cda-storage/src/local_collection.rs
floroks: fixed, could only find a single extra one though

floroks:
added a second commit https://github.com/eclipse-opensovd/classic-diagnostic-adapter/pull/437/changes/860488cfc502529128aaa4a690ea400c2366c3c2 to fix that runtimfiles-nextupdate should also show the current files when no file was previously uploaded. Now delete or upload triggers the actual initial copy of the data from current into next, and when not nex collection exists, the current collection is returned.

cda-storage/tests/local_storage_tests.rs
alexmohr: Makes sense, might be worth documenting this (if it not already is)? Not important for this PR though

floroks: Created an issue, so we can fix/change the behavior in the future https://github.com/eclipse-opensovd/classic-diagnostic-adapter/issues/438

cda-storage/src/local_collection.rs
```diff
@@ -62,17 +62,74 @@ impl LocalCollection {
...
+                let _ = std::fs::rename(&path, &target);
+                break;
+            }
+        }
+    }
 }
 
 /// Normalize a key to lowercase for case-insensitive storage.
 fn normalize_key(key: &str) -> String {
     key.to_lowercase()
```
theswiftfox: suggestion: as we are using `eq_ignore_ascii_case` in `ensure_normalized_on_disk`, we should probably also normalize to ascii here via `key.to_ascii_lowercase()`

cda-storage/src/local_collection.rs
```diff
@@ -62,17 +62,74 @@ impl LocalCollection {
...
+        for entry in entries.flatten() {
+            let path = entry.path();
+            let Some(name) = path.file_name().and_then(|n| n.to_str()) else {
+                continue;
+            };
+            // For valid files which would be listed, check if the name deviates, and normalize
+            if is_valid_key(&path) && name != key && name.eq_ignore_ascii_case(key) {
+                // Best-effort: ignore errors (e.g. lost the race to another renamer, or the
+                // target now exists).
+                let _ = std::fs::rename(&path, &target);
```
theswiftfox, line 107: note: this will overwrite if target already exists (eg. if a transaction just created the updated file xyz.mdd and we are renaming the old XYZ.mdd this would cause us to overwrite the new file with the older one)
[source](https://doc.rust-lang.org/std/fs/fn.rename.html)

floroks: There's already an issue created for this, see https://github.com/eclipse-opensovd/classic-diagnostic-adapter/pull/437#discussion_r3585643291 - it should pretty much not happen in real world scenarios, but we need to find & discuss a proper resolution.

theswiftfox: oh whops :D Resolving then!

cda-storage/src/local_collection.rs
floroks: good catch, fixed it

alexmohr (approved):

PR #440, opened 2026-07-17
fix(diag_kernel): invalidate stale service cache

theswiftfox:
## Summary

Observed following error when the ECU-State changed:
```
payload_encode.rs:75 msg: error=Not found: "Cached diagnostic service 'service_abc_redacted' not found at stored location"
    in create_uds_payload::create_uds_payload
    in cda_comm_uds::variant::detect_variant
    in cda_comm_uds::transport::send_with_optional_timeout
    in cda_sovd::request
```

This PR changes the logic so it:
- Associates cached service locations with their ECU variant so lookups ignore stale entries after variant changes or reconnects. This ensures that we don't hit a stale cache if the variant index changed but the db-cache was not reset.
- Clears db-cache when variant is cleared (NotTested / Duplicate)

## Checklist

## Notes for Reviewers

---

### DIFF
--- a/cda-core/src/diag_kernel/variant.rs
+++ b/cda-core/src/diag_kernel/variant.rs
@@ -23,14 +23,21 @@ use super::ecumanager::{EcuManager, VariantData};
 use crate::diag_kernel::variant_detection;
 
 impl<S: SecurityPlugin> EcuManager<S> {
-    fn clear_variant(&self, variant_state: VariantState, connectivity: Option<Connectivity>) {
-        let mut ecu_state = std_ext::lock_write(&self.runtime_state.ecu_state);
-        ecu_state.variant_state = variant_state;
-        ecu_state.variant_index = None;
-        if let Some(connectivity) = connectivity {
-            ecu_state.connectivity = connectivity;
+    async fn clear_variant_and_unload(
+        &mut self,
+        variant_state: VariantState,
+        connectivity: Option<Connectivity>,
+    ) {
+        {
+            let mut ecu_state = std_ext::lock_write(&self.runtime_state.ecu_state);
+            ecu_state.variant_state = variant_state;
+            ecu_state.variant_index = None;
+            if let Some(connectivity) = connectivity {
+                ecu_state.connectivity = connectivity;
+            }
         }
-        drop(ecu_state);
+        self.db_cache.reset().await;
+        self.diag_database.unload();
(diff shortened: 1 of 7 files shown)


cda-interfaces/src/ecumanager.rs
```diff
@@ -767,11 +767,11 @@ pub trait VariantDetection: Send + Sync + 'static {
 
     /// Mark this ECU as duplicate. Sets the variant state to [`VariantState::Duplicate`] and
     /// unloads the database. Database will be reloaded before next variant detection.
-    fn mark_as_duplicate(&mut self);
+    fn mark_as_duplicate(&mut self) -> impl Future<Output = ()> + Send;
 
     /// Mark this ECU as having no variant detected. Sets the variant state to
     /// [`VariantState::NotDetected`] and unloads the database.
-    fn mark_as_no_variant_detected(&mut self);
+    fn mark_as_no_variant_detected(&mut self) -> impl Future<Output = ()> + Send;
```
mbfm: Is it intentional here that you used `impl Future` rather than `#[async_trait]` ?

theswiftfox: yes and no, it avoided having to put it on the impl as the trait was not already async_trait.
But it makes no sense to not just mark it as async_trait if 3/5 functions are async. So thanks for the hint! :D will change!

cda-core/src/diag_kernel/variant.rs
```diff
@@ -21,6 +22,19 @@ use cda_plugin_security::SecurityPlugin;
 use super::ecumanager::{EcuManager, VariantData};
 use crate::diag_kernel::variant_detection;
 
+impl<S: SecurityPlugin> EcuManager<S> {
+    fn clear_variant(&self, variant_state: VariantState, connectivity: Option<Connectivity>) {
```
alexmohr: suggestion: all usages of this function call `self.db_cache.reset()` directly after calling. Maybe it would be better to call the reset here instead of the caller, so future usages of this function can't forget to reset the cache.

theswiftfox: thanks!
I went one step further and moved both, `self.db_cache.reset().await` and `self.diag_database.unload()` to the function, as all 3 were always caleld together.

alexmohr (approved):

PR #450, opened 2026-07-21
feat(transport): add transport orchestrator for multi-transport routing

### LINKED ISSUE #442: Improvement: extract multi-transport orchestration into its own layer
## Problem

`MultiTransportGateway` (per-ECU transport pinning, sticky first-detection
binding, routing between DoIP and CAN) lives inside the `cda-comm-can`
crate, so the CAN transport crate is no longer purely a transport: it hosts
the layer that arbitrates between transports, and `cda-main` wires DoIP
through it.

From the PR #323 review:

> theswiftfox: "this should probably be in another crate. Maybe the current
> architecture needs another thing between the pure transport and the higher
> layers to support this, like cda-transport-orchestrator or something. As we
> now mix in DoIP into the CAN transport."

> alexmohr: "+1 for an intermediate layer it will make the implementation of
> the deferred mode easier if we can do this on a common layer instead of
> taking care of multiple transports"

## Proposal

Extract the orchestration into its own crate/layer between the pure
transports and the UDS layer (working title: transport orchestrator). Each
transport crate then only implements `EcuGateway` for its own bus; routing
policy, bindings and cross-transport concerns (and later the deferred mode)
live in one common place.

Follow-up from #323.

alexmohr:
## Summary

Extract multi-transport gateway from cda-comm-can into a dedicated cda-transport-orchestrator crate for better separation of concerns.

## Changes

- Add new `cda-transport-orchestrator` crate with `MultiTransportGateway` for routing diagnostic messages over multiple transports (DoIP and CAN)
- Add `TransportType` enum and UDS protocol module to interfaces for transport-agnostic communication
- Add `EcuCanGateway` trait and `CanComParams` for CAN-specific operations
- Move DoIP gateway error types from interfaces to cda-comm-doip
- Rename `com_param_handling` to `com_param` for consistency
- Update all dependent crates to use the new transport orchestrator

 
image
image
image

## Related Issue

Fixes https://github.com/eclipse-opensovd/classic-diagnostic-adapter/issues/442


cda-comm-doip/src/connection_receiver.rs
```diff
@@ -161,7 +140,7 @@ pub(super) async fn handle_response(
     }
 }
 
-/// Handles all variants of a successfully decoded `DiagnosticResponse`.
+/// Handles all variants of a successfully decoded [`DiagnosticResponse`].
```
theswiftfox, line 143: curiosity: was that a new clippy lint that we format it like this now in the docstring?

cda-comm-doip/src/connection_receiver.rs
```diff
@@ -182,22 +161,21 @@ async fn handle_ok_response(
...
-        DiagnosticResponse::Pending { source_address, .. }
-        | DiagnosticResponse::BusyRepeatRequest { source_address, .. }
-        | DiagnosticResponse::TemporarilyNotAvailable { source_address, .. } => {
-            outtx
-                .get(&source_address)
-                .map(|router| router.send(Ok(response)));
-        }
-        DiagnosticResponse::Msg(msg) => {
-            handle_diagnostic_message(gateway_name, outtx, msg).await;
+        DiagnosticResponse::Msg {
```
theswiftfox, line 164: q: do we no longer need to handle `DiagnosticResponse::Pending` and the others that are removed here? if so, should we document this in the docstring of the function? (currently states ' all variants handled'?

cda-comm-uds/src/lib.rs
```diff
@@ -311,7 +312,7 @@ impl<S: EcuGateway, T: EcuManager> cda_interfaces::Shutdown for UdsManager<S, T>
     }
 }
 
-impl<S: EcuGateway, T: EcuManager> UdsEcu for UdsManager<S, T> {}
+impl<S: EcuGateway + FunctionalTransport, T: EcuManager> UdsEcu for UdsManager<S, T> {}
```
theswiftfox: q: could this be made part of EcuGateway, such that EcuGateway implies FunctionalTransport?
`trait EcuGateway : FunctionalTransport { ... }`

cda-transport-router/src/lib.rs
```diff
@@ -0,0 +1,411 @@
...
+/// detected on (`DoIP` preferred); see the module docs for the full strategy.
+#[doc(alias = "MultiTransportGateway")]
+pub struct DiagnosticTransportRouter<
+    D: EcuGateway + FunctionalTransport + TransportProbe,
+    C: EcuGateway + TransportProbe,
+> {
+    /// Optional `DoIP` gateway
+    doip_gateway: Option<D>,
+    /// Optional CAN gateway
+    can_gateway: Option<C>,
```
theswiftfox: question: couldn't we implement functionalTransport simply with error returns for CAN? then we would avoid having this differentiation here? Could also add a `const fn functional_supported(&self) -> bool`  to the trait and decide on that if we want to create functional routes for that gateway or not?

cda-comm-doip/src/connection_receiver.rs
alexmohr: It's been a while already, but not that I can remember. But it improves to docstring as symbols then can be navigated and are checked.

cda-comm-doip/src/connection_receiver.rs
alexmohr: 'Pending' is no longer a diagnostic response. It has been moved into `TransportResponse` and is handled in `transport.rs` instead. The diagnostic response now only handles 'final' responses NACK/ACK/Message

cda-transport-router/src/lib.rs
alexmohr: I prefer the differentiation here instead of 'pretending' CAN supports functional communication to be honest and have this a compile time restriction until the CAN implementation actually supports this. Otherwise you either notice at runtime or you have to read the CAN implementation that the functional support is actually fake.

alexmohr: .... well actually .... it is already implemented huh... oh well then I won't remove this out of spite 😆 
Will update this as requested. 

```
impl FunctionalTransport for CanDiagGateway {
    fn send_functional(
        &self,
        _transmission_params: cda_interfaces::TransmissionParameters,
        _message: cda_interfaces::ServicePayload,
        _expected_ecu_logical_addrs: cda_interfaces::HashMap<u16, String>,
        _timeout: std::time::Duration,
        _expect_positive_response: bool,
    ) -> impl Future<
        Output = Result<
            cda_interfaces::HashMap<
                String,
                Result<cda_interfaces::ServicePayload, DiagServiceError>,
            >,
            DiagServiceError,
        >,
    > + Send {
        // CAN functional addressing is not implemented yet, see #417.
        // Fail the whole request honestly; the UDS layer maps a gateway-level
        // error to a per-ECU error result, so clients see WHY it failed
        // instead of every ECU appearing to be offline.
        std::future::ready(Err(DiagServiceError::RequestNotSupported(
            "functional addressing is not implemented for the CAN transport".to_owned(),
        )))
    }
}
```

alexmohr: But we still need two type arguments, as otherwise we could not have to different concrete implementation for CAN and DOIP

theswiftfox: hmm, that sounds like a case for a marker trait then maybe? if it is only used to differentiate the concrete type?
eg. `pub trait MarkerCAN {}` and `pub trait MarkerDOIP {}`, similar to how Send / Sync work
then we can just `impl MarkerCan for CanGateway {}` to mark a struct as compatible in those places where we need the distinction?
(open for different name suggestions, but wanted to get the idea across quickly :D)

cda-comm-uds/src/lib.rs
alexmohr: yes, done

cda-transport-router/src/lib.rs
alexmohr: added in the last commit, not 100% sure if that is what you wanted tbh.

cda-comm-can/src/lib.rs
```diff
@@ -137,3 +136,305 @@ impl cda_interfaces::ReusableTransportResource for CanDiagGateway {
...
+        fn send(
+            &self,
+            _transmission_params: TransmissionParameters,
+            _message: ServicePayload,
+            _response_sender: mpsc::Sender<
+                Result<Option<cda_interfaces::TransportResponse>, DiagServiceError>,
+            >,
+            _expect_uds_reply: bool,
+        ) -> impl Future<Output = Result<(), DiagServiceError>> {
+            std::future::ready(Ok(()))
```
theswiftfox: q: do we want to add a debug log or something that doip is not enabled. its fine if you see no benefit in it. I just thought about potential analysis scenarios :)

theswiftfox: wait nvm, this is the test module, i missed that (skimmed for the test cfg, but missed the test + can cfg xD). resolving this.

cda-comm-doip/src/lib.rs
```diff
@@ -63,30 +65,26 @@ mod vir_vam;
...
+    Msg {
         source_address: u16,
-        request_sid: u8,
+        target_address: u16,
+        data: Vec<u8>,
     },
     Ack((u16, Vec<u8>)),
     Nack(DiagnosticMessageNack),
     AliveCheckResponse,
+    /// `TesterPresent` NRC - intercepted at the decoding layer so the receiver
```
theswiftfox, line 89: nit: as we are documenting some fields of the enum, it would make sense to add a docstring to all of them maybe?

cda-comm-doip/src/lib.rs
```diff
@@ -769,6 +703,39 @@ impl<T: EcuAddresses + DoipComParams> EcuGateway for DoipDiagGateway<T> {
...
+        let Some(ecu_lock) = self.state.ecus.get(&ecu_name) else {
+            return RouteStatus::NotConfigured;
+        };
+        // Check if the gateway connection for this ECU is established
+        let gateway_addr = ecu_lock.read().await.logical_gateway_address();
+        if self.get_doip_connection(gateway_addr).await.is_ok() {
+            RouteStatus::Ready
+        } else {
+            // DoIP reachability is driven by VAM discovery and connection
+            // management; the router cannot repair it with a bounded ECU probe.
```
theswiftfox: q: is `the router cannot repair it with a bounded ECU probe.` something we want to have in the future? The comment reads a bit like this is something we have to do for now, as we are missing functionality?

cda-comm-doip/src/lib.rs
```diff
@@ -968,6 +933,41 @@ async fn read_ecu_responses(
...
+/// caller channel.
+///
+/// Pending NRCs are classified via [`pending_nrc_from_raw`]; all other frames
+/// are classified as final via [`uds_response_from_raw`].
+///
+/// `TesterPresentNRC` is absorbed by [`connection_receiver`] before it is
+/// broadcast to any per-ECU channel -- it never reaches this function.
+/// All other [`DiagnosticResponse`] variants (Ack, Nack, etc.) must be
+/// handled before calling this function.
+fn doip_event_to_transport(
```
theswiftfox: nit: `diagnostic_response_to_transport` maybe? or `diagnostic_response_event_to_transport`. Otherwise we are mixing quite a few names, with the docstring and param using the diagnostic response phrasing but the fn doip_event.

cda-comm-uds/src/transport.rs
```diff
@@ -263,57 +263,57 @@ impl<S: EcuGateway, T: UdsEcuDb + VariantDetection> UdsManager<S, T> {
...
+                                ) {
+                                    break 'read_uds_messages Err(e);
+                                }
+                                let sleep_time = uds_params.rc_94_repeat_request_time;
+                                tracing::debug!(
+                                    sleep_time = ?sleep_time,
+                                    "TemporarilyNotAvailable received, resending after delay"
+                                );
+                                cda_interfaces::util::tokio_ext::sleep_for(sleep_time).await;
+                                continue 'send; // continue 'send, will resend the message
```
theswiftfox: nit: while this nit itself is not introduced by the PR i just realized that both match blocks are 99% identical. thinking we could maybe do something like:
```rust
PendingNrc::BusyRepeatRequest { .. } | PendingNrc::TemporarilyNotAvailable { .. } => {
    let (policy, timeout, sleep_time, request) = match pending {
        PendingNrc::BusyRepeatRequest { .. } => (&uds_params.rc_21_retry_policy, &ds_params.rc_21_completion_timeout,  uds_params.rc_21_repeat_request_time, "BusyRepeatRequest".to_string()),
        PendingNrc::TemporarilyNotAvailable { .. } => (...)
        _ => unreachable!()
    };
	if let Err(e) = validate_timeout_by_policy(
	    ecu_name,
	    policy,
	    &start.elapsed(),
	    timeout,
	) {
	    break 'read_uds_messages Err(e);
	}

	tracing::debug!(
	    sleep_time = ?sleep_time,
	    "{request} received, resending after delay"
	);
	cda_interfaces::util::tokio_ext::sleep_for(sleep_time).await;
	continue 'send; // continue 'send, will resend the message
}
```
what do you think about that?

alexmohr: love it. fixed it.

cda-comm-doip/src/lib.rs
alexmohr: renamed to `diagnostic_response_to_transport`

cda-comm-doip/src/lib.rs
alexmohr: No, this mainly is due to the fact that we somehow have to unify CAN and DoIP. CAN does support probing single ECUs, but this is not possible with DoIP as is. 
The gateway connection should be re-established automatically already, so I would not trigger a re-connect here. 
Returning Unavailable means, that the router will start an ECU probe which is also only supported for CAN. 
I rephrased the comment to

```
            // DoIP addressing is gateway-scoped, so there is
            // no per-ECU probe to offer. Reconnecting is this transport's job
            // (VAM listener, connection-reset task), not the router's.
```

The only other viable option that I can come up with is to split the transport probe between CAN and DoIP but I don't think that made sense. Especially because we unified CAN and DoIP on the Gateway level already on the last PR.

cda-comm-uds/src/transport.rs
theswiftfox: ooh your version is even better! 🤩️

theswiftfox (approved):

PR #456, opened 2026-07-22
ci: make clippy findings blocking

### LINKED ISSUE #451: ci: make clippy findings blocking
The clippy workflow runs with `fail-on-clippy-error: false`, so findings are annotations only and a PR can merge with fresh clippy errors (came up in #323: an allow-without-reason slipped through). Flip the flag so clippy gates like fmt does. May need a one-time cleanup of existing findings per feature set (`--all-features` included).

alexmohr:
## Summary

This PR implements #451 by making clippy findings blocking in CI.

**Changes**:
- Changed `fail-on-clippy-error` from `"false"` to `"true"` in pr-checks.yml
- Updated pinned nightly toolchain from `nightly-2025-07-14` to `nightly-2026-07-21` in all workflow files
- Added `cfg_attr` annotations to suppress `duration_suboptimal_units` lint on nightly (not available in Rust 1.88 MSRV)
- Fixed `doc_markdown` errors in integration tests

**Testing**:
- Verified with `cargo +1.88 clippy --all-targets` - no warnings
- Verified with `cargo +nightly-2026-07-21 clippy --all-targets --all-features` - passes on PR code

**Note**: The new nightly toolchain detects additional pre-existing issues in `cda-interfaces/src/ecuuds.rs` (unused_async_trait_impl). 

## Checklist

## Related

Fixes #451

## Notes for Reviewers

The nightly toolchain update (2025-07-14 → 2026-07-21) introduces new lints that may detect pre-existing issues. The `duration_suboptimal_units` lint is suppressed with `cfg_attr` annotations since `from_mins`/`from_hours` are not available in Rust 1.88 MSRV.

Reason for making these failure mandatory: It's a _fixed_ nightly version, no surprises will happen. The optional clippy findings where too invisible and even regular Contributors missed them. 
Updating the nighlty version so something more recent to get more analytics. 

---

### DIFF
--- a/cda-comm-uds/src/test_helpers.rs
+++ b/cda-comm-uds/src/test_helpers.rs
@@ -13,7 +13,7 @@
 
 //! Shared test doubles for `cda-comm-uds` tests.
 
-use std::time::Duration;
+use std::{future::Future, time::Duration};
 
 use async_trait::async_trait;
 use cda_interfaces::{
@@ -164,24 +164,24 @@ impl EcuStateManager for TestEcuDb {
         async move { states.lock().await.get(&sid).cloned() }
     }
 
-    async fn session(&self) -> Result<String, DiagServiceError> {
-        Ok("default".to_string())
+    fn session(&self) -> impl Future<Output = Result<String, DiagServiceError>> + Send {
+        std::future::ready(Ok("default".to_string()))
     }
 
     fn default_session(&self) -> Result<String, DiagServiceError> {
         Ok("default".to_string())
     }
 
-    async fn security_access(&self) -> Result<String, DiagServiceError> {
-        Ok("locked".to_string())
+    fn security_access(&self) -> impl Future<Output = Result<String, DiagServiceError>> + Send {
+        std::future::ready(Ok("locked".to_string()))
     }
 
     async fn lookup_session_change(&self, _session: &str) -> Result<DiagComm, DiagServiceError> {
         unimplemented!()
     }
 
-    async fn set_default_states(&self) -> Result<(), DiagServiceError> {
-        Ok(())
+    fn set_default_states(&self) -> impl Future<Output = Result<(), DiagServiceError>> + Send {
+        std::future::ready(Ok(()))
     }
 }
 
--- a/cda-comm-uds/src/test_helpers.rs
+++ b/cda-comm-uds/src/test_helpers.rs
@@ -13,7 +13,7 @@
 
 //! Shared test doubles for `cda-comm-uds` tests.
 
-use std::{future::Future, time::Duration};
+use std::time::Duration;
 
 use async_trait::async_trait;
 use cda_interfaces::{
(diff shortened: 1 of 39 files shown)


cda-comm-uds/src/test_helpers.rs
theswiftfox, line 1: question: is it needed to manually async the functions here for clippy? (also again we have the future::Future import ^^ that should not be needed)

cda-comm-uds/src/transport.rs
```diff
@@ -689,6 +689,7 @@ mod tests {
 #[cfg(test)]
 mod send_tests {
     use std::{
+        future::Future,
```
theswiftfox: nit: unnecessary import (should be part of preulde?)

cda-interfaces/src/ecumanager.rs
```diff
@@ -11,10 +11,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-use std::{fmt::Display, future::Future};
+use std::future::Future;
```
theswiftfox: nit: prelude import

theswiftfox (commented):
i only commented on one occurance, but there are quite a lot of changes that turn a `async fn` into manually async `fn (...) -> Future<...>`. Is that required by clippy? I thought it usually complains about manual async impls.

alexmohr:
> i only commented on one occurance, but there are quite a lot of changes that turn a `async fn` into manually async `fn (...) -> Future<...>`. Is that required by clippy? I thought it usually complains about manual async impls.

Example:
```
  fn session(&self) -> impl Future<Output = Result<String, DiagServiceError>> + Send {
        std::future::ready(Ok("default".to_string()))
    }
```

This is because clippy _will_ complain when you have an async function without an await. But because we have to adhere to the trait contract we just can't drop it entirely. Using `future::ready` like above is the solution clippy itself suggests. See https://rust-lang.github.io/rust-clippy/master/index.html#unused_async_trait_impl . We could disable the lint, but as it creates computational overhead *not* to use `ready` I preferred changing the implementations.

cda-comm-uds/src/test_helpers.rs
alexmohr: https://github.com/eclipse-opensovd/classic-diagnostic-adapter/pull/456#issuecomment-5166241399

theswiftfox:
> > i only commented on one occurance, but there are quite a lot of changes that turn a `async fn` into manually async `fn (...) -> Future<...>`. Is that required by clippy? I thought it usually complains about manual async impls.
> 
> Example:
> 
> ```
>   fn session(&self) -> impl Future<Output = Result<String, DiagServiceError>> + Send {
>         std::future::ready(Ok("default".to_string()))
>     }
> ```
> 
> This is because clippy _will_ complain when you have an async function without an await. But because we have to adhere to the trait contract we just can't drop it entirely. Using `future::ready` like above is the solution clippy itself suggests. See https://rust-lang.github.io/rust-clippy/master/index.html#unused_async_trait_impl . We could disable the lint, but as it creates computational overhead _not_ to use `ready` I preferred changing the implementations.

yeah i missed the part that it was only for tests. so we can't simply remove the async requirement from the trait.
thanks for the explainer / example :)

theswiftfox (approved):
LGTM. can merge if the pipeline agrees ;)

PR #464, opened 2026-07-27
Refactor: Code cleanup

mbfm:
## Summary

General code cleanups that came out of my work on moving the configuration into a separate crate.

In particular, the error handling rework makes a first step towards being able to see source errors, which were so far at best visible through logging.

This does not implement backtraces, which would be a separate step.

## Checklist

## Notes for Reviewers

Most of the Rust ecosystem seems to use `anyhow` to print the source errors. Since we don't currently have it as a dependency, I opted for implementing that printing manually.  
I believe, `anyhow` would also give backtraces for free within `cda-main`. Not sure, if that's worth the dependency then...

---

### DIFF
--- a/cda-database/src/datatypes/mod.rs
+++ b/cda-database/src/datatypes/mod.rs
@@ -433,6 +433,12 @@ impl Parameter<'_> {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct LongName {
+    pub value: Option<String>,
+    pub ti: Option<String>,
+}
+
 impl From<dataformat::LongName<'_>> for LongName {
     fn from(val: dataformat::LongName<'_>) -> Self {
         LongName {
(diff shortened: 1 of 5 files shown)


cda-database/src/datatypes/mod.rs
```diff
@@ -442,22 +448,6 @@ impl From<dataformat::LongName<'_>> for LongName {
     }
 }
 
-#[self_referencing]
```
alexmohr, line 445: question: Why the or-ordering here?

mbfm: I had initially looked at the `DiagnosticDatabase` as a place to collect the `ComParams` and I just found this code file needlessly difficult to read. The `impl` blocks were in a different place than the struct definitions. And the struct definitions were not logically grouped either, so you just ended up scrolling up and down all the time.

That's all there is to it. It's not a particularly riveting change. 😅

mbfm:
Just realized, I didn't follow Conventional Commits message style for the first commit. Will force-push.

cda-database/src/datatypes/mod.rs
alexmohr: oh ya that makes sense, thank you. It just wasn't evident from the diff alone 👍

alexmohr (approved):

PR #466, opened 2026-07-27
fix(update): make the update plugin actually pluggable

alexmohr:
## Summary

Prior to this commit the update plugin was not plugable
without rewriting the app init logic,
as setup_vehicle_and_routes hard coded the default plugin.

The init now gains a builder pattern where new plugins
be added easily later.

## Checklist

## Notes for Reviewers

Commits will be squashed when merging, they are kept verbose right now on purpose to make the history clear, as this was partially reviewed already.

### DIFF
--- a/cda-interfaces/Cargo.toml
+++ b/cda-interfaces/Cargo.toml
@@ -38,6 +38,7 @@ thiserror = { workspace = true }
 async-trait = { workspace = true }
 #tokio
 tokio = { workspace = true, features = ["sync", "time"] }
+futures = { workspace = true }
 # runtime size info
 hex = { workspace = true }
 strum = { workspace = true }
--- a/cda-main/src/lib.rs
+++ b/cda-main/src/lib.rs
@@ -266,28 +266,11 @@ where
     };
 
     let (mut config, disk_loaded) = config::load_config_with_fallback(config_file);
-
-    match cda_storage::LocalStorage::new(&config.runtime_update_config.storage_dir) {
-        Ok(storage) => {
-            if config.runtime_update_config.init_storage_from_config_file {
-                config::seed_storage_from_config_file(&storage, config_file).await?;
-            }
-
-            if let Some(storage_config) =
-                config::load_config_with_storage_override(&storage).await?
-            {
-                config = storage_config;
-            } else if !disk_loaded {
-                config::require_config_source()?;
-            }
-        }
-        Err(e) => {
-            tracing::warn!(error = %e, "Storage not available, skipping seed and config override");
-            return Ok(());
-        }
+    if !disk_loaded {
+        config::require_config_source()?;
     }
 
-    // Command line arguments always take precedence over stored configuration
+    // Command line arguments always take precedence over file configuration.
     args.update_config(&mut config);
 
     config.validate_sanity().map_err(AppError::from)?;
--- a/cda-main/src/lib.rs
+++ b/cda-main/src/lib.rs
@@ -31,7 +31,7 @@ use cda_core::EcuManager;
 use cda_database::FileManager;
 use cda_interfaces::{
     EcuConnectivityHandler, FunctionalDescriptionConfig, HashMap, HashMapExtensions, UdsQuery,
-    UdsVariant, config::ConfigSanity, datatypes::FaultConfig, dlt_ctx,
+    UdsVariant, config::ConfigSanity, datatypes::FaultConfig, dlt_ctx, health::HealthProvider,
 };
 use cda_plugin_security::{
     DefaultSecurityPlugin, DefaultSecurityPluginData, SecurityPlugin, SecurityPluginLoader,
@@ -39,21 +39,25 @@ use cda_plugin_security::{
 use cda_sovd::Locks;
 use cda_tracing::{OtelGuard, TracingSetupError, TracingWorkerGuard};
 use clap::{Parser, Subcommand};
-pub use error::AppError;
-use futures::future::FutureExt;
 use tokio::sync::{Mutex, RwLock, mpsc};
 use tracing_subscriber::layer::SubscriberExt;
 
 use crate::{
-    config::configfile::Configuration,
+    config::{configfile::Configuration, generate::generate_config_cmd},
     mdd::{load_databases, resolve_mdd_paths},
-    update::{RuntimeUpdateContext, security::UpdateSecurityHandler},
+    setup::PreLoadHook,
+    update::{UpdatePluginBuilder, create_default_update_plugin, update_plugin_fn},
 };
 
+pub mod cda_factory;
 pub mod config;
 pub mod error;
 pub mod mdd;
+pub mod setup;
 pub mod update;
+
+pub use error::AppError;
+pub use setup::Setup;
 
 #[global_allocator]
 static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;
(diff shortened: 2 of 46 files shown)


alexmohr:
Added @theswiftfox as reviewer, because she reviewed the old version of this PR

cda-interfaces/Cargo.toml
```diff
@@ -37,7 +37,8 @@ serde_json = { workspace = true }
 thiserror = { workspace = true }
 async-trait = { workspace = true }
 #tokio
-tokio = { workspace = true, features = ["sync", "time"] }
+tokio = { workspace = true, features = ["sync", "time", "rt"] }
```
theswiftfox: q: is the "rt" feature needed here? I tested it locally without it and cargo-check did not complain

cda-main/src/update.rs
```diff
@@ -10,242 +10,188 @@
...
+    }
 }
 
-pub struct DefaultRuntimeFileReloadHandler<S, F, P>
+/// Concrete [`VehicleComponentFactory`] that delegates to
+/// [`crate::create_vehicle_components`].
+///
+/// Used by [`DefaultRuntimeReloaderPlugin`]
+/// and called every time the diagnostic databases are reloaded.
+pub struct CdaMainVehicleFactory<SP>
```
theswiftfox: q: is this really the right place for the factory struct or should that live in either cda-main/src/lib.rs or cda-main/src/cda_factory.rs?

cda-sovd/src/sovd/update_guard.rs
```diff
@@ -73,6 +65,23 @@ impl UpdateGuardState {
...
+        self.busy.load(Ordering::Acquire)
+    }
+
+    /// Returns a shared handle to the busy flag.
+    #[must_use]
+    pub fn busy_handle(&self) -> Arc<AtomicBool> {
+        Arc::clone(&self.busy)
+    }
+
+    /// Returns a shared reference to the exempt routes list.
```
theswiftfox: nit: this is not a shared reference, but `a snapshot of the exempt_routes at the time of calling this function`

theswiftfox (commented):
first few comments (very generic). I'll have to take a deeper look at all the new wiring and setup code, so there will probably be a second set of comments with a bit more 'meat' once i'm through that :D

alexmohr:
FIxed the review findings. A **very(!)** rough draft for the discussed "update registry" is here https://github.com/eclipse-opensovd/classic-diagnostic-adapter/commit/095252d5e66608b1a1bb8afb5f7e1f5938461784 

We will implement this after the deferred communication, as the two other PRs already building on this implement the suggestion partially anyway.

cda-plugin-runtime-update/src/default_runtime_reloader_plugin.rs
```diff
@@ -0,0 +1,417 @@
...
+            functional_group_config,
+        } = self
+            .factory
+            .create(
+                &cfg,
+                &mdd_paths,
+                self.update_guard.busy_handle(),
+                reusable_transport_resource,
+            )
+            .await?;
```
theswiftfox: q: if this fails the cda is in a degraded state, right? is there any way to recover to the 'old' runtime from here? (thinking scenarios where an udpate causes an issue that is also not resolved by 'full reboot')

cda-main/src/lib.rs
```diff
@@ -314,59 +292,149 @@ where
...
+pub async fn run_with_ext_from_config<SP, SL, UPB>(
+    config: Configuration,
+    setup: Setup<SP, SL, UPB>,
+) -> Result<(), AppError>
+where
+    SP: SecurityPlugin,
+    SL: SecurityPluginLoader,
+    UPB: UpdatePluginBuilder<SP>,
+{
+    let webserver_state = init_webserver(&config, setup.pre_load).await?;
```
theswiftfox, line 312: note: we should ensure that any early return here first ensures the webserver is joined. if we simply return from this function the webserver_state is dropped and hence the join handle simply detached, meaning the server could continue to run.

alexmohr: implemented 'drop' for the webserver state. Imho the cleanest, even though we cannot await the task in `drop`. Alternative would have been calling a cleanup in every branch.

cda-plugin-runtime-update/src/default_runtime_reloader_plugin.rs
alexmohr: added a rollback mechanism for this.

cda-main/src/lib.rs
theswiftfox: one option would be to wrap everything after creating of the webserver state into a fn or closure that returns a result and then match on that and if it is err abort the task before dropping. then it would allow to await on it

cda-main/src/lib.rs
```diff
@@ -732,28 +643,61 @@ pub fn create_uds_manager<S: SecurityPlugin>(
...
-    pub doip: Arc<cda_health::StatusHealthProvider>,
-    pub database: Arc<cda_health::StatusHealthProvider>,
+/// Collected webserver state produced by [`init_webserver`].
+///
+/// Passed to [`setup::setup_runtime_routes`] and [`run_with_ext_from_config`] so that
+/// the shutdown signal and health provider are accessible after the webserver is started.
+///
+/// Dropping this value aborts the webserver task, so all error paths are covered
+/// automatically without any explicit cleanup calls.
+pub(crate) struct WebserverState {
```
theswiftfox: note: ideally this struct would be in cda-sovd, as that should be the place in relation to webserver etc.

alexmohr:
99e33405476b654717a201ffc997b898fcb7cb80 disables the configuration update as discussed. It disables a bit more than only returning an error on the http layer but does not remove it entirely either. I will not squash this commit later, so we have a reference point when reparing this feature. Also created https://github.com/eclipse-opensovd/classic-diagnostic-adapter/issues/497

cda-main/src/lib.rs
alexmohr: Not sure if I can agree with this but mainly due to `main_health_provider` being used as the health status of "main". I'm not sure if it's a great idea to move it as it cda-sovd. 
What we could though is to split this struct

```rust
// SOVD:
struct WebServerRuntime { // state already exists in cda-sovd
    _tracing_guards: TracingGuards,
    pub dynamic_router: cda_sovd::dynamic_router::DynamicRouter,
    webserver_task: Option<tokio::task::JoinHandle<()>>,
    pub shutdown_signal: cda_interfaces::ShutdownSignal,
    health_state: Option<cda_health::HealthState>,
}

// Main:
struct WebServerContext { // should not clash with 'state' or runtime
    main_health_provider: Option<Arc<cda_health::StatusHealthProvider>>,
    runtime: WebServerRuntime
}
```

Would that work for you? (I'm open to alternative naming)

theswiftfox: oh didn't see that with health_state. god everything is so coupled now 🙈️
in that case you can also keep it in cda-main, but maybe we should name it ApplicationState or SovdState or something? Reading WebserverState immediately made my mind jump to the conclusion that it should be inside cda-sovd ^^

alexmohr: arguably the coupling isn't much worse than it is on main. Although the runtime update plugin itself did make it worse. When we are implementing the update registry it might get a bit better but eventually, I'll look into that when I am solving this. Renamed to ApplicationState

theswiftfox (approved):
lgtm now i think :) thank you!

PR #469, opened 2026-07-28
fix(cda-core): match generic service by full request prefix, not SID alone

floroks:
## Summary

check_genericservice (used by the /genericservice raw UDS passthrough endpoint) only matched services by their SID byte and then took the first match. For write-style requests such as WriteDataByIdentifier (0x2E xx yy), multiple DB service entries commonly share the same SID (one per DID), so matching on the SID alone picked an arbitrary same-SID service instead of the one actually addressed by the raw payload.

This caused the access/precondition check to be evaluated against the wrong service, and any resulting error to report that wrong service's name instead of the one the client actually invoked.

Fix by reusing the same full request-prefix matching already used by lookup_diagcomms_by_request_prefix (which compares the sequential coded-constant bytes - SID, sub-function, DID, etc. - against the raw payload), extracted into a shared DiagService::matches_request_prefix helper.

Adds a regression test with two WriteDataByIdentifier services sharing SID 0x2E but different DIDs, one gated behind a precondition, verifying the correct service is matched and reported.

## Checklist

## Notes for Reviewers

### DIFF
--- a/cda-database/src/datatypes.rs
+++ b/cda-database/src/datatypes.rs
@@ -273,6 +273,58 @@ impl DiagService<'_> {
 
         result
     }
+
+    #[must_use]
+    /// Check whether this service's sequential coded-constant parameters (SID,
+    /// sub-function, DID, etc.) match the given raw request bytes.
+    ///
+    /// This compares every coded-constant parameter of the service (as returned
+    /// by [`Self::extract_sequential_coded_consts`]) against the corresponding
+    /// bytes in `request_bytes`. If `request_bytes` is shorter than a given
+    /// parameter's expected position, the match is still considered successful
+    /// for the bytes provided so far (partial prefixes are accepted).
+    ///
+    /// This is used to disambiguate between multiple services that share the
+    /// same SID (e.g. several `WriteDataByIdentifier` service entries, one per
+    /// DID, all sharing SID `0x2E`), by also comparing the bytes following the
+    /// SID (such as the DID) rather than only the first byte.
+    pub fn matches_request_prefix(&self, request_bytes: &[u8]) -> bool {
+        let mut byte_idx = 0usize;
+        for param in self.extract_sequential_coded_consts() {
+            let param_byte_count = param.byte_count();
+            if param_byte_count > 4 {
+                return false;
+            }
+            let Some(end_idx) = byte_idx.checked_add(param_byte_count) else {
+                return false;
+            };
+            // Ran out of caller-provided bytes, all provided bytes matched, accept
+            if end_idx > request_bytes.len() {
+                return true;
+            }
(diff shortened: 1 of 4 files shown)


cda-database/src/datatypes.rs
```diff
@@ -273,6 +273,58 @@ impl DiagService<'_> {
...
+    ///
+    /// This is used to disambiguate between multiple services that share the
+    /// same SID (e.g. several `WriteDataByIdentifier` service entries, one per
+    /// DID, all sharing SID `0x2E`), by also comparing the bytes following the
+    /// SID (such as the DID) rather than only the first byte.
+    pub fn matches_request_prefix(&self, request_bytes: &[u8]) -> bool {
+        let mut byte_idx = 0usize;
+        for param in self.extract_sequential_coded_consts() {
+            let param_byte_count = param.byte_count();
+            if param_byte_count > 4 {
```
theswiftfox, line 295: q: why do we abort on byte_count > 4?

cda-database/src/datatypes.rs
```diff
@@ -273,6 +273,58 @@ impl DiagService<'_> {
...
+                return false;
+            }
+            let Some(end_idx) = byte_idx.checked_add(param_byte_count) else {
+                return false;
+            };
+            // Ran out of caller-provided bytes, all provided bytes matched, accept
+            if end_idx > request_bytes.len() {
```
theswiftfox, line 302: q: if the caller provides 0 bytes this would always match? should we at least expect **some** bytes from the caller? (can be a check before entering the loop)

cda-database/src/datatypes.rs
floroks: That was taken over from the previous code in `diag_comm_lookup.rs‎` - in the comparison later it converts the buffer to a u32, so this limits the maximum length to 32bits.

We probably need to change that in the future, I created https://github.com/eclipse-opensovd/classic-diagnostic-adapter/issues/475 to track this, but it should be very rare to have more than 4 bytes defined as constant part, since uds requests (without payload) will be at max 4 bytes, only if you start to include payload as a constant you'll run into this.

cda-database/src/datatypes.rs
floroks: yes, but 0 length requests are already rejected by the genericservice, so it has no effect, and from a security perspective, permitting to sending an empty request (so no request) is also not critical

Check in `check_genericservice`:
 ```rust
let raw_data_sid = rawdata.first().copied().ok_or_else(|| {
    DiagServiceError::BadPayload("Expected at least 1 byte to read SID".to_owned())
})?;
```

floroks:
rebased due to merge conflicts

theswiftfox (approved):

PR #471, opened 2026-07-29
refactor(sovd): move diagnostic DB update executions to /operations/runtimefilesupdate & allow octet-stream 

floroks:
## Summary

**refactor(sovd): move diagnostic DB update executions to /operations/runtimefilesupdate**

Moves the Apply/Rollback/Cleanup execution endpoints for the diagnostic
database update out of the bulk-data resource
(/apps/sovd2uds/bulk-data/runtimefiles-nextupdate/executions) into a
proper operations resource
(/apps/sovd2uds/operations/runtimefilesupdate/executions), for standard
SOVD operations compliance and consistent naming.

- Add new cda-sovd operations::runtimefilesupdate module hosting the
    executions list/start/status handlers, moved out of
    bulk_data/runtimefiles.rs (which now only handles
    runtimefiles-current/-nextupdate/-backup file management).
- Remove the old literal /operations/diagnostic-database-update alias
    route; the new path is now the sole endpoint.
- Wrap the execution mode in a "parameters" object in the request body
    ({"parameters": {"mode": "apply"}}), matching the standard
    convention used by other operation executions (comparams,
    functional-group, ECU-service operations).
- Update architecture docs and integration tests accordingly.

**feat(sovd): accept application/octet-stream uploads for runtime files**

The runtimefiles-nextupdate upload endpoint now accepts a single file
as a raw application/octet-stream body, in addition to the existing
multipart/form-data support. For octet-stream uploads, the filename
must be provided via the Content-Disposition header (e.g.
'attachment; filename="foo.mdd"'), supporting both the quoted and
unquoted filename= forms. Requests with an unsupported Content-Type,
or octet-stream requests missing a usable filename, are rejected with
400 Bad Request.

- Split nextupdate::post into content-type-dispatching handlers
    (handle_multipart_upload / handle_octet_stream_upload).
- Add a small hand-rolled Content-Disposition filename parser (no
    crate in the dependency tree currently exposes this).
- Update architecture docs and add integration tests covering the
    success path (quoted and unquoted filename) and the two rejection
    cases (missing header, missing filename parameter).

## Checklist

## Notes for Reviewers

### DIFF
--- a/cda-plugin-runtime-update/src/storage.rs
+++ b/cda-plugin-runtime-update/src/storage.rs
@@ -108,6 +108,7 @@ pub(crate) async fn list_collection_files(
         let mut item = BulkDataDescriptor {
             id: key.clone(),
             mimetype: mime_for_key(key),
+            name: Some(key.clone()),
             size: None,
             hash: None,
             hash_algorithm: None,
@@ -257,7 +258,18 @@ pub(crate) async fn delete_nextupdate_file(
 
 pub(crate) async fn delete_all_nextupdate(
     storage: &impl Storage,
-) -> Result<(), RuntimeUpdateError> {
+) -> Result<Vec<String>, RuntimeUpdateError> {
+    let mut deleted_ids = Vec::new();
+    for collection_name in [
+        CollectionName::DiagnosticDatabaseNextUpdate,
+        CollectionName::ConfigurationNextUpdate,
+    ] {
+        match storage.get_collection(&collection_name).await {
+            Ok(collection) => deleted_ids.extend(collection.list().await?),
+            Err(StorageError::CollectionNotFound(_)) => {}
+            Err(e) => return Err(RuntimeUpdateError::from(e)),
+        }
+    }
     let mut tx = storage.begin_transaction()?;
 
     match storage
@@ -277,7 +289,7 @@ pub(crate) async fn delete_all_nextupdate(
     }
 
     tx.commit().await?;
-    Ok(())
+    Ok(deleted_ids)
 }
 
 /// Returns `true` if both backup collections (MDD and configuration) are empty or do not exist.
@@ -293,7 +305,9 @@ pub(crate) async fn is_backup_empty(storage: &impl Storage) -> Result<bool, Runt
     Ok(mdd_backup.is_empty().await? && cfg_backup.is_empty().await?)
 }
 
-pub(crate) async fn delete_all_backup(storage: &impl Storage) -> Result<(), RuntimeUpdateError> {
+pub(crate) async fn delete_all_backup(
+    storage: &impl Storage,
+) -> Result<Vec<String>, RuntimeUpdateError> {
     let mdd_backup = storage
         .get_or_create_collection(&CollectionName::DiagnosticDatabaseBackup)
         .await?;
@@ -302,10 +316,12 @@ pub(crate) async fn delete_all_backup(storage: &impl Storage) -> Result<(), Runt
         .await?;
 
     let mut tx = storage.begin_transaction()?;
+    let mut deleted_ids = mdd_backup.list().await?;
+    deleted_ids.extend(cfg_backup.list().await?);
     mdd_backup.delete_all(&mut tx).await?;
     cfg_backup.delete_all(&mut tx).await?;
     tx.commit().await?;
-    Ok(())
+    Ok(deleted_ids)
 }
 
 /// Computes the "next update" state from the pending `NextUpdate` collections, falling back to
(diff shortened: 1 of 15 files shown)


cda-sovd/src/sovd/apps/sovd2uds/bulk_data/runtimefiles.rs
```diff
@@ -292,6 +438,59 @@ pub(crate) mod nextupdate {
...
+            &*route_state.vehicle_lock_states,
+            *claims,
+            route_state.retry_after_seconds,
+        )
+        .await
+        {
+            // As with the multipart path: drain the (potentially large) body before
+            // responding, so an early rejection doesn't race the client's in-flight
+            // write and cause a transport-level error instead of a clean response.
+            let _ = axum::body::to_bytes(request.into_body(), usize::MAX).await;
```
theswiftfox: q: do we want / need to limit the upload size?

cda-sovd/src/sovd/apps/sovd2uds/bulk_data/runtimefiles.rs
```diff
@@ -245,11 +240,162 @@ pub(crate) mod nextupdate {
...
+    ///
+    /// Supports both the quoted form (`filename="foo.mdd"`) and the unquoted form
+    /// (`filename=foo.mdd`). Returns `None` if the header is missing, or if no
+    /// non-empty `filename` parameter can be found.
+    fn parse_content_disposition_filename(headers: &HeaderMap) -> Option<String> {
+        headers
+            .get(axum::http::header::CONTENT_DISPOSITION)
+            .and_then(|v| v.to_str().ok())
+            .and_then(|value| {
+                value.split(';').find_map(|part| {
```
theswiftfox: suggestion: this is very theoretical but if we unconditionally split at ';' we could run into an issue if a quoted filename uses ';' inside for some reason. 
it could make sense to move this into a helper fn and hhave a bit more complex logic there like tracking if it is inside quotes or so.

cda-sovd/src/sovd/apps/sovd2uds/bulk_data/runtimefiles.rs
floroks: there's a configuration option `upload_body_limit_bytes` in the plugin configuration, set by default to 50MB which limits the request body and should be sufficient.

theswiftfox: ooh i did miss that on. but shouldn't we use `let _ = axum::body::to_bytes(request.into_limited_body(), usize::MAX).await?;` instead then to enforce that limit?

floroks: can't hurt - I changed it to `into_limited_body`, and fixed an issue where the body wasn't drained with invalid content-dispositions

cda-sovd/src/sovd/apps/sovd2uds/bulk_data/runtimefiles.rs
floroks: I added a function `split_content_disposition_parameters` that is aware of quotes for splitting, and a test case for `"attachment; filename="database;backup.mdd"`

floroks:
rebased due to conflicts

cda-plugin-runtime-update/src/storage.rs
```diff
@@ -257,7 +258,18 @@ pub(crate) async fn delete_nextupdate_file(
...
+) -> Result<Vec<String>, RuntimeUpdateError> {
+    let mut deleted_ids = Vec::new();
+    for collection_name in [
+        CollectionName::DiagnosticDatabaseNextUpdate,
+        CollectionName::ConfigurationNextUpdate,
+    ] {
+        match storage.get_collection(&collection_name).await {
+            Ok(collection) => deleted_ids.extend(collection.list().await?),
+            Err(StorageError::CollectionNotFound(_)) => {}
+            Err(e) => return Err(RuntimeUpdateError::from(e)),
```
theswiftfox, line 270: question: wouldn't we want to collect them into errors, so we can return them as part of the errors in the BulkDataDelete response?

floroks: We've currently defined all-or-nothing semantics for all of these endpoints, since that is the most useful one in practice (e.g. when deletion of single elements fails, you usually can't do anything anyway). Changing this would require adapting the specification and transaction around each individual delete, which I currently do not see the need for.

theswiftfox (approved):

PR #473, opened 2026-07-30
docs: add heap profiling guide

floroks:
## Summary

Adds instructions on profiling the cda.

## Checklist

## Notes for Reviewers

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -175,6 +175,9 @@ opt-level = 3
 [profile.release-with-debug]
 inherits = "release"
 debug = true
+opt-level = 3
+codegen-units = 1
+strip = false
 
 [patch.crates-io]
 # Necessary because upstream is still using an older version of serde_qs (0.14 instead of 1.0.0-rc.3)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -177,6 +177,9 @@ opt-level = 3
 [profile.release-with-debug]
 inherits = "release"
 debug = true
+opt-level = 3
+codegen-units = 1
+strip = false
 
 [patch.crates-io]
 # Necessary because upstream is still using an older version of serde_qs (0.14 instead of 1.0.0-rc.3)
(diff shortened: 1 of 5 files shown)


cda-main/src/lib.rs
```diff
@@ -51,6 +51,10 @@ pub mod error;
 pub mod mdd;
 pub mod update;
 
+// Heaptrack and Valgrind intercept the system allocator, whereas mimalloc
+// manages allocations internally. Keep mimalloc in normal builds but omit it
+// from profiling builds so allocation call stacks remain visible.
+#[cfg(not(feature = "heap-profiling"))]
```
theswiftfox, line 57: note: at least heaptrack should support mimalloc (https://github.com/microsoft/mimalloc/issues/522#issuecomment-1079451649)

Cargo.toml
```diff
@@ -174,6 +174,10 @@ opt-level = 3
 [profile.release-with-debug]
 inherits = "release"
 debug = true
+# Prioritize useful allocation call stacks over maximum runtime performance.
+# This profile is intended for Heaptrack and similar startup-memory analysis.
+opt-level = 1
```
theswiftfox: question: did we check if this doesn't skew the produced picture? changing opt-level can result in different inlining and especially reduce allocation elimination in lower opt levels afaik.

while codegen-units can also change allocation behavior I'd assume it should be less noticeable than the general opt-level.

i'd suggest we use something like this to have it closer to the 'original' produced by the release build:

```toml
opt-level = 3
codegen-units = 1
strip = false
```

cda-main/src/lib.rs
floroks: changed the description, it's probably still a good thing to disable it for other profiling tools which do not support it.

Cargo.toml
floroks: changed as suggested

alexmohr (approved):

PR #477, opened 2026-07-31
fix(storage): retain current collection when apply rename fails

floroks:
## Summary

A failed collection-copy rename could leave a pre-existing current collection without a backup. Rollback then treated it as newly created and removed it, causing subsequent current requests to return no files.

Record whether a copy destination existed before the transaction and use that state during rollback and recovery so untouched current collections are preserved.

## Checklist

## Notes for Reviewers


theswiftfox (approved):
lgtm

Issue #479, opened 2026-07-31
[Bug] Seed config overwrites `Storage` on every startup


mbfm:
## Description

When the `init_storage_from_config_file` flag is set, we currently override what is written into `Storage` on every startup with the configuration:

https://github.com/eclipse-opensovd/classic-diagnostic-adapter/blob/dcfb72b2351631b46a933f9d73b1d1865dee498e/cda-main/src/lib.rs#L275-L280

The same will be true respectively for MDD files once #478 is fixed.

This means that when users use the Update-plugin to change the configuration or the MDD files, these changes get lost when CDA is restarted.

## Proposed Solution

* [x] The configuration file and MDD files passed via CLI should be treated exclusively for seeding. This means that when the `Storage` already contains them, they will not be loaded into `Storage`.  

* [x] The command-line flags should be renamed to make this behaviour more obvious:
  * `--seed-config`
  * `--seed-databases-path`

* [ ] To allow for changing the configuration via the configuration file, e.g. when testing, we could offer separate flags:
  * [x] `--force-seed-config`
  * [x] Fix MDD files not being seeded in the first place.
  * [ ] `--force-seed-databases`

## Additional Context

This was discussed with @alexmohr.

> [!IMPORTANT]
> Link the pull request that addresses this bug once created.

#480

mbfm:
@alexmohr realized that this actually is correctly handled already, by this condition:

https://github.com/mbfm/classic-diagnostic-adapter/blob/57f9cbff669fedf2e89297a86b585a59c8c3e3e8/cda-storage/src/storage_seed.rs#L37-L51

Will need to adjust the scope of this issue. Renaming the CLI flags to reflect this behavior still seems like it would be a good idea, since not even the devs knew what the current behavior is.

mbfm:
The config seeding is now also obsolete, since config hot reloading has been removed.

PR #482, opened 2026-08-04
Code cleanup from work on #479

mbfm:
## Summary

Smaller code cleanup changes surrounding configuration loading/seeding.

## Checklist

## Related

#479

## Notes for Reviewers

---

### DIFF
--- a/cda-main/src/config.rs
+++ b/cda-main/src/config.rs
@@ -10,6 +10,8 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  */
+use std::path::Path;
+
 use cda_interfaces::storage_api::{Collection, RandomAccessData, Storage};
 use figment::{
     Figment,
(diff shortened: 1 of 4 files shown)


cda-main/src/config.rs
```diff
@@ -91,44 +82,49 @@ pub fn require_config_source() -> Result<(), crate::AppError> {
...
-        Err(e) => {
-            tracing::warn!(error = %e, config_file_path, "Cannot read config file for seeding");
-            return;
-        }
-    };
+///
+/// # Errors
+/// Returns [`AppError`](crate::AppError) when no configuration file
+/// is inaccessible.
+pub async fn seed_storage_from_config_file(
```
alexmohr, line 89: suggestion: 
```suggestion
pub async fn seed_storage_if_empty_from_config_file(
```

cda-main/src/config.rs
```diff
@@ -91,44 +82,49 @@ pub fn require_config_source() -> Result<(), crate::AppError> {
...
+    storage: &cda_storage::LocalStorage,
+    config_file: &Path,
+) -> Result<(), crate::AppError> {
+    let data =
+        std::fs::read(config_file).map_err(|source| crate::AppError::ConfigurationError {
```
alexmohr: q: why are we not using `tokio::fs::read(config_file).await` instead?

cda-main/src/config.rs
```diff
@@ -91,44 +82,49 @@ pub fn require_config_source() -> Result<(), crate::AppError> {
...
-    let storage = match cda_storage::LocalStorage::new(storage_dir) {
-        Ok(s) => s,
-        Err(e) => {
-            tracing::warn!(error = %e, "Storage not available, skipping seed");
-            return;
-        }
-    };
+        .map_or_else(
+            || config_file.to_string_lossy(),
+            |file| file.to_string_lossy(),
```
alexmohr: nit: We probably should _not_ file when `file_name` is empty but reject to load the configuration file.

> Returns [None](https://doc.rust-lang.org/std/option/enum.Option.html#variant.None) if the path terminates in `...`

alexmohr (commented):
couple of minor things. 
I won't be approving this PR though because I was pairing you quite a bit and I think it defeats the purpose of an unbiased review. So @theswiftfox or @floroks should review this as well.

cda-main/src/config.rs
mbfm: Ah, that renaming will come with the next PR. I wanted to get these changes underway, since these commits were a valid change by themselves and I don't know how much more will be added to the other PR.

cda-main/src/config.rs
mbfm: I've switched over to `tokio::fs`. It was already in the code like that before, but I'd hope that if there really was a good reason to use `std::fs`, that it would be documented with a comment...

cda-main/src/config.rs
mbfm: Ah, good point. I changed it to return an error instead.

cda-main/src/config.rs
alexmohr: alright, makes sense :) thx.

floroks (approved):
lgtm

Issue #484, opened 2026-08-05
Improvement: Enable configured ISO-TP mixed addressing in CAN ECU connections


lorinciubeica10:
## Description

CAN ECU mappings already allowed configuring address_extension but the value was not applied when creating ISO-TP socket options. As a result, mixed addressing settings in config had no runtime effect.

## Motivation
This change is needed because mixed ISO-TP addressing was configurable but not effectively usable end-to-end in the CAN path.

Users can now enable mixed addressing directly in CAN ECU mappings with a clear config field [address_extension], which makes setup explicit and predictable.

## Proposed Solution

Pass [address_extension] through CAN gateway connection creation for explicit CAN mappings.
Update ISO-TP socket option construction to conditionally enable mixed addressing when [address_extension] is present.
Enable [CAN_ISOTP_EXTEND_ADDR] and [CAN_ISOTP_RX_EXT_ADDR] flags together with TX padding in mixed mode.
Keep previous behavior unchanged when no [address_extension] is configured.

## Additional Context

(Optional) Add any screenshots, related issues, or architectural concerns.

> [!IMPORTANT]
> Make sure to link this issue with the PR for your improvement.

PR #485, opened 2026-08-05
feat: add optional ISO-TP address extension for mixed addressing

lorinciubeica10:
## Summary

Pass [address_extension] through CAN gateway connection creation for explicit CAN mappings.
Update ISO-TP socket option construction to conditionally enable mixed addressing when [address_extension] is present.
Enable [CAN_ISOTP_EXTEND_ADDR] and [CAN_ISOTP_RX_EXT_ADDR] flags together with TX padding in mixed mode.
Keep previous behavior unchanged when no [address_extension] is configured.
## Checklist

## Related

https://github.com/eclipse-opensovd/classic-diagnostic-adapter/issues/484


opensovd-cda-can.toml
```diff
@@ -97,11 +99,13 @@ protocol = "UDS_CAN"
 # ecu_name = "MyECU"
 # request_id = 0x7E0    # Physical request CAN ID
 # response_id = 0x7E8   # Physical response CAN ID
+# address_extension = 0xF1  # Optional mixed-addressing extension byte
```
alexmohr, line 102: suggestion: I would prefer if we could have a more literal naming instead of putting the iso tp configuration bitmask directly into the configuration. The communication library should be an implementation detail not something we expose in the configuration. Even a bunch of booleans would be better imho. 

Tagging @theswiftfox because I will be on vacation and you might want faster feedback on this.

cda-comm-can/src/gateway/connection.rs
```diff
@@ -53,14 +62,17 @@ impl CanEcuConnection {
         let src = self.response_id.to_socket_id()?;
         let dst = self.request_id.to_socket_id()?;
 
-        // Enable TX padding to send 8-byte CAN frames (required by many ECUs)
+        let mixed_addressing = self.address_extension.is_some();
+        let ext_address = self.address_extension.unwrap_or(0);
+        let behaviour = Self::isotp_behaviour(mixed_addressing);
+
```
theswiftfox: suggestion: i'd structure this a bit differently. something like this:
```rust
let (iso_tp_behavior, ext_address) = Self::resolve_addressing(&self.address_extension);

...

fn resolve_addressing(address_ext: &Option<u8>) -> (IsoTpBehaviour, u8) {
    let mut behaviour = IsoTpBehaviour::CAN_ISOTP_TX_PADDING;
    if address_ext.is_some() {
        behavior = behaviour
                | IsoTpBehaviour::CAN_ISOTP_EXTEND_ADDR
                | IsoTpBehaviour::CAN_ISOTP_RX_EXT_ADDR;
    }
    (behavior, address_ext.unwrap_or(0))
}
```

lorinciubeica10: fixed it

integration-tests/tests/util/http.rs
```diff
@@ -194,63 +195,91 @@ pub(crate) async fn send_request(
     headers: Option<&HeaderMap>,
     url: reqwest::Url,
 ) -> Result<Response, TestingError> {
-    let client = reqwest::Client::new();
-    let mut request_builder = client.request(method, url.clone()).header(
-        reqwest::header::CONTENT_TYPE,
-        mime::APPLICATION_JSON.essence_str(),
-    );
+    let deadline = TokioInstant::now()
```
alexmohr: question: did you have such an error? 
The commit is intermangled with the addressing changes. I am fine with fixing this in the same PR, if this is an actual problem, but please separate the commits cleanly and provide some details in the commit message or the PR why this was fixed and what the error case is / was.

alexmohr: In general this feels like a hack doing the update check now for *every* send request. I would prefer it *a lot* if we could instead make sure that the test that is triggering the update is making sure it finishes instead of this drive by fix in the send path.

alexmohr (changes requested):

integration-tests/tests/util/http.rs
lorinciubeica10: Yes, I had this issue in my first commit. 
I think it s happening sporadically since I didnt encounter it on other branch.
I reverted and will create a new PR for it.

alexmohr (dismissed):
thank you. Looks good to me now. Only reason why I am not merging this right away is to give @theswiftfox the chance to resolve her own thread. :)

alexmohr:
fyi @lorinciubeica10 I removed the approval because the commits are still not separated cleanly. Please squash or seperate the changes.

lorinciubeica10:
@alexmohr I squashed them

alexmohr (approved):

PR #348, opened 2026-05-20
Add basic bazel build (S-Core Integration)

theswiftfox:
## Summary

This PR adds a basic bazel build for the cda. Some adaptions were made to the mbedtls build.rs to support building via bazel instead of during the rust build step.

## Checklist

- [ ] I have tested my changes locally
- [ ] I have added or updated documentation
- [ ] I have linked related issues or discussions
- [ ] I have added or updated tests

## Notes for Reviewers

---
Elena Gantner [elena.gantner@mercedes-benz.com](mailto:elena.gantner@mercedes-benz.com), Mercedes-Benz Tech Innovation GmbH
[Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)

PR #488, opened 2026-08-07
Feat: add bazel build (S-Core Integration)

### LINKED ISSUE #439: Add bazel build (S-Core Integration)
## Description

Based on the Draft in #348 a bazel build shall be added to the CDA to support integration into S-Core.

To achieve a clean integration we need to adapt the CDA in a few ways:
- build needs to support building completely without openssl dependency (avoid transitive dependencies that drag it in / gate behind openssl feature)
- mbedtls needs to be moved into a standalone repo that provides the patched mbedtls source + rust bindings (avoids having to fetch during compilation of cda and allows to create a seperate bazel module that can be referenced then)
- ensure cross compilation works without workarounds (to support bazel builds for different targets our cross-compile target must not reference host libraries except for running build tasks)
- (probably more, but those will have to be resolved once we have a solution for the first points and can run the build)

> [!IMPORTANT]

theswiftfox:
## Summary

Clean slate follow up for #348

- Removed mbedtls from repository and uses https://github.com/eclipse-opensovd/mbedtls-rs as dependency instead.
- provides bazel build & pipeline

Closes #439 

## Checklist

## Notes for Reviewers

---


theswiftfox:
@FScholPer if you want you can test this PR for integration in s-core. It is built up cleanly from mainbranch and uses the external mbedtls module.
note: it is draft as the mbedtls is not yet merged to main and therefore the refs are not final.

cda-main/BUILD.bazel
```diff
@@ -0,0 +1,76 @@
...
+        "tokio",
+        "toml",
+        "tracing",
+        "tracing-subscriber",
+    ],
+    crate_name = "opensovd_cda_lib",
+    crate_root = "src/lib.rs",
+    proc_macro_crate_deps = ["async-trait"],
+    rustc_env = {
+        "BUILD_DATE": "1970-01-01T00:00:00Z",
```
alexmohr: q: Why are these hard coded values? This means a different build system will yield different versions etc and thus yield a different result on the health endpoint.

Same for the package version

cda-main/BUILD.bazel
```diff
@@ -0,0 +1,76 @@
...
+        "CARGO_PKG_VERSION": "0.1.0",
+        "GIT_COMMIT_HASH": "unknown",
+    },
+    rustc_flags = [
+        "--cfg=feature=\"config-optional\"",
+        "--cfg=feature=\"health\"",
+        "--cfg=feature=\"mbedtls\"",
+    ],
+    visibility = ["//visibility:public"],
+    deps = [
```
alexmohr, line 61: q: transport router create is missing here, compared to main. But it probably makes sense to hold rebasing off until #503 is merged and then adjust the dependencies in all files.

cda-comm-can/BUILD.bazel
```diff
@@ -0,0 +1,32 @@
...
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+
+load("//bazel:rust_crate.bzl", "workspace_rust_library")
+
+workspace_rust_library(
```
alexmohr, line 14: q: How is support for can enabled on the bazel level? If it's not possbile yet it might make sense to expose a bazel setting, to toggle can support.

.github/workflows/build.yml
```diff
@@ -28,6 +28,33 @@ concurrency:
...
+      - name: Verify mbedTLS revision consistency
+        run: python3 bazel/check_mbedtls_revision.py
+      - name: Install Bazelisk
+        uses: bazelbuild/setup-bazelisk@v3
+      - name: Build mbedTLS library
+        run: bazel build //:cda_mbedtls
+      - name: Build reference binary
+        run: bazel build //:opensovd-cda
+      - name: Query targets
+        run: bazel query //...
```
alexmohr, line 50: q: should `target` be excluded via bazel ignore? Not relevant for the pipeline but might interfere on local builds.

FScholPer:
Anything I can do to get the thing merged?!

theswiftfox:
> Anything I can do to get the thing merged?!

I don't think so. We will do a final review and get it in soon. But thanks for asking :) From S-Core side I think it should still be okay as the only thing I did was a rebase and small cleanup.

Issue #493, opened 2026-08-11
[Bug] UDS session state is not updated after a successful session-control response


lorinciubeica10:
## Description

When a UDS session-control request succeeds, CDA sends the correct diagnostic communication but does not update the ECU runtime state with the newly active session.

As a result, subsequent state-transition lookups use the previous session, which can leave the ECU logically stuck in ProgrammingSession and prevent valid transitions to other sessions.

## Steps to Reproduce

1. Request Programming Session
2. Request Extended Session

## Expected Behavior

After a positive UDS session-control response:

The runtime [SESSION_CONTROL] state is updated.
The stored value is the canonical target session name from the active state chart.
Subsequent session transitions resolve from the newly active session.

## Actual Behavior

Internal state stuck to Programming Session

## Screenshots or Logs (if applicable)

Add logs, screenshots, or crash traces that help explain the problem.

## Environment

- OS: [e.g. Ubuntu 22.04, Windows 11]
- Version: [e.g. v0.2.1, main]

## Additional Context

> [!IMPORTANT]
> Link the pull request that addresses this bug once created.

PR #494, opened 2026-08-11
Fix: Session state stuck in programming session

lorinciubeica10:
lookup_state_transition_by_diagcomm_for_active lacked the default-state fallback that prevented session updates from non-default sessions.

## Checklist

## Related

https://github.com/eclipse-opensovd/classic-diagnostic-adapter/issues/493


cda-interfaces/src/ecumanager.rs
```diff
@@ -621,13 +621,14 @@ pub trait EcuStateManager: Send + Sync + 'static {
...
     /// Lookup the transition between the active session and the requested one.
+    /// Returns the diag comm used for the transition and the canonical target session name.
     /// # Errors
     /// * `DiagServiceError::AccessDenied` if no transition exists
     /// * `DiagServiceError::NotFound` on various lookup errors.
     fn lookup_session_change(
         &self,
         session: &str,
-    ) -> impl Future<Output = Result<DiagComm, DiagServiceError>> + Send;
+    ) -> impl Future<Output = Result<(DiagComm, String), DiagServiceError>> + Send;
```
alexmohr: suggestion: Please do not use string as error, revert this and keep using DiagServiceError

alexmohr (changes requested):
Thanks for the change. In general looks good. One finding inline. 
Aside from this, it would be great to add an integration test or at least a unit test to prevent regressions.

alexmohr:
```
    sovd::runtimefiles::runtimefiles_execution_responses_follow_operation_standard
```

This is a flaky test. I have fixed in #450, so you can ignore that.

cda-comm-uds/src/session.rs
```diff
@@ -118,6 +138,7 @@ impl<S: EcuGateway, T: EcuManager> UdsSession for UdsManager<S, T> {
         let result = self
             .send_with_optional_timeout(ecu_name, dc, security_plugin, None, true, None)
             .await?;
+        update_session_state_on_positive_response(ecu_diag_service, session, &result).await;
         match result.response_type() {
```
alexmohr, line 121: nitpick: Why did you add `update_session_state_on_positive_response` which checks the response, when the match block below has a branch for positive response anyway? It would be more efficient to do it in the match block before the start task, so we do not have to check the response type twice in a row.

cda-comm-uds/src/session.rs
```diff
@@ -167,3 +175,67 @@ impl<S: EcuGateway, T: EcuManager> UdsSession for UdsManager<S, T> {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
theswiftfox, line 172: q: i don't fully understand the reason for this tests. they do not test the check you added but rather duplicate the code block you added above? if someone changed the code the tests would still be green as they do not check the actual logic.
You would need to test `set_ecu_session` with mocked positive / negative responses to get a valuable result from the tests.

alexmohr:
@lorinciubeica10 After you addressed the @theswiftfox comment, you can rebase onto main to fix the duplicate code check :)

cda-comm-uds/src/transport.rs
```diff
@@ -75,7 +75,7 @@ impl<S: EcuGateway, T: EcuManager> UdsManager<S, T> {
                     variant_state = ?status.variant_state,
                     "Triggering variant detection before send"
                 );
-                if let Err(e) = self.detect_variant_if_needed(ecu_name).await {
+                if let Err(e) = self.detect_variant(ecu_name).await {
```
alexmohr: question: Why are you changing this from `detect_variant_if_needed` to `detect_variant`, which forces a variant detection for all `send_with_optional_timeout` calls? `send_with_optional_timeout` is called in a bunch of places, not related to session handling.

lorinciubeica10: the change from detect_variant_if_needed to detect_variant was wrong, forcing variant detection would affect unrelated requests. 
I restored detect_variant_if_needed

cda-comm-uds/src/session.rs
```diff
@@ -167,3 +175,141 @@ impl<S: EcuGateway, T: EcuManager> UdsSession for UdsManager<S, T> {
...
+        assert_eq!(current_session(&manager).await.as_deref(), Some("Default"));
+    }
+
+    /// Regression test for the ECU appearing stuck in the programming session.
+    /// The state chart lookup that populates `ServicePayload::new_session` only
+    /// resolves transitions whose source is the active state, so once the ECU
+    /// had left the default session that fallback stopped firing and the stored
+    /// state was never overwritten again.
+    ///
+    /// The double cannot reproduce that lookup (it has no state chart), so what
```
alexmohr: q: ` The double cannot reproduce that lookup ` what do you mean by 'the double`?

lorinciubeica10: I meant test double. 
TestEcuDb is rudimentary, has no state chart and therefore cannot reproduce that lookup. 
For clarity, I can replace double with TestEcuDb.

alexmohr: > For clarity, I can replace double with TestEcuDb.

That would be great, thank you.

lorinciubeica10: fixed it

cda-comm-uds/src/transport.rs
```diff
@@ -916,7 +916,7 @@ mod tests {
 }
 
 #[cfg(test)]
-mod send_tests {
+pub(crate) mod send_tests {
```
alexmohr, line 919: suggestion: as this is re-used in another test now imho it would make sense to make the re-used functions part of a test_helpers module instead, similar to test_helpers in cda-comm-uds.

cda-comm-uds/src/session.rs
```diff
@@ -120,6 +120,14 @@ impl<S: EcuGateway, T: EcuManager> UdsSession for UdsManager<S, T> {
             .await?;
         match result.response_type() {
             DiagServiceResponseType::Positive => {
+                ecu_diag_service
+                    .read()
+                    .await
+                    .set_service_state(
+                        cda_interfaces::service_ids::SESSION_CONTROL,
+                        session.to_owned(),
```
alexmohr: todo: Don't call `set_service_state` here. This won't apply the changes when `send_genericservice` is used. 
Also this is already tracked through `lookup_state_transition_by_diagcomm_for_active`. 
But speaking of that, this function:

`lookup_state_transition_by_diagcomm_for_active` is currently *not* doing the same default session fallback as `lookup_state_transition_for_active` so we should close this gap. Although strictly speaking it's an issue on main and I only found it due to your PR. I attached a patch, you can apply but don't have to. If not I will crate a follow up. 
In any case. These added lines do not belong here.
See cda-comm-uds/src/transport.rs:562-565 for reference:
```rust
if let Some(new_session) = payload.new_session {
    ecu_read
        .set_service_state(service_ids::SESSION_CONTROL, new_session)
        .await;
}
```

[session.patch](https://github.com/user-attachments/files/31959621/session.patch)

alexmohr (approved):
thank you, lgtm.

PR #500, opened 2026-08-12
Feat: Support RequestSeed parameters in security access API

lorinciubeica10:
Add parameters field 
Route parameters to RequestSeed and authentication to SendKey.

## Summary

Added optional parameters that can be passed when requesting seed.
## Checklist


integration-tests/tests/sovd/flash_download.rs
```diff
@@ -124,6 +124,7 @@ async fn test_flash_download_transfer_sequence() {
             value: "Level_7_RequestSeed".to_owned(),
             mode_expiration: None,
             key: None,
+            parameters: None,
```
alexmohr, line 127: note: It would be great to have an integration test that is actually setting the parameters. It's an important feature and it is easy to break without such a test. 
To achieve this you'd have to adjust the ECU sim accept and one of the demo ecus to request such a parameter on the request seed service. (If you do so, please do not change the FLXC1000 ecu.)

@floroks Is most likely to help you here if you have questions on how to implement this in detail.

lorinciubeica10: test was added

cda-sovd-interfaces/src/components/ecu/modes.rs
```diff
@@ -64,6 +64,9 @@ pub mod security_and_session {
 
             #[serde(rename = "Key", alias = "SendKey")]
             pub key: Option<ModeKey>,
+
+            /// Optional parameters forwarded as payload to the `RequestSeed` service.
```
alexmohr, line 82: suggestion: This request is shared by session and security requests. 
The added field for `parameters` is only viable for 'security` but not for session. 
The problem pre-exists because the same is true for `key`. This means in the openapi documentation fields are advertised for session which are acutally never read. 
To fix this, it would be correct to split this into two structs.

```rust
pub struct SessionRequest {
    pub value: String,
    pub mode_expiration: Option<u64>,
}

pub struct SecurityRequest {
    pub value: String,
    pub mode_expiration: Option<u64>,
    pub key: Option<ModeKey>,
    pub parameters: Option<HashMap<String, serde_json::Value>>,
}
```

Question to @floroks: 
ISO 17978-3 section 7.16.4, Table 237 only defines value and optional mode_expiration, meaning `key` already is an extension. Aside from backwards compatibilty, why didn't we label the field clearly as `x-sovd...` so it is clearer that this is an extension.

cda-sovd/src/sovd/components/ecu/modes.rs
```diff
@@ -468,7 +468,7 @@ pub(crate) mod security {
             .into_response();
         }
 
-        let payload = if let Some(key) = key {
+        let (seed_payload, key_payload) = if let Some(key) = key {
```
alexmohr, line 480: suggestion: given the payload

```json
{
  "value": "Level_5",
  "Key": {
    "Send_Key": "0x12 0x34"
  },
  "parameters": {
    "Foo": 90
  }
}
```

The 'parameters' are accepted on the http layer, but then ignored for the actual send key routine. 
I would prefer if we would reject send key with parameters instead (or support them there as well)
To deny requests containing them, something like below could be used:

```rust
if !is_request_seed && request_body.parameters.is_some() {
    return ErrorWrapper {
        error: ApiError::BadRequest(
            "RequestSeed parameters cannot be used with SendKey.".to_string(),
        ),
        include_schema,
    }
    .into_response();
}
```

lorinciubeica10: fixed

cda-sovd-interfaces/src/components/ecu/modes.rs
floroks: I believe it's only due to backwards compatibility.

alexmohr: Thanks, in that case imho it makes sense to keep the parameters as is.

integration-tests/tests/sovd/ecu.rs
```diff
@@ -460,6 +462,159 @@ async fn test_ecu_session_switching() {
     .await;
 }
 
+/// A `RequestSeed` parameter must be encoded into the UDS request; FSNR2000's
+/// simulator rejects the seed request unless it receives the configured byte.
+#[tokio::test]
```
alexmohr, line 467: suggestion: It might make sense to add two test cases here, that the request seed without and with an invalid  value, a missing paramter and a wrong parameter



```diff 
diff --git a/integration-tests/tests/sovd/ecu.rs b/integration-tests/tests/sovd/ecu.rs
index 6e0360a..b3fea86 100644
--- a/integration-tests/tests/sovd/ecu.rs
+++ b/integration-tests/tests/sovd/ecu.rs
@@ -511,6 +511,71 @@ async fn request_seed_forwards_parameters_to_fsnr2000() {
     force_variant_detection(&runtime.config, &auth, ecu_endpoint)
         .await
         .expect("FSNR2000 boot variant should be detected");
+
+    let response: Option<RequestSeedResponse> = put_mode(
+        &runtime.config,
+        &auth,
+        ecu_endpoint,
+        "security",
+        sovd_interfaces::components::ecu::modes::security_and_session::put::Request {
+            value: "Level_5_RequestSeed".to_owned(),
+            mode_expiration: None,
+            key: None,
+            parameters: None,
+        },
+        StatusCode::BAD_GATEWAY,
+    )
+    .await
+    .expect("RequestSeed without parameters should be rejected by FSNR2000");
+    assert!(
+        response.is_none(),
+        "a rejected RequestSeed should not return a body"
+    );
+
+    let mut parameters = HashMap::new();
+    parameters.insert("Invalid".to_owned(), json!(0x5A));
+    let response: Option<RequestSeedResponse> = put_mode(
+        &runtime.config,
+        &auth,
+        ecu_endpoint,
+        "security",
+        sovd_interfaces::components::ecu::modes::security_and_session::put::Request {
+            value: "Level_5_RequestSeed".to_owned(),
+            mode_expiration: None,
+            key: None,
+            parameters: Some(parameters),
+        },
+        StatusCode::BAD_GATEWAY,
+    )
+    .await
+    .expect("RequestSeed without SeedRequestParameter should be rejected by FSNR2000");
+    assert!(
+        response.is_none(),
+        "a rejected RequestSeed should not return a body"
+    );
+
+    let mut parameters = HashMap::new();
+    parameters.insert("SeedRequestParameter".to_owned(), json!(0x5B));
+    let response: Option<RequestSeedResponse> = put_mode(
+        &runtime.config,
+        &auth,
+        ecu_endpoint,
+        "security",
+        sovd_interfaces::components::ecu::modes::security_and_session::put::Request {
+            value: "Level_5_RequestSeed".to_owned(),
+            mode_expiration: None,
+            key: None,
+            parameters: Some(parameters),
+        },
+        StatusCode::BAD_GATEWAY,
+    )
+    .await
+    .expect("RequestSeed with an invalid parameter should be rejected by FSNR2000");
+    assert!(
+        response.is_none(),
+        "a rejected RequestSeed should not return a body"
+    );
+
     ecusim::start_recording(&runtime.ecu_sim, "fsnr2000")
         .await
         .expect("FSNR2000 recording should start");
```

```

alexmohr (commented):
99% great only 1 extension when it comes to tests.

alexmohr (approved):

PR #501, opened 2026-08-12
Clarify that seeding only triggers on first load

mbfm:
## Summary

It is already the current behavior, that seeding the configuration and databases only does something when nothing has been previously seeded. This PR documents that behavior.

For this, the CLI flags `--config` and `--databases-path` were renamed to `--seed-config` and `--seed-databases-dir`. The configuration flag for the databases directory was similarly adjusted.

Additionally, a `--force-seed-config` flag was introduced to seed from the configuration file even when the storage already contains an entry.

## Checklist

## Notes for Reviewers

* I've taken the liberty to rename it from `databases-path` to `-dir`, to make it more obvious that it should be a directory. Since the renaming is already a breaking change, it makes sense to group these together.

* The respective `--force-seed-databases` flag will come later. To implement it, I will also need to fix that the databases are currently not seeded to begin with, so this felt like something for a separate PR.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ It handles the communication to the ECUs, by using the communication parameters 
 2. Start the CDA, pointing it at your databases directory and DoIP interface:
 
    ```shell
-   cargo run --release -- --databases-path ./databases --tester-address <YOUR_IP>
+   cargo run --release -- --seed-databases-dir ./databases --tester-address <YOUR_IP>
    ```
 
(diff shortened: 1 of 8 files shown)


mbfm:
Converting to draft, since with the changes on `main`, this PR might not make sense anymore. Do we still need to seed the configuration, when hot reloading of the configuration is now removed?

mbfm:
Due to config hot reloading having been removed, most of the changes in this PR could be removed. As a result, I'm closing this PR and instead add the remaining commit into #508.

Issue #504, opened 2026-08-13
Add OEM usage examples to build/test pipeline for early API-break detection


floroks:
## Problem

The workspace exposes several extension points that OEMs implement/reuse to customize the diagnostic adapter (custom SOVD routes, storage backends, security plugins, runtime-update plugins). There is currently no example/consumer crate exercising these public APIs as part of this repo's build/test/CI, so breaking visibility/API changes to the underlying crates aren't caught until an external OEM integration actually breaks. We want early feedback (at PR time) instead.

## Reference example

`eclipse-opensovd/demo` (branch `feat/demo-oem-route`, directory [`cda-oem-routes`](https://github.com/eclipse-opensovd/demo/tree/feat/demo-oem-route/cda-oem-routes)) contains a concrete example of an OEM consumer crate implementing custom SOVD routes as a plugin. This should be used as a reference/basis for what gets mirrored into this repo's CI-checked examples.

## Extension points / plugins that may be reused by OEMs

- **`cda-storage`** (traits `Storage`, `Collection`, `DirectFileAccess` in `cda-interfaces::storage_api`) — custom storage backend (e.g. database-backed, cloud, encrypted) replacing `LocalStorage`.
- **`cda-plugin-security`** (traits `SecurityPlugin`, `SecurityPluginInitializer`, `AuthorizationRequestHandler`) — custom auth/authz (OAuth2/OIDC, LDAP, custom token validation, RBAC/ABAC) replacing `DefaultSecurityPlugin`.
- **`cda-plugin-runtime-update`** + `cda-interfaces::runtime_update_api` (traits `RuntimeUpdateSecurityPlugin`, `RuntimeReloaderPlugin`, `RuntimeFilesUpdatePlugin`, `VehicleComponentFactory`) — custom runtime MDD-database update handling (integrity/signature checks, live reload, vendor-specific file/component management).
- **`cda-sovd-interfaces`** — shared SOVD REST API types/traits used to build custom SOVD routes/handlers, as demonstrated by `cda-oem-routes` in the demo repo.

## Proposal

- Add example crate(s) (e.g. new workspace member(s) under an `examples/` directory) implementing minimal OEM-style consumers for each extension point above, mirroring the pattern used by `cda-oem-routes` in the demo repo.
- Wire these examples into `cargo build`/`cargo test` and the existing CI workflows (`build.yml`, `pr-checks.yml`, `nightly.yml`).
[...]

PR #503, opened 2026-08-13
feat(communication): Add deferred communication

alexmohr:
## Summary

Implement deferred initialization and plugin Infrastructure for this.
This allows a plugin, with a provided default impl,
to prevent the CDA starting up with active communication.

Communication will be re-enabled once requests are received that do require
real DoIP communication.
As alternative the previous behaviour with fully enabled communication
from the start is still available.

This feature can be used in scenarios where the CDA must not
start with active communication to prevent disruption of other systems.

The re-usable socket is removed as well because communication now shuts down
cleanly and we can restart communication properly without keeping the socket.

## Checklist

- [ ] I have tested my changes locally
- [ ] I have added or updated documentation
- [ ] I have linked related issues or discussions
- [ ] I have added or updated tests

## Notes for Reviewers

This is ... a bit ... much to review, start with the ADR, it should explain the ideas and reasonings quite well. 

Alexander Mohr [alexander.m.mohr@mercedes-benz.com](mailto:alexander.m.mohr@mercedes-benz.com), Mercedes-Benz Tech Innovation GmbH
[Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)

PR #507, opened 2026-08-13
[playground/testing] feat(oem): add extension API, decouple update plugin, add OEM examples 

alexmohr:
## Summary

Closes the gap issue #504 describes: nothing in this repository consumed the public extension points, so a breaking visibility or signature change was only found when an external OEM integration failed to build.

Extension API (`cda-main`):

- Add `opensovd_cda_lib::extensions`, a trait-based capability surface handed to hooks registered with `Setup::with_extension`: route registration under a reserved `/vehicle/<version>/x-*` namespace, diagnostics, and health. The diagnostics handle holds a component slot rather than a snapshot, so it stays valid across a runtime database update.
- Move vehicle construction (databases, transports, gateway, UDS manager) into a private `vehicle` module. Those signatures name concrete transport types and changed on every transport refactor, which is what broke integrations built on them; `extensions` expresses the same capabilities as traits.
- Add an `[oem]` configuration section, passed through verbatim.

Update plugin decoupling (`cda-plugin-runtime-update`):

- Drop the dependencies on `cda-database`, `cda-plugin-security`, `cda-sovd`, `cda-storage` and `cda-plugin-communication-management`; the plugin now builds against `cda-interfaces` alone, with those crates as dev-dependencies.
- Put MDD-specific reads behind `RuntimeFileInspector`, so the plugin carries no database format.
- Move the ISO 17978-3 bulk-data wire types to `cda-sovd-interfaces`; the plugin reports domain `RuntimeFile` values and `cda-sovd` renders them.
- Split `RuntimeFilesUpdatePlugin` into catalog / store / executor halves with a blanket impl, so an integration can replace one concern.
- Move the default authorization policy to `cda-main` as `DefaultUpdateSecurityHandler`, and decide the caller's identity by asking the security plugin instead of only checking that some vehicle lock exists.
- Move `RuntimeUpdateConfig` to `cda-main`: a replacement plugin should not inherit the standard plugin's configuration schema.

Examples:

- Add one runnable binary per extension point under `examples/`, each customising exactly one thing: SOVD routes, security plugin, storage backend, and runtime-update policy plus database format. They are workspace members but not default members, so a plain `cargo build` skips them while CI builds them in a step that names the extension point that broke.

## Checklist

## Notes for Reviewers

This is not really ready for a review yet. It is here to put a few ideas out and to play around how we can solve #503 and make it easier to add examples and use the whole thing as a library. 
At this time it's pretty much a side quest for me.
Note that you probably shouldn't review this in too much detail. Most of these things are only governed by myself and implemented by AI without getting a proper review yet. 
Main reason for opening this anyway is to make it public that this is something being worked.


cda-comm-uds/src/transport.rs
```diff
@@ -666,6 +666,120 @@ impl<S: EcuGateway, T: EcuManager> UdsTransport for UdsManager<S, T> {
...
+            .lookup_service_by_sid_and_name(service_id, name, None)?;
+
+        self.send(
+            ecu_name,
+            diag_comm,
+            security_plugin,
+            Some(UdsPayloadData::ParameterMap(params)),
+            map_to_json,
+        )
+        .await
```
alexmohr, line 781: Todo: move both functions back.

Issue #506, opened 2026-08-13
Multiple instances of `LocalStorage` used


mbfm:
Currently, we create multiple instances of `LocalStorage` throughout the codebase.  
This breaks the transaction system of `LocalStorage`, since there can now be multiple parallel transactions.

It also means that various fallbacks have been put into the codebase for when the storage isn't available, when we should have a more appropriate strategy, as the partition with the storage might not be mounted yet during startup of the CDA.

## Proposed solution

* [x] Initialize the `LocalStorage` once relatively at the start of `cda-main` and pass it into where it is used. Remove the fallback solutions.

* [x] Implement retries for accessing the storage, which waits a few minutes, if the storage isn't accessible. If the storage does not become available, log the problem and terminate the CDA.

> [!IMPORTANT]
> Link the pull request that addresses this bug once created.

mbfm:
Implemented.

PR #508, opened 2026-08-14
Fix: MDD files do not get seeded during startup

mbfm:
Up until now, the function for seeding the storage from the MDD files in the database directory was never actually called.
This meant that during the first update that came in through the Update-plugin, there was nothing to rollback to, in case of an error.

This PR fixes that and includes some minor code cleanups as well.

## Checklist

## Related

Came out of the work on issue #479.
Prerequisite for merging: #501

## Notes for Reviewers

---

### DIFF
--- a/cda-main/src/mdd.rs
+++ b/cda-main/src/mdd.rs
@@ -236,7 +236,7 @@ pub async fn resolve_mdd_paths(storage_dir: &str, database_dir: &str) -> Vec<Pat
             configured_dir = %database_dir,
             "No MDD files found in storage, falling back to configured database dir"
         );
-        match std::fs::read_dir(database_dir) {
+        let mdd_files = match std::fs::read_dir(database_dir) {
             Ok(files) => get_mdd_files_and_size(files)
                 .into_iter()
                 .map(|(p, _)| p)
@@ -245,7 +245,9 @@ pub async fn resolve_mdd_paths(storage_dir: &str, database_dir: &str) -> Vec<Pat
                 tracing::error!(error = %e, "Failed to read directory");
                 vec![]
             }
-        }
+        };
+        seed_storage_if_empty_from_mdd_files(storage_dir, &mdd_files).await;
+        mdd_files
     }
 }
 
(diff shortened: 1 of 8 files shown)


mbfm:
The duplicate code check is right that in the tests, the `LocalStorage` shouldn't be created multiple times and rather be part of the fixture: https://github.com/eclipse-opensovd/classic-diagnostic-adapter/actions/runs/31819325813/job/94828530807?pr=508#step:7:87

I decided against fixing that, because it would have required passing a `LocalStorage` instance into `seed_storage_if_empty_from_mdd_files()`, which should be done as part of: https://github.com/eclipse-opensovd/classic-diagnostic-adapter/issues/506

(I assume the duplicate check wouldn't fail, if the HTTP request went through.)

mbfm:
Feedback from @floroks:
* [x] The existing seeding logic should not load all MDD files into memory.
* [x] The logic should rather be a `seed_if_nonexistent`.
      Otherwise, users might clear the collection via the REST API and this triggers seeding anew. Seeding should only happen during the first run.
    * [x] Ensure creating the collection is covered by the transaction.
      * [ ] ~~Introduce test to ensure `create_collection()` is rolled back correctly.~~
      → This is already covered by the test `recovery_create_collection_with_write_removes_orphaned_dir`.

cda-main/src/mdd.rs
```diff
@@ -293,41 +295,30 @@ async fn load_mdd_paths_from_storage(storage_dir: &str) -> Option<Vec<PathBuf>>
...
-
-    if mdd_files.is_empty() {
-        tracing::debug!(database_path, "No MDD files found in database path to seed");
-        return;
-    }
+pub async fn seed_storage_if_empty_from_mdd_files(storage_dir: &str, mdd_files: &[PathBuf]) {
+    let mut seed_entries = vec![];
 
-    let entries = mdd_files.into_iter().filter_map(|(path, _)| {
+    for path in mdd_files {
```
mbfm, line 304: Had to switch to a more imperative `for` loop, so the `.await` when opening the file works.

cda-storage/src/storage_seed.rs
```diff
@@ -11,57 +11,58 @@
...
             return None;
         }
     };
 
-    match collection.is_empty().await {
-        Ok(true) => {}
-        Ok(false) => {
-            tracing::debug!(collection = %collection_name, "Collection already populated, skipping seed");
+    match storage.get_collection(collection_name).await {
+        Err(StorageError::CollectionNotFound(_)) => {} //continue
```
alexmohr: nit:
```suggestion
        Err(StorageError::CollectionNotFound(_)) => {} // continue
```

cda-main/src/mdd.rs
```diff
@@ -219,24 +219,24 @@ pub async fn load_databases<S: SecurityPlugin>(
 }
 
 /// Returns paths to MDD files, preferring files found in the CDA storage at `storage_dir`.
-/// Falls back to the configured `database_path` directory if storage is unavailable or empty.
-pub async fn resolve_mdd_paths(storage_dir: &str, database_path: &str) -> Vec<PathBuf> {
+/// Falls back to the configured `database_dir` directory if storage is unavailable or empty.
```
alexmohr: nitpick: The _configured_ value is `&config.database.seed_dir`. It didn't match 100% before either, but now the mismatch more obvious.

alexmohr: Similar, it might make sense to rename 'configured_dir' into 'seed_dir' in 236?

cda-main/src/mdd.rs
```diff
@@ -293,41 +295,30 @@ async fn load_mdd_paths_from_storage(storage_dir: &str) -> Option<Vec<PathBuf>>
...
         if key.is_empty() {
-            return None;
-        }
-        match std::fs::read(&path) {
-            Ok(data) => Some((key, data)),
-            Err(e) => {
-                tracing::warn!(path = %path.display(),
-                    error = %e, "Failed to read MDD file for seeding, skipping");
-                None
+            tracing::warn!(path = %path.display(), "Determined key for MDD file was empty. Skipping.");
```
alexmohr: nit: Could we rephrase this message to something like `Unable to determine filename for MDD file. Skipping.`? It's a bit less opaque than the current message.

alexmohr (commented):
only few nitpick, else lgtm :)

cda-main/src/mdd.rs
mbfm: > nitpick: The _configured_ value is `&config.database.seed_dir`. It didn't match 100% before either, but now the mismatch more obvious.

Good point. I changed `database_dir` to `database.seed_dir`, as that's the TOML path in the config file.

cda-main/src/mdd.rs
mbfm: Changed it to: `Unable to determine filename of MDD file as key for seeding. Skipping.`

alexmohr (approved):

PR #516, opened 2026-08-26
fix(doip): recover ECUs that are unreachable during discovery

alexmohr:
## Summary

Two discovery paths could permanently latch a transient ECU unreachability:

- Re-broadcast the DoIP Vehicle Identification Request every 10 seconds, up to three attempts, while a gateway remains unconnected.
- Continue scheduling an ECU after it is marked offline during variant detection so its configured offline-verdict retry budget can recover transient failures.

## Checklist

## Notes for Reviewers

This _should_ fix a flaky integration tests where some ECUs in the mixed case are not detected properly. I was not able to reproduce the issue locally or in the CI so far again. 

---

### DIFF
--- a/cda-comm-doip/src/vir_vam.rs
+++ b/cda-comm-doip/src/vir_vam.rs
@@ -29,6 +29,31 @@ use crate::{
     connections::{GatewayState, handle_gateway_connection},
     socket::DoIPUdpSocket,
 };
+
+/// Delay between VIR re-broadcasts while a gateway is unconnected.
+const GATEWAY_REDISCOVERY_INTERVAL: Duration = Duration::from_secs(10);
+
+/// How often to re-broadcast before giving up. A gateway that stays silent this
+/// long is absent rather than slow, and announces itself when it boots.
+const GATEWAY_REDISCOVERY_ATTEMPTS: u8 = 3;
+
+/// Broadcasts a VIR on `gateway_port`. Entities answer with a VAM.
+async fn broadcast_vir(
+    socket: &mut DoIPUdpSocket,
+    gateway_port: u16,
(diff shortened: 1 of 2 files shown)


theswiftfox:
General question: do we know how gateways behave when we send a VIR while being connected? Does it cause reconnects?
(As i read the PR description, we would send those as long as we have some gateway ecu loaded that is not yet connected up to 3 times right?)

alexmohr:
> do we know how gateways behave when we send a VIR while being connected?

No that's a good question, which I didn't consider. I had some issues with flaky integration tests on my local machine (but the system was under quite some CPU load at the time). Meaning I did not encounter the issue on a real vehicle or even with real ECUs. 

You read the PR (description) correctly but honestly given your comment I am a bit hesitant if we even want this considering ECUs might behave weird when they get a VIR while being connected. Probably worst case is that the ECU disconnects and we have to reconnect. This already should happen automatically but it might yield an unnecessary error on the API.

floroks:
To my knowledge, ECUs should only respond with the VIR-Response, and it should have no effect on pre-existing connections. When implemented, this recovery behavior should however be configurable - e.g. have configuration parameters for the amount of retries and delays, so it can be deactivated through configuration. 

Those issues should mostly happen on test benches, on real vehicles, ecus should always respond to the VIR, and if you persist the ECU detection, and ensure the initial detection was correct, it wouldn't be used at all.

alexmohr:
Thanks, as this intended to fix a one-off test flake I don't think it is worth the effort to make this configurable when it's not necessary anymore anyway once we implemented the persistence. I will close this PR. In case this should happen more often we can re-open it.

PR #526, opened 2026-09-02
fix(cda-sovd): use RAII instead of manually tracking states

### LINKED ISSUE #525: SOVD: ExecutionGuard has no Drop; a missed cleanup() permanently blocks runtime updates
## Description

`ExecutionGuard` in `cda-sovd/src/sovd.rs` is named and documented as a guard
("Owns cleanup for a reserved async operation execution and its communication
lease") but has no `Drop` implementation. Its `cleanup()` is invoked by hand at
ten call sites. Missing one leaks a `CommunicationGuard`, and a leaked
`CommunicationGuard` permanently blocks runtime updates.

`acquire_and_reserve_execution` makes two coupled mutations before an operation
runs:

1. `communication_access.acquire()` -> a `CommunicationGuard` stored in
   `communication_activities[exec_id]`
2. `reserve_execution(...)` -> an entry in `executions[service][exec_id]`

Both must be undone if the handler returns early. Which exit is correct differs
per branch, and the third case hands the obligation to a different part of the
file entirely:

| case | `cleanup()` |
| --- | --- |
| error / early return | yes |
| operation completes synchronously | yes |
| operation runs asynchronously | no - released later by one of ~9 other `release_communication_activity` call sites |

The cleanup cannot be a `Drop` impl today because it awaits: both maps are
behind `tokio::sync` locks. However, no acquisition of either map anywhere in
`cda-sovd` holds the lock across an `await` - every one is a short map
operation (`insert`, `remove`, `get_mut`, `len`). The async locks therefore buy
nothing and are what forces the cleanup to be manual:

```
async locks -> cleanup must be async -> Drop impossible -> manual cleanup x10 -> leak risk
```

## Steps to Reproduce

1. Add an early `return` to an operations handler between
   `acquire_and_reserve_execution` and completion, without calling
   `guard.cleanup()` (i.e. reproduce the mistake the current design allows).
2. Issue the corresponding SOVD operation request so that path is taken.
3. `POST /vehicle/v15/apps/sovd2uds/bulk-data/runtimefiles/executions` to start
   a runtime update.

## Expected Behavior

The reservation and its `CommunicationGuard` are released when the handler
returns, regardless of which path it took, and the runtime update starts.

## Actual Behavior

The `CommunicationGuard` is never released. There is no sweeper, reaper or
expiry for held guards - `guard_admission` reports `GuardsHeld`, `disable()`
returns `DisableError::InUse`, and every runtime update fails with `409
OperationsInProgress` for the remaining lifetime of the process. Only a restart
clears it.

n/a - the failure is silent until an update is attempted.

## Environment

- OS: any
- Version: main

## Additional Context

Suggested fix, in one self-contained change:

1. Change `executions` and `communication_activities` from `tokio::sync::RwLock`
   / `tokio::sync::Mutex` to `std::sync` equivalents. Safe today: no site holds
   either across an `await`.
2. With the cleanup now synchronous, give the type a real `Drop`, and rename it
   to what it is (e.g. `ExecutionReservation`).
3. Make the async hand-off explicit and consuming (e.g. `into_async() -> Uuid`),
   so "reserved but never resolved" is impossible rather than merely unlikely.

This is pre-existing on `main` and independent of any in-flight refactor, so it
is best done on its own rather than folded into unrelated work.

> [!IMPORTANT]

alexmohr:
## Summary
[fix(cda-sovd): make execution cleanup cancellation-safe](https://github.com/eclipse-opensovd/classic-diagnostic-adapter/commit/c4ac5693129e077737638e624209443931c11acd)

Store communication leases with their executions and roll back pending
reservations on drop so early returns cannot block runtime updates.

[fix(cda-sovd): reset in-flight state on cancellation](https://github.com/eclipse-opensovd/classic-diagnostic-adapter/commit/3a0fb02dcfe16e018fbb72ac4ccc6e49a6218760)

Use an RAII request guard so interrupted GET and STOP handlers cannot leave
executions permanently marked as busy.

## Checklist

## Related

Closes #525 

## Notes for Reviewers

---- 

### DIFF
--- a/cda-sovd/src/sovd.rs
+++ b/cda-sovd/src/sovd.rs
@@ -157,9 +157,12 @@ pub(crate) fn acquire_communication_activity(
 /// operations, `FgServiceExecution` for functional-group operations).
 pub(crate) trait ExecutionStatus {
     fn execution_status(&self) -> &sovd_ecu::operations::ExecutionStatus;
-    /// Returns `true` if a request for this execution is currently being processed.
-    /// Used to avoid sending multiple UDS requests simultaneously for the same execution.
+    /// Returns `true` if an HTTP handler currently has exclusive processing access to this
+    /// execution. Used to avoid sending multiple UDS requests simultaneously for the same
+    /// execution.
     fn is_in_flight(&self) -> bool;
+    /// Sets whether an HTTP handler currently has exclusive processing access to this execution.
+    /// This is independent of the diagnostic execution status.
     fn set_in_flight(&mut self, in_flight: bool);
     /// Should return `false` if an execution was created as a placeholder,
     /// but not finalized.
@@ -273,6 +276,8 @@ impl std::ops::DerefMut for ComparamExecution {
 pub(crate) struct ServiceExecution {
     pub parameters: serde_json::Map<String, serde_json::Value>,
     pub status: sovd_ecu::operations::ExecutionStatus,
+    /// Whether an HTTP handler currently has exclusive processing access to this execution.
+    /// This is independent of the diagnostic execution status.
     pub in_flight: bool,
     pub is_created: bool,
     communication_lease: Option<ExecutionLease>,
@@ -333,6 +338,8 @@ impl ExecutionStatus for ServiceExecution {
 pub(crate) struct FgServiceExecution {
     pub parameters: HashMap<String, serde_json::Map<String, serde_json::Value>>,
     pub status: sovd_ecu::operations::ExecutionStatus,
+    /// Whether an HTTP handler currently has exclusive processing access to this execution.
+    /// This is independent of the diagnostic execution status.
     pub in_flight: bool,
     pub is_created: bool,
     communication_lease: Option<ExecutionLease>,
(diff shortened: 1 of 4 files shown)


cda-sovd/src/sovd/functions/functional_groups/operations.rs
```diff
@@ -325,7 +318,7 @@ pub(crate) mod diag_service {
             } else {
                 None
             };
-            let executions = fg_executions.read().await;
+            let executions = lock_read(&fg_executions);
```
theswiftfox, line 321: q: are we sure this does not cause issues to not use an async lock here? I imagine if anything is holding a write lock the thread handling this request will completely lock up as the stdlock will block the future.

alexmohr: Yes, pretty much with the same argumentation that the kameo actors are using sync locks. The lock is only held pretty short for single lookups / inserts. 

We could promote https://rust-lang.github.io/rust-clippy/master/?search=await_holding_lock to error, but as we are not allowing warnings anyway either I don't see a need for this. 

I added a 2nd commit which improves this a bit. We had some 'clone' calls running while the lock was held unnecessarily. I moved them out of the lock scope. But aside from this I do not see how this could give us any issue. 

Especially because the guards are not 'Send' anyway, so no 'await' can run as long as the std lock guards are held.

cda-sovd/src/sovd/functions/functional_groups/operations.rs
```diff
@@ -652,25 +625,18 @@ pub(crate) mod diag_service {
...
-            crate::sovd::release_communication_activity(&communication_activities, &exec_id).await;
             build_operation_response(response_data, errors, include_schema)
         } else {
-            // force=false and Stop had errors - reset in_flight, keep execution alive for retry.
-            if let Some(op_map) = fg_executions.write().await.get_mut(&operation)
-                && let Some(exec) = op_map.get_mut(&exec_id)
-            {
-                exec.in_flight = false;
-            }
+            // force=false and Stop had errors - keep execution alive for retry.
```
theswiftfox, line 639: q: do we not need to reset in_flight here? or was that a bug before?

cda-sovd/src/sovd.rs
```diff
@@ -158,57 +157,149 @@ pub(crate) fn acquire_communication_activity(
...
 }
 
 /// Stored state for a single ECU routine execution (async lifecycle).
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub(crate) struct ServiceExecution {
     pub parameters: serde_json::Map<String, serde_json::Value>,
     pub status: sovd_ecu::operations::ExecutionStatus,
+    /// Whether an HTTP handler currently has exclusive processing access to this execution.
+    /// This is independent of the diagnostic execution status.
```
theswiftfox: nit: its a bit strange that only this field is documented. and the documentaiton repeats the trait docs of execution status.

cda-sovd/src/sovd.rs
```diff
@@ -227,25 +318,50 @@ impl ExecutionStatus for ServiceExecution {
...
 
 /// Stored state for a functional-group routine execution (async lifecycle).
 /// Unlike `ServiceExecution`, parameters are keyed by ECU name so that
 /// per-ECU identity is preserved across the execution lifecycle.
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub(crate) struct FgServiceExecution {
     pub parameters: HashMap<String, serde_json::Map<String, serde_json::Value>>,
     pub status: sovd_ecu::operations::ExecutionStatus,
+    /// Whether an HTTP handler currently has exclusive processing access to this execution.
```
theswiftfox: nit: (same as above) repeats documentaiton from trait. and is the only documented field

cda-sovd/src/sovd.rs
```diff
@@ -1237,6 +1387,43 @@ pub(crate) mod tests {
         activation_requests: AtomicUsize,
     }
 
+    struct CountingCommunicationAccess {
+        active: Arc<AtomicUsize>,
+    }
+
+    struct CountingGuard(Arc<AtomicUsize>);
```
theswiftfox: q: couldn't this just be `type CountingGuard = Arc<()>`
Arc should do internal ref counting and this seems to do it twice, without ever really using the counter value itself?

cda-sovd/src/sovd/functions/functional_groups/operations.rs
alexmohr: No bug before, no need to reset it manually now. 
Line 580 `let _request_guard = match guard_execution..` which returns `ExecutionRequestGuard` and its `Drop` impl now does `exec.set_in_flight(false);`
The whole idea of this change is that we do not have to do this manually but let the type system do its job :)

cda-sovd/src/sovd.rs
alexmohr: fixed

cda-sovd/src/sovd.rs
alexmohr: fixed

cda-sovd/src/sovd.rs
alexmohr: Technically yes, but I think it would make the tests assertions a bit more opaque because the active variable the tests hold of this will count towards the `strong_count` of the arc. So something like 

```rust
    #[test]
    fn reservation_drop_releases_execution_and_lease() {
       let active = Arc::new(AtomicUsize::new(0));
....
        drop(reservation);

        assert_eq!(active.load(Ordering::SeqCst), 0);
}
```

would become 
```rust
    #[test]
    fn reservation_drop_releases_execution_and_lease() {
       let active = Arc::new(());
....
        drop(reservation);

        // Strong count == 1 because local 'active' counts towards this count
        assert_eq!(Arc::strong_count(&active), 1);
}
```

Making the comment load bearing (scnr 😆)

cda-sovd/src/sovd/functions/functional_groups/operations.rs
theswiftfox: ooh, yah makes sense :D

cda-sovd/src/sovd.rs
theswiftfox: hmm, i think i'd still prefer the tests more verbose and the code 'cleaner'. its super small but in an arc atomic bool we pay the sync cost basically twice. then i'd rather have a load bearing comment in the test ;)
though tbf: `// expect only the local reference to remain` would be enough for me as a comment here :D

alexmohr: Sure, documented the +1 thing on the struct field.

theswiftfox (approved):
LGTM :)

PR #527, opened 2026-09-02
Fix: Use single instance of `LocalStorage`

mbfm:
## Summary

Implements the first step of #506. Retries will come in another PR.

## Checklist

## Notes for Reviewers

* Decided to deduplicate the use of `LocalStorage::new` in the test code of `cda-storage`, in hopes of making LLMs less prone of repeating this pattern.
* I considered introducing a static flag to track, if the `LocalStorage` has been initialized before. Then we could throw an error with a `debug_assert!()`, in case anyone introduces a second instance.
  Problem is that in principle, it is legal to have multiple `LocalStorage` instances, so long as they use a separate root directory, and it would break running tests in parallel. We also can't easily use `#[cfg(not(test))]` to skip this check in tests, because `test` is not set in integration tests.

---


cda-main/src/mdd.rs
```diff
@@ -246,22 +246,14 @@ pub async fn resolve_mdd_paths(storage_dir: &str, database_dir: &str) -> Vec<Pat
...
-        Ok(s) => s,
-        Err(e) => {
-            tracing::debug!(error = %e, "Storage not available, skipping storage MDD lookup");
-            return None;
-        }
-    };
-
+/// Returns paths to all MDD files found in the CDA `storage`.
+/// Falls back to an empty list if the storage collection cannot be accessed.
+async fn load_mdd_paths_from_storage(storage: &LocalStorage) -> Option<Vec<PathBuf>> {
```
alexmohr, line 256: q: Although present in the original code, why do we return None here instead of an error?

cda-storage/tests/recovery.rs
```diff
@@ -0,0 +1,359 @@
...
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#![cfg(test)] // Workaround to make Clippy in pre-commit hook recognize this file as test code
```
alexmohr, line 14: q: why do we need this workaround? The adjacent `[general.rs](https://github.com/eclipse-opensovd/classic-diagnostic-adapter/pull/527/changes#diff-3eb18b719f440f1ed7e4d4c9396451684ce5fa097b4da192f277303c17e4cc3a)` does not need it either

mbfm: It's not a problem in `general.rs`, because most code is in test functions, which have an implicit `#[cfg(test)]` thanks to the `#[test]` annotation.

So, the only code not implicitly made into test code is the fixture code, which in `general.rs` is just the short `create_test_storage()` function.
That function does not trigger the Clippy lint, because it does not use `.unwrap()`.

I tried to find a general solution, but we could also replace all the unwraps in the Fixture code with `.expect()` calls here: https://github.com/mbfm/classic-diagnostic-adapter/blob/d9d52c17ea4b4067b5ff2a009f18d20dd10281b2/cda-storage/tests/recovery.rs#L329-L359

cda-main/src/mdd.rs
mbfm: The only argument I can find, is that the only call is in the function directly above, where the Option is matched and the error information would at best be logged anyways.

But yeah, this whole code file uses Options for error-handling, which feels like a historic decision...

alexmohr: Should we create an issue to refactor this separately?

alexmohr: Created https://github.com/eclipse-opensovd/classic-diagnostic-adapter/issues/543

alexmohr (dismissed):

alexmohr (approved):

PR #528, opened 2026-09-03
feat(server): support binding the SOVD webserver to a Unix domain socket

### LINKED ISSUE #368: Improvement: Bind the CDA webserver to a Unix domain socket
## Description

Allow the CDA to serve the SOVD HTTP API over a Unix domain socket instead of a TCP host/port, so it can be reached with:

```bash
curl -s --unix-socket cda.sock http://localhost/...
```

This affects the webserver bootstrap in `cda-sovd` and the server configuration in `cda-main` (CLI args, TOML, env).

## Motivation

In the gateway -> CDA architecture described in eclipse-opensovd/opensovd#108, the CDA only needs to be reachable by the co-located `opensovd-gateway`, not from the network. A Unix socket is the natural transport for that internal link: access is governed by filesystem permissions (and peer credentials) rather than network exposure, keeping the CDA local-only and removing the need to allocate a port. This is the same approach Docker uses, exposing its daemon over `/var/run/docker.sock`.

## Proposed Solution

- Configure a socket path via CLI (`--unix-socket`), TOML (`[server] unix_socket`), or env var (`CDA_SERVER_UNIX_SOCKET`).
- When a socket path is set, bind only the Unix socket; otherwise keep today's TCP behavior unchanged (`0.0.0.0:20002` default). The two modes are mutually exclusive.

axum supports Unix sockets natively (`axum::serve` accepts a `tokio::net::UnixListener`), so no extra dependencies are needed - see the [axum unix-domain-socket example](https://github.com/tokio-rs/axum/blob/main/examples/unix-domain-socket/src/main.rs).

## Testing

The [`opensovd-client`](https://github.com/eclipse-opensovd/opensovd-core/tree/main/opensovd-client) doubles as a test client e.g. to verify Unix domain socket transports (`Client::connect_unix` / `Client::builder()`).

## Additional Context

Related: eclipse-opensovd/opensovd#108

floroks:
## Summary

Adds Unix domain socket support as an alternative to TCP for the SOVD HTTP server, configurable via --unix-socket, [server] unix_socket in TOML, or CDA_SERVER_UNIX_SOCKET. When set, it silently takes priority over the TCP address/port (both settings ignored, no error), and the server logs how/where it is listening.

- cda-sovd: WebServerConfig gains unix_socket; launch_webserver binds a UnixListener, removing any stale socket file left behind by a previous unclean shutdown (debug-logged), failing loudly if removal is not possible.
- cda-main: ServerConfig.unix_socket (TOML), --unix-socket CLI flag, wired into update_config/init_webserver. Env var support comes for free via the existing figment Env::prefixed("CDA") layer.
- Regenerated the reference opensovd-cda.toml to include the new field.
- Tests: unit tests in cda-sovd (bind, TCP-priority, stale-socket cleanup and failure), an end-to-end integration test, and config/CLI parsing tests in cda-main. Unix-socket-capable HTTP client (hyper/hyper-util/hyperlocal/http-body-util) added as a shared workspace dev-dependency for these tests.

Closes #368

## Checklist

## Notes for Reviewers

### DIFF
--- a/cda-main/src/config/configfile.rs
+++ b/cda-main/src/config/configfile.rs
@@ -135,12 +135,20 @@ pub struct ServerConfig {
     pub address: String,
     /// TCP port the server listens on.
     pub port: u16,
+    /// Path to a Unix domain socket to bind the server to, e.g. `/run/cda.sock`.
+    ///
+    /// When set, the server binds only to this Unix socket and `address`/`port`
+    /// are silently ignored - the two transports are mutually exclusive, with
+    /// the Unix socket taking priority.
+    #[serde(default)]
+    pub unix_socket: Option<String>,
 }
 impl Default for ServerConfig {
     fn default() -> Self {
         Self {
             address: "0.0.0.0".to_owned(),
             port: 20002,
+            unix_socket: None,
         }
     }
 }
@@ -413,6 +421,41 @@ description_database = "teapot"
                 vec!["Control_".to_string()]
             ))
         );
+        Ok(())
+    }
+
+    /// `server.unix_socket` defaults to `None`, keeping today's TCP-only
+    /// behavior unchanged when the field is absent from the TOML file.
+    #[tokio::test]
+    async fn load_config_toml_server_unix_socket_defaults_to_none()
+    -> Result<(), Box<dyn std::error::Error>> {
+        let config_str = r#"
+[server]
+address = "127.0.0.1"
+port = 12345
+"#;
+        let figment = Figment::from(Serialized::defaults(Configuration::default()))
+            .merge(Toml::string(config_str));
+        let config: Configuration = figment.extract()?;
+        assert_eq!(config.server.address, "127.0.0.1");
+        assert_eq!(config.server.port, 12345);
+        assert_eq!(config.server.unix_socket, None);
+        Ok(())
+    }
+
+    /// `server.unix_socket` can be set via TOML, alongside `address`/`port`
+    /// (the two are not mutually exclusive at the parsing level - the Unix
+    /// socket simply takes priority at bind time in `cda-sovd`).
+    #[tokio::test]
+    async fn load_config_toml_server_unix_socket() -> Result<(), Box<dyn std::error::Error>> {
+        let config_str = r#"
+[server]
+unix_socket = "/run/cda.sock"
+"#;
+        let figment = Figment::from(Serialized::defaults(Configuration::default()))
+            .merge(Toml::string(config_str));
+        let config: Configuration = figment.extract()?;
+        assert_eq!(config.server.unix_socket, Some("/run/cda.sock".to_owned()));
         Ok(())
     }
 
--- a/cda-sovd/src/lib.rs
+++ b/cda-sovd/src/lib.rs
@@ -31,6 +31,8 @@ pub use http::Method;
 use opensovd_axum_extra::ExtractHost;
 use sovd::apps::sovd2uds::bulk_data::runtimefiles::RuntimeUpdateRouteState;
 use tokio::net::TcpListener;
+#[cfg(unix)]
+use tokio::net::UnixListener;
 use tower::{Layer, ServiceExt as TowerServiceExt};
 use tower_http::{normalize_path::NormalizePathLayer, trace::TraceLayer};
 
@@ -50,6 +52,10 @@ pub const OPENAPI_JSON_ROUTE: &str = "/openapi.json";
 pub struct WebServerConfig {
     pub host: String,
     pub port: u16,
+    /// When set, the server binds to this Unix domain socket path instead of
+    /// `host`/`port`. Takes priority silently over the TCP settings when
+    /// present - the two are not combined.
+    pub unix_socket: Option<String>,
 }
 
 /// Static configuration for vehicle SOVD routes.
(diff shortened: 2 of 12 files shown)


cda-main/src/config/configfile.rs
```diff
@@ -135,12 +135,20 @@ pub struct ServerConfig {
     pub address: String,
     /// TCP port the server listens on.
     pub port: u16,
+    /// Path to a Unix domain socket to bind the server to, e.g. `/run/cda.sock`.
+    ///
+    /// When set, the server binds only to this Unix socket and `address`/`port`
+    /// are silently ignored - the two transports are mutually exclusive, with
+    /// the Unix socket taking priority.
+    #[serde(default)]
+    pub unix_socket: Option<String>,
```
alexmohr, line 144: suggestion, I am not a big fan of silently picking a default. 
What about rewriting this instead to the following

```rust
pub struct TcpTransport {
    /// IP address the server listens on.
    pub address: String,
    /// TCP port the server listens on.
    pub port: u16,
}

pub struct UnixSocketTransport {
    pub path: String,
}

pub enum ServerTransport {
    Tcp(TcpTransport),
    UnixSocket(UnixSocketTransport), 
}

pub struct ServerConfig {
    pub transport: ServerTransport,
}
```

This way there is no implied default and no silently ignored option.

cda-sovd/src/lib.rs
```diff
@@ -94,34 +101,75 @@ where
...
+                ))
+            })?;
+            tracing::info!("SOVD HTTP server listening on unix socket {socket_path}");
+            cda_interfaces::spawn_named!("webserver", async move {
+                let _ = axum::serve(listener, tower::make::Shared::new(service_with_middleware))
+                    .with_graceful_shutdown(shutdown_signal)
+                    .await;
+            })
+        }
+        #[cfg(not(unix))]
```
alexmohr, line 133: suggestion: While it makes sense to keep the validation here, an implementation of `ConfigSanity` for server config could catch this earlier.

cda-sovd/src/lib.rs
```diff
@@ -94,34 +101,75 @@ where
...
+                DoipGatewaySetupError::ServerError(format!(
+                    "Failed to bind to unix socket {socket_path}: {e}"
+                ))
+            })?;
+            tracing::info!("SOVD HTTP server listening on unix socket {socket_path}");
+            cda_interfaces::spawn_named!("webserver", async move {
+                let _ = axum::serve(listener, tower::make::Shared::new(service_with_middleware))
+                    .with_graceful_shutdown(shutdown_signal)
+                    .await;
+            })
```
alexmohr, line 131: suggestion:


```rust
cda_interfaces::spawn_named!("webserver", async move {
            let _ = axum::serve(listener, tower::make::Shared::new(service_with_middleware))
                .with_graceful_shutdown(shutdown_signal)
                .await;
        })
```

This is duplicated between both transports, why not move it outside, below, the `if`

cda-sovd/src/lib.rs
```diff
@@ -94,34 +101,75 @@ where
...
 
+    let middleware = tower::util::MapRequestLayer::new(rewrite_request_uri);
+    let trim_trailing_slash_middleware = NormalizePathLayer::trim_trailing_slash();
+    let service_with_middleware = middleware.layer(trim_trailing_slash_middleware.layer(service));
+
+    let webserver_task = if let Some(socket_path) = config.unix_socket {
+        #[cfg(unix)]
+        {
+            remove_stale_unix_socket(&socket_path)?;
+            let listener = UnixListener::bind(&socket_path).map_err(|e| {
```
alexmohr, line 121: q: Will fail if the parent dir of `socket_path` does not exist. We should at least document this on the `  pub unix_socket: Option<String>` field, so it's visible on the default config comments.

cda-sovd/src/lib.rs
```diff
@@ -341,3 +389,162 @@ pub(crate) mod test_utils {
...
+
+    use futures::FutureExt;
+    use http_body_util::Empty;
+    use hyper_util::{client::legacy::Client, rt::TokioExecutor};
+    use hyperlocal::UnixConnector;
+
+    use super::*;
+
+    /// Sends a GET request over a Unix domain socket using a real HTTP
+    /// client (`hyperlocal` on top of `hyper-util`) and returns the response
```
alexmohr, line 405: suggestion: The integration tests are using `reqwest` can we please stick to one client library if there isn't a very good reason to do so. 

I gave it a test and when building the client as shown below, it works just fine with `reqwest`, no need for a new dependency 

```rust

let client = reqwest::Client::builder()
    .unix_socket("/tmp/myapp.sock")
    .build()?;

```

cda-sovd/src/lib.rs
```diff
@@ -341,3 +389,162 @@ pub(crate) mod test_utils {
...
+        let (_dynamic_router, webserver_task) = launch_webserver(config, shutdown_signal)
+            .await
+            .expect("failed to launch webserver on unix socket");
+
+        assert_reachable_then_shutdown(&socket_path, shutdown_tx, webserver_task).await;
+    }
+
+    #[tokio::test]
+    async fn launch_webserver_unix_socket_takes_priority_over_tcp() {
+        let dir = tempfile::tempdir().expect("failed to create temp dir");
```
alexmohr, line 475: suggestion: Creating the socket and the `WebServerConfig` is the same for every test and could be consolidated.

alexmohr (changes requested):

PR #532, opened 2026-09-07
Retry loading `LocalStorage`

mbfm:
## Summary
Initializing the `LocalStorage` is now retried, if it fails. For example, the partition might not yet be available when the ECU is powered on, on which the CDA runs.

## Checklist

## Related
Implements the second step of #506.

## Notes for Reviewers
Not sure, how to test this without running an entire container for it. I used the [`backon`](https://crates.io/crates/backon) library, so most of the code is already tested by itself. Mainly would've been nice to see the error logging...

Personally, not a huge fan of the API of `backon`, but it has worked fine for us in openDuT, and I couldn't find another library that really looked better.
Perhaps the most popular library [`retry`](https://crates.io/crates/retry) doesn't work for `async` code, which would have been fine for this particular use-case, but not if we want to use the library for anything else.

---


cda-main/src/lib.rs
```diff
@@ -317,6 +320,8 @@ where
     UPB: UpdatePluginBuilder<SP>,
     CPB: CommunicationPluginBuilder,
 {
+    let storage = initialize_storage_with_retries(&config).await?; //done as first step, since the application will not function without storage
```
alexmohr, line 323: nit: 
```suggestion
    // done as first step, since the application will not function without storage
    let storage = initialize_storage_with_retries(&config).await?;
```

cda-main/src/lib.rs
```diff
@@ -1027,6 +1035,55 @@ pub fn setup_tracing(config: &Configuration) -> Result<TracingGuards, TracingSet
...
+
+    let storage = (async || {
+        LocalStorage::new(&config.runtime_update_config.storage_dir)
+    })
+    .retry(
+        backon::ConstantBuilder::new()
+            .with_delay(retry_delay)
+            .with_max_times(attempts),
+    )
+    .when(|error| match error {
```
alexmohr: q: is backon silently discarding the errors? It might be interesting to log them at least on debug.

cda-plugin-runtime-update/src/config.rs
```diff
@@ -20,28 +20,29 @@ pub struct RuntimeUpdateConfig {
     pub upload_body_limit_bytes: usize,
     /// Directory where the updatable database is stored.
     pub storage_dir: String,
+    /// How long to wait between trying to load the storage directory,
+    /// if it is unavailable, in milliseconds.
+    pub storage_dir_load_retry_delay_ms: u64, //FIXME want more granular than seconds
```
alexmohr, line 25: suggestion: remove the fixme, milliseconds are more granular than seconds. (Although even seconds are granular enough imho)

cda-main/src/lib.rs
alexmohr: But more important: The tracing is only setup in `init_webserver` (which by itself is wrong, as it should be hoisted up here.), this means the CDA stays silent for up to the configured retry time and the logs are silently dropped.

cda-main/src/lib.rs
```diff
@@ -1027,6 +1035,55 @@ pub fn setup_tracing(config: &Configuration) -> Result<TracingGuards, TracingSet
...
+        StorageError::Io(_)
```
alexmohr, line 1056: todo: std::io errors are currently converted into StorageError::IO through the `from` conversion on the error value 

```
    /// An I/O error occurred during a storage operation.
    #[error("I/O error: {0}")]
    Io(#[from] std::io::Error),
```

This means, for example, [std::io::ErrorKind::PermissionDenied](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.PermissionDenied) is turned into a plain StorageError::Io here and therefore a non-recoverable error is treated like a recoverable one. 
This applies to several more`ErrorKind` values. 

Although one could argue that it *is* an IO error that happened, I would have used the API the same way you did here and would expect a correct mapping on the storage side if corresponding types between the io kind and StorageError exist.

As this implementation leads to a useless retry loop I would suggest fixing the conversion on the cda-storage / cda-interfaces side to make the mapping behave like you are expecting here.

cda-main/src/lib.rs
```diff
@@ -1027,6 +1035,55 @@ pub fn setup_tracing(config: &Configuration) -> Result<TracingGuards, TracingSet
     })
 }
 
+/// Retry initializing the Storage, as for example the partition
+/// may not yet be available on which we want to store the files.
+async fn initialize_storage_with_retries(
```
alexmohr, line 1040: q: Can we add a unit test for this too?

alexmohr (changes requested):

cda-main/src/lib.rs
mbfm: In case you didn't see it in the PR text:

> Not sure, how to test this without running an entire container for it. I used the `backon` library, so most of the code is already tested by itself. Mainly would've been nice to see the error logging...

So, if I mocked `LocalStorage::new()` in the test, then we're basically just testing the retry-logic, which should already be tested because it's from a library.

I guess, we discussed the `testcontainers` crate since then. So, I could start the CDA with a read-only partition (similar to the test we still want to do for the seeding replacement, but obviously without seed data), and then check the logs that CDA keeps retrying, e.g. because it cannot create the storage dir.

Still feels weird, because even then, I'm ultimately just testing that `LocalStorage::new()` reports errors correctly.

So, I don't know, is this worth it? Any other ideas how to test this?

alexmohr: Tbh. I missed this in the PR description. 

>  which should already be tested because it's from a library.

That's a bold claim though. Eventually we probably should add a test to prevent regressions on readonly filesystems, but it is out of scope for this PR.

alexmohr:
lgtm, please squash the review commit, then we're good to merge.

alexmohr (approved):

mbfm:
I've squashed the commit. 🙂

PR #538, opened 2026-09-08
feat(doc): Improve documentation for deferred communication

alexmohr:
## Summary

* Link tests and implementation into docs
* Add Post-Update Deferred Communication chapter.
* Update existing charts
* Add Communication Initialization Mode State Flow

## Checklist

## Notes for Reviewers

### DIFF
--- a/cda-plugin-communication-management/src/plugin/default.rs
+++ b/cda-plugin-communication-management/src/plugin/default.rs
@@ -48,6 +48,7 @@ use crate::lifecycle::{
 
 /// Default communication plugin: applies `init_mode` policy on top of the
 /// framework-serialized lifecycle handle.
+/// [[ dimpl~default-communication-policy, Default deferred communication policy, dimpl ]]
 pub struct DefaultCommunicationPlugin {
     handle: CommunicationHandle,
     mode: CommunicationInitMode,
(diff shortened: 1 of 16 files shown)


cda-plugin-communication-management/src/plugin/default.rs
```diff
@@ -262,6 +265,38 @@ mod tests {
...
+                    | CommunicationState::Enabled
+            ));
+            tokio::time::timeout(std::time::Duration::from_secs(5), async {
+                while handle.state() != CommunicationState::Enabled {
+                    tokio::task::yield_now().await;
+                }
+            })
+            .await
+            .expect("activation did not complete");
+            assert_eq!(control.enables.load(Ordering::Relaxed), 1, "{mode:?}");
```
theswiftfox: q: is there a specific reason this is atomicusize instead of atomicbool? (i know this pr didn't change it, i just noticed it here)

theswiftfox (dismissed):
change in general looks fine to me. We could consider cleaning up the C-Style bools (usize 0/1) in tests before merging ;)

cda-plugin-communication-management/src/plugin/default.rs
theswiftfox: edit: nvm i saw that they are used as a counte rin one place. so my suggestion would be to rename it to `enable_count` or something so its clearer that it is **NOT** a bool :D

theswiftfox (approved):

Issue #544, opened 2026-09-10
Mismatch of parameter "path" after PR can-support migration to main


FEVLuM:
## Description

opensovd-cda.toml and opensovd-cda-can.toml have to be aligned
**old** opensovd-cda-can.toml keyword...
```
[database]
path = "./mdds" 
```
...leads to in case of using feature CAN.
```
Error: Error: Configuration error `[can] is configured but no usable ECU addressing was found: add [[can.ecu_mappings]] entries for ECUs present in the database (or provide CAN COM parameters in the MDD)
```
## Steps to Reproduce

1. opensovd-cda-can.toml is configured based on the template in main ./
```
[database]
path = "./mdds" 
```
2.
```
export CDA_CONFIG_FILE="opensovd-cda-can.toml"
cargo run --features can
```

## Expected Behavior

CDA should start successfully
```
12:26:29.119 INFO in cda_storage::local_storage msg: Local storage initialized root=.
12:26:29.243 INFO in cda_sovd msg: Runtime update routes added to webserver
12:26:29.335 INFO in opensovd_cda_lib msg: CDA fully initialized and ready to serve requests
```

## Actual Behavior

CDA isn't starting

## Screenshots or Logs (if applicable)

```
Error: Error: Configuration error `[can] is configured but no usable ECU addressing was found: add [[can.ecu_mappings]] entries for ECUs present in the database (or provide CAN COM parameters in the MDD)
```

## Environment

- OS: Ubuntu 26.04
- Version: main

## Additional Context

at least opensovd-cda-can.toml
```
[database]
path = "./mdds" 
```
....must be aligned with the opensovd-cda.toml and exchanged by i.e.:
```
[database]
seed_dir = "./mdds" 
```

alexmohr:
Thanks for reporting this bug. It would make sense to regenerate and validate the CAN configuration in the CI the same as the 'non CAN' configuration is.

PR #546, opened 2026-09-14
(config): rename CAN database path to seed_dir in opensovd-cda-can.toml

shegazyy:
## Summary

The `[database]` section in `opensovd-cda-can.toml` still used the old `path` key, which was renamed to `seed_dir` during the can-support migration to main. This caused CDA to fail at startup with:

Renamed `path` to `seed_dir` to match the schema used by `opensovd-cda.toml`.

## Checklist

## Notes for Reviewers

One-line config fix — no code changes. Verified by running cda-main's existing `config::generate::tests::can_example_config_parses_as_valid_config` test, which parses `opensovd-cda-can.toml` directly; it now passes. Didn't add a new test since this existing test already covers the exact failure mode, but it currently only checks parseability rather than key names — worth flagging if maintainers want stricter coverage there, though that's out of scope for this one-line fix.


alexmohr:
While this does fix the configuration it does not fix the underlying issue. the configuration will drift apart again as long as this is not addressed via a CI check. 
We can merge this change, but please remove 'Closes #544' from the PR description, as I would like to keep the issue to make sure we fix this eventually. Even better would be to fix this via CI right away.

shegazyy:
@alexmohr comment is updated and new PR is opened
https://github.com/eclipse-opensovd/classic-diagnostic-adapter/pull/548

alexmohr (approved):

PR #548, opened 2026-09-15
test(config): add CI check to catch stale/renamed keys in example configs

### LINKED ISSUE #544: Mismatch of parameter "path" after PR can-support migration to main

shegazyy:
## Summary

Per @alexmohr's review on #546: the config-drift class of bug (a standalone example `.toml` using a key the schema no longer has) can recur, since Figment silently ignores unknown keys instead of failing.

Adds `can_example_config_has_no_unknown_keys`, which reuses the round-trip-diff approach already used by `com_params::find_unknown_keys` (previously private, now `pub(crate)`) to validate per-ECU com_param overrides: serialize the merged `Configuration` back to TOML and diff its keys against the raw file. Any key present in the file but absent from the round-trip is one Figment silently dropped.

Also includes the `opensovd-cda-can.toml` `path` -> `seed_dir` fix itself, so the new test is green independent of whether #546 has merged yet.

## Checklist

## Related

Closes #544

## Notes for Reviewers

Verified this catches the exact #544 regression: reverting `opensovd-cda-can.toml`'s `seed_dir` back to `path` fails the new test with a clear message pointing at the offending key.

This only guards the standalone example config (`opensovd-cda-can.toml`) against drift from the schema — the auto-generated `opensovd-cda.toml` reference is already covered by existing generator tests (`generate_reference_config_matches_committed_file`, etc.). Happy to extend this pattern to other example/reference files if there are more I'm missing.


alexmohr:
Thank you, while this makes sure that the configuration does not drift when it comes to stale keys it still isn't as good as the procedure for the 'normal' config. 
The configuration still will drift for new keys and changed  docstrings. We must be able to re-generate the CAN configuration the same as 'normal' config and use this as diff mechanism instead. All other approaches will drift eventually.

